PR #2, opened 2021-07-09
Breaking: refine public API

kalzoo:
Reduces the number of visible functions and type definitions to those reasonably needed for outside use. Notably, this means removing public visibility for raw parsing and lexing functions, while leaving access to useful constructs like `Program`, `ScheduledProgram`, and all `Instruction` members and variants.

### DIFF
--- a/src/parser/error.rs
+++ b/src/parser/error.rs
@@ -43,17 +43,6 @@ pub enum ErrorKind {
         actual: Token,
         expected: String,
     },
-    /// An unknown identifier was encountered
-    UnknownIdentifier,
-
-    /// An invalid literal was encountered.
-    ///
-    /// When encountered, this generally means a bug exists in the data that
-    /// was passed in or the parsing logic.
-    InvalidLiteral,
-
-    /// A full parse was requested, but data was left over after parsing finished.
-    Partial,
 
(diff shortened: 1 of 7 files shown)


src/parser/error.rs
```diff
@@ -43,17 +43,6 @@ pub enum ErrorKind {
         actual: Token,
         expected: String,
     },
-    /// An unknown identifier was encountered
```
genos, line 46: Deleting code is such a pleasing feeling 😄

src/program/graph.rs
```diff
@@ -26,11 +26,12 @@ use indexmap::IndexMap;
 #[derive(Debug, Clone)]
 pub enum ScheduleErrorVariant {
     DuplicateLabel,
-    DurationNotRealConstant,
-    DurationNotApplicable,
-    InvalidFrame,
     UncalibratedInstruction,
-    UnscheduleableInstruction,
+    UnscheduableInstruction,
```
genos, line 30: Typo: "Unscheduable" is probably "Unschedu **l** able"

genos (commented):
Excellent, thanks @kalzoo! Modulo one typo comment, this LGTM 👍

PR #3, opened 2021-07-10
Make `program::graph` public

kalzoo:
This fixes an oversight from #2 which prevented access to necessary types in `program::graph`. That _could_ be fixed by re-exporting from `program` but there are enough of them, and they're cohesive around the concept of a `graph`, so I'll just expose the entire `graph` module itself.

I took the liberty of removing the `pub use graph::...` statement from the program module as well, which is a breaking change but in the interest of avoiding further confusion, I'll merge this as a fix and a patch version bump.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -16,5 +16,5 @@
 pub mod expression;
 pub mod instruction;
 mod macros;
-pub mod parser;
+pub(crate) mod parser;
 pub mod program;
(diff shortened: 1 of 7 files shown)


kalzoo:
I mistakenly created this from the wrong branch due to parsing of the link by my IDE. Closing in favor of #4

PR #9, opened 2021-07-15
Fix parsing/printing double precision

notmgsk:
See #12

### DIFF
--- a/src/expression.rs
+++ b/src/expression.rs
@@ -192,22 +192,6 @@ macro_rules! format_complex {
     }};
 }
 
-// Display floats in a human-friendly way using scientific notation.
-// See https://internals.rust-lang.org/t/pre-rfc-draft-g-or-floating-points-for-humans/9110/8
-#[macro_export]
-macro_rules! pretty_float {
-    ($value: expr) => {{
-        let mut attempt = format!("{:.2e}", $value);
-        if attempt.ends_with("e0") {
-            attempt = format!("{:.2}", $value)
-        }
-        if attempt.ends_with(".00") {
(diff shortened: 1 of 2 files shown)


src/expression.rs
```diff
@@ -17,12 +17,19 @@ use crate::{imag, instruction::MemoryReference, real};
...
 
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Debug, PartialEq, Eq, Error)]
 pub enum EvaluationError {
     Incomplete,
 }
 
+impl fmt::Display for EvaluationError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "SuperError is here!")
```
kalzoo, line 29: this looks unfinished

Cargo.toml
```diff
@@ -12,3 +12,4 @@ nom = "6.1.0"
 num-complex = "0.3.0"
 petgraph = "0.5.1"
 serde = { version = "1.0.125", features = ["derive"] }
+thiserror = "1.0.26"
```
kalzoo, line 15: Can you provide an MR description pitching this change?

kalzoo (changes requested):

Cargo.toml
notmgsk: removed

src/expression.rs
notmgsk: removed

kalzoo:
Good call, let's remove `pretty_float!` altogether then

genos (approved):

PR #13, opened 2021-07-22
New: support DEFCIRCUIT

notmgsk:


kalzoo (approved):

src/instruction.rs
```diff
@@ -479,7 +479,7 @@ impl fmt::Display for Instruction {
             } => {
                 let mut parameter_str: String = parameters
                     .iter()
-                    .map(|p| p.to_string())
+                    .map(|p| format!("%{}", p.to_string()))
```
genos, line 482: Are both `format!` and `to_string()` necessary?

src/parser/command.rs
```diff
@@ -414,4 +440,78 @@ mod tests {
             data: Some("{'module':'lodgepole.filters.io','filter_type':'DataBuffer','source':'q35_ro_rx/filter','publish':true,'params':{},'_type':'FilterNode'}".to_owned())
         }
     );
+
+    make_test!(
```
genos, line 444: Thank you for testing!

src/parser/common.rs
```diff
@@ -170,6 +170,26 @@ pub fn parse_qubits(input: ParserInput) -> ParserResult<Vec<Qubit>> {
...
+        Some((Token::Variable(name), remainder)) => Ok((remainder, name.clone())),
+        Some((Token::Identifier(name), remainder)) => Ok((remainder, name.clone())),
+        Some((other_token, _)) => {
+            expected_token!(input, other_token, stringify!($expected_variant).to_owned())
+        }
+    }
+}
+
+/// Parse zero or more variable qubits in sequence
+pub fn parse_variable_qubits(input: ParserInput) -> ParserResult<Vec<String>> {
```
genos, line 189: Adding this seems not worth the cost of having another `parse_X` function, to me. You only use it in one place, and typing `many0(parse_variable_qubit)(input)` directly doesn't seem like that much of a hardship.

genos (commented):
2 questions, to address or ignore as you see fit. Nicely done!

src/instruction.rs
kalzoo: good point ^

src/parser/common.rs
genos: Ah, but I see we did the same with `parse_qubit`/`parse_qubits`. Carry on.

genos: Actually, yeah, I'll open a small PR removing these two.

PR #11, opened 2021-07-16
Breaking: Properly Handle Instruction Scheduling

kalzoo:
- Implement `NONBLOCKING` parsing
- Correctly implement frame blocking/utilitization on `Program#get_frames_for_instruction`, replacing the misnamed `Program#get_blocking_frames`.
- Fix the parsing of `RAW-CAPTURE`, `CAPTURE`
- Add parsing for `FENCE`
- Fix display for several instructions including `DELAY`
- Correctly express instruction dependencies on frames
- Support DOT format export for `InstructionBlock` and `ScheduledProgram` to aid in program debugging
- Add snapshot testing and several snapshot tests to assert that dependencies are working as expected
- Fix how labels are auto-generated when `ScheduledProgram` cuts a program into blocks; prevent creation of empty blocks with no label
- Fix expression parsing to allow memory references (although the spec has ambiguity here)

TODO:
- [x] Tests for frame blocking
- [x] `CAPTURE` and memory dependencies

PR #16, opened 2021-07-22
Fix: Unbox calibrations and waveform invocations

### LINKED ISSUE #15: Directly store WaveformInvocation
Currently, in select `Instruction`s, there's a field `waveform: Box<WaveformInvocation>`. As pointed out by @genos in #11 , that's unnecessary and the field can be simply `waveform: WaveformInvocation`.

genos:
Both `CalibrationDefinition` and `Instruction` have some unnecessary `Box`es—around a `Calibration` in the former and a `WaveformInvocation` in the latter. This PR removes them, so only `Expression`s have a `Box` in them (and this one is necessary due to recursive definitions).

Resolves #15, first observed in #11

### DIFF
--- a/src/instruction.rs
+++ b/src/instruction.rs
@@ -246,11 +246,11 @@ pub enum Instruction {
     Reset {
         qubit: Option<Qubit>,
     },
-    CalibrationDefinition(Box<Calibration>),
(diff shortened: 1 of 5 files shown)


kalzoo (approved):
Thanks @genos !

Issue #8, opened 2021-07-14
Should Qubit::Fixed be usize?


notmgsk:
`Qubit::Fixed` is synonymous with an _index_, and so it is to be expected that they will be used as such (an index into an array). This is an issue now because `Qubit::Fixed` is `u64` which must be converted into a `usize` so that it can be used to index an array. That's usually fine because most of us have 64bit computers and so `u64` wouldn't be truncated. But it might be polite (and cleaner, in my experience -- no more `index as usize`) to define them as `usize`.

WDYT?

genos:
I'm a fan @notmgsk; lean on the type system a bit more, since to me `usize` says "I'm used as a size or index."

genos:
PR attempt in #17, though y'all should feel free to cast it into the fire if you don't like it.

kalzoo:
I'm not sure - `usize` is useful because it expresses a size corresponding to the target architecture...but in this case it represents a size corresponding to the _compiler's_ architecture. Should more qubits or longer memory references be available to a Quil program, executed on the same hardware, when compiled on a 64-bit system than 32-bit?

I agree that `u64 as usize` should be avoided when able. I could see this working.

genos:
Point well taken! Much as I like using `usize` for an index or length, having the bit length specified arch-agnostically seems a good idea. Does the language spec define the necessary integral size? And I don’t think I saw any `as` casting in my editing, so I may be swinging at shadows here anyway. I’ll kill #17 for now, and should we want we can revisit some other time.

kalzoo:
Going to close this for now. My assumption with Quil is that bit depth is unspecified and left to the compiler. I think it's fair to specify `u64`/`f64` for literal integers and real values respectively within our own compiler, returning errors on overflow.

PR #17, opened 2021-07-22
Update: change Qubit::Fixed, Token::Integer, Vector.length, MemoryRefence.index to usize

genos:
This was originally meant to only address #8, changing `Qubit::Fixed` from `u64` to `usize`. In doing so, the quickest fix was to also update `Token::Integer` to `usize`, which meant that to make the compiler happy, I needed to make `Vector.length` and `MemoryReference.index` into `usize`s too. I'm morally OK with this, as both `length` and `index` to me _sound_ like they should be of type `usize` instead of `u64`, but I leave it to the reviewer(s) whether this PR should be merged or burned with fire.

### DIFF
--- a/src/instruction.rs
+++ b/src/instruction.rs
@@ -162,7 +162,7 @@ impl fmt::Display for ScalarType {
 #[derive(Clone, Debug, Hash, PartialEq)]
 pub struct Vector {
     pub data_type: ScalarType,
-    pub length: u64,
+    pub length: usize,
 }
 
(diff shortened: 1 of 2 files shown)


notmgsk (approved):
I'm in favour, but will defer to @kalzoo

genos:
Closing (for now?) as this seems a non-issue. We’d like to be explicit about the integral size for the compiler, not have it tied to architecture specifics.

PR #19, opened 2021-07-28
Feature: Add proptest

genos:
*Note: I'm putting this one up as an early draft prior to sinking more time into it. If y'all like it, great! I'll keep going. If not, I'll be happy to throw away this PR & branch.*

[I've previously mentioned](https://github.com/rigetti/quil-rust/pull/11#discussion_r674219699) being interested in adding property-based testing using [`proptest`](https://github.com/AltSysrq/proptest) to do the heavy lifting for us. So far, this PR adds:

- two new `dev` dependencies, viz.
  - [`proptest`](https://github.com/AltSysrq/proptest) to do the testing,
  -  and [`paste`](https://github.com/dtolnay/paste) to do some macro naming tomfoolery
- one new testing macro called `roundtrip_proptest!` to assert that `parse_thing(lex(thing.to_string)) == thing`
- two examples of using `roundtrip_proptest!` to start, one for `parse_qubit` and one for `parse_variable_qubit`

The `roundtrip_proptest!` macro takes a name, a `proptest` generator of valid data, and a parsing function:

```rust
roundtrip_proptest!(thing, any<data>().prop_map(Thing::stuff), parse_thing);
```

and expands that into

```rust
mod thing_roundtrip {
    use super::*;
    use proptest::prelude::*;
    proptest! {
        #[test]
        fn roundtrip(datum in any<data>().prop_map(Thing::stuf)) {
            let string = datum.to_string();
            let lexed = lex(&string);
            let parsed = parse_thing(&lexed);
            prop_assert!(parsed.is_ok());
            prop_assert_eq!(datum, parsed.unwrap().1);
            }
        }
}
```

As an example, I found something that wasn't clear to me from the docstrings when running these tests. The docs for `parse_qubit` say that a qubit designator can be
> an integer (`1`), variable (`%q1`), or identifier (`q1`).
So I wrote my test:

```rust
roundtrip_proptest!(
    qubit,
    prop_oneof![
        any::<u64>().prop_map(Qubit::Fixed),
        "%[a-z][a-z0-9]+".prop_map(Qubit::Variable)
    ],
    parse_qubit
);
```
and it failed!

```
<SNIP>
thread 'parser::common::tests::qubit_roundtrip::roundtrip' panicked at 'Test failed: assertion failed: `(left == right)`
  left: `Variable("%a0")`,
 right: `Variable("a0")` at quil-rust/src/parser/common.rs:255; minimal failing input: datum = Variable("%a0")
<SNIP>
```

I didn't know that the lexer removes the percent symbol. Changing the generator for that case to `"[a-z][a-z0-9]+".prop_map(Qubit::Variable)` makes the test pass.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /target
 Cargo.lock
 node_modules
+tags
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,3 +15,5 @@ serde = { version = "1.0.125", features = ["derive"] }
 
 [dev-dependencies]
 insta = "1.7.1"
+paste = "1.0.5"
+proptest = "1.0.0"
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -31,3 +31,27 @@ macro_rules! imag {
         Complex64::new(0f64, $value)
     }};
 }
+
+/// Construct round-tripping test for Data -> String -> Parsed Data
+#[macro_export]
+macro_rules! roundtrip_proptest {
+    ($name:ident, $generator:expr, $parser:expr) => {
+        paste::paste! {
+            #[allow(unused_imports)]
+            mod [<$name _roundtrip>] {
+                use super::*;
+                use proptest::prelude::*;
+                proptest! {
+                    #[test]
+                    fn roundtrip(datum in $generator) {
+                        let string = datum.to_string();
+                        let lexed = lex(&string);
+                        let parsed = $parser(&lexed);
+                        prop_assert!(parsed.is_ok());
+                        prop_assert_eq!(datum, parsed.unwrap().1);
+                    }
+                }
+            }
+        }
+    };
+}
--- a/src/parser/common.rs
+++ b/src/parser/common.rs
@@ -223,9 +223,10 @@ pub fn skip_newlines_and_comments<'a>(input: ParserInput<'a>) -> ParserResult<'a
 
 #[cfg(test)]
 mod tests {
-    use crate::{expression::Expression, instruction::MemoryReference, parser::lex, real};
-
-    use super::parse_waveform_invocation;
+    use super::*;
+    use crate::{
+        expression::Expression, instruction::MemoryReference, parser::lex, real, roundtrip_proptest,
+    };
 
     #[test]
     fn waveform_invocation() {
@@ -250,4 +251,14 @@ mod tests {
             .collect()
         )
     }
+
+    roundtrip_proptest!(
+        qubit,
+        prop_oneof![
+            any::<u64>().prop_map(Qubit::Fixed),
+            "[a-z][a-z0-9]+".prop_map(Qubit::Variable)
+        ],
+        parse_qubit
+    );
+    roundtrip_proptest!(variable_qubit, "[a-z][a-z0-9]+", parse_variable_qubit);
 }


notmgsk:
Neato.

The docstr for `parse_qubit` is wrong. A qubit is an identifier (e.g. `q1`) or an integer (e.g. `1`). The `%`-prefix is strictly for circuit parameters.

genos:
In the latest two commits (6b8f3dd mainly, 75a28b3 is a short `cargo fmt` fix), I've simplified & renamed the roundtrip testing macro and added more tests. The format is now

```rust
parser_roundtrip!(parse_thing, generate_thing());
```

where `parse_thing` is a predefined parsing function you want to test, and `generate_thing` is a generator defined as in

```rust
fn generate_thing() -> impl Strategy<Value = Thing> {
  todo!()
}
```

See the bottom of `src/parser/common.rs` for examples. One can also put native `proptest` generators directly in that spot, but I opted for writing functions for legibility.

In running these, some test failures arose, either highlighting bugs in the code or (_much_ more likely) my own ignorance.

1. `parse_arithmetic_operand` fails to correctly (?) parse the string version of `ArithmeticOperand::LiteralReal(0.0)`, instead returning `ArithmeticOperand::LiteralInteger(0)`. Should we allow this to occur, in which case I should fix my test?
2. `parse_frame_identifier` fell to an even more interesting trap: it looks like the underlying `parse_qubit` fails if
  A. a `FrameIdentifier` has a `Qubit::Variable` as one of its qubits, and
  B. that qubit's name is the string version of a poorly behaved float, like `"inf"` or `"nan"`

This can be surfaced by a simpler unit test:

```rust
 #[test]
fn bad_variable_qubits() {
    let datum = Qubit::Variable("inf".to_string());
    let string = datum.to_string();
    let lexed = lex(&string);
    let parsed = parse_qubit(&lexed);
    assert_eq!(datum, parsed.unwrap().1);
}
```

fails with

```
Error(ExpectedToken { actual: Float(inf), expected: "$ expected_variant" } (at Float(inf)))
```

What do y'all think, dear reviewers? In **1** above, should parsing return an integer when the data started out as a real? In **2**, should I tweak the generator for variable qubits to filter out `"inf"` and `"nan"`?

notmgsk:
I'm not _totally_ sold on this type of testing... but it's already caught some bugs so maybe it is valuable. The current "bug" is that `0.0` is printed as `0` which is interpreted as a `LiteralInteger` -- does that constitute a bug? Wherever we expect a `LiteralReal` I think it's probably safe to allow a `LiteralInteger` too, and in this case nothing is lost.

genos:
@notmgsk totally understandable! I recognize I’m a fanboi of property testing (repeatable randomized generation plus automatic test case shrinking/minimization! 🤩), and I’ll happily hush up about it if the addition here seems more trouble than it’s worth.

As for the literal integer/real-ness of 0.0, bowing to subject matter expertise and adjusting the generation is absolutely an option. As is completely nuking this PR of course!

kalzoo:
This is cool, and I think it's something that would be great to have in the library! Adds new coverage with little code. As you note though, round-tripping might find a lot of formatting quirks rather than parsing bugs. I'd say that shouldn't bar this from inclusion right now, if the tests are green when it's merged.

PR #22, opened 2021-07-30
Breaking: Instruction variants are top-level structs

### LINKED ISSUE #21: Make all Instruction variants into the tuple type
Repeating a comment from @kalzoo here

> Currently, the instruction enum looks like this, using structured variants:
> 
> pub enum Instruction {
>     SomeVariant { field_1: bool, field_2: u64 },
>     ...
> }
> 
> After using llvm-ir, I've seen an advantage to using an alternative structure, which allows the data within a variant to be manipulated and stored independently of the Instruction enum itself. If we make this change, Instruction might look like this:
> 
> pub enum Instruction {
>     SomeVariant(SomeVariant),
>     ...
> }
> 
> pub struct SomeVariant {
>   field_1: bool,
>   field_2: u64
> }

Which I think is a good change to make. This unlocks the possibility of more directly typing instructions. For example, previously if you have a function that wanted a `Gate` it would have to take a `Instruction` and then match it with `Instruction::Gate` which is long-winded and unclear.

notmgsk:
This enables stuff like
```
fn compile_gate(gate: Gate) -> Option<Vec<Gate>> { ... }
```
where before you would've needed
```
fn compile_gate(gate: Instruction) -> Option<Vec<Instruction>> {
    match gate {
        Gate { .. } => {
            // ...
        }
        _ => // not a gate :(
    }
}
```

It does make some stuff more verbose, but generally I think that's constrained to library code, and not consumer code.

Closes #21

### DIFF
--- a/src/instruction.rs
+++ b/src/instruction.rs
@@ -261,9 +261,6 @@ pub struct Reset {
 }
 
 #[derive(Clone, Debug, PartialEq)]
-pub struct CalibrationDefinition(pub Calibration);
-
-#[derive(Clone, Debug, PartialEq)]
 pub struct Capture {
     pub blocking: bool,
     pub frame: FrameIdentifier,
@@ -424,7 +421,7 @@ pub enum Instruction {
     Declaration(Declaration),
     Measurement(Measurement),
     Reset(Reset),
-    CalibrationDefinition(CalibrationDefinition),
+    CalibrationDefinition(Calibration),
     Capture(Capture),
     Delay(Delay),
     Fence(Fence),
@@ -463,7 +460,7 @@ pub enum InstructionRole {
 impl From<&Instruction> for InstructionRole {
     fn from(instruction: &Instruction) -> Self {
         match instruction {
-            Instruction::CalibrationDefinition(CalibrationDefinition(_))
+            Instruction::CalibrationDefinition(_)
             | Instruction::CircuitDefinition(CircuitDefinition { .. })
             | Instruction::Declaration(Declaration { .. })
             | Instruction::FrameDefinition(FrameDefinition { .. })
@@ -564,7 +561,7 @@ impl fmt::Display for Instruction {
                 destination,
                 source,
             }) => write!(f, "{} {} {}", operator, destination, source),
-            Instruction::CalibrationDefinition(CalibrationDefinition(calibration)) => {
+            Instruction::CalibrationDefinition(calibration) => {
                 let parameter_str = get_expression_parameter_string(&calibration.parameters);
                 write!(
                     f,
--- a/src/instruction.rs
+++ b/src/instruction.rs
@@ -218,142 +218,238 @@ impl fmt::Display for MemoryReference {
 }
 
 #[derive(Clone, Debug, PartialEq)]
+pub struct Gate {
+    pub name: String,
+    pub parameters: Vec<Expression>,
+    pub qubits: Vec<Qubit>,
+    pub modifiers: Vec<GateModifier>,
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct CircuitDefinition {
+    pub name: String,
+    pub parameters: Vec<String>,
+    // These cannot be fixed qubits and thus are not typed as `Qubit`
+    pub qubit_variables: Vec<String>,
+    pub instructions: Vec<Instruction>,
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct GateDefinition {
+    pub name: String,
+    pub parameters: Vec<String>,
+    pub matrix: Vec<Vec<Expression>>,
+    pub r#type: GateType,
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct Declaration {
+    pub name: String,
+    pub size: Vector,
+    pub sharing: Option<String>,
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct Measurement {
+    pub qubit: Qubit,
+    pub target: Option<MemoryReference>,
+}
+
+#[derive(Clone, Debug, PartialEq)]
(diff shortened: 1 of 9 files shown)


src/instruction.rs
```diff
@@ -213,143 +213,239 @@ impl fmt::Display for MemoryReference {
...
+    pub parameters: Vec<Expression>,
+    pub qubits: Vec<Qubit>,
+    pub modifiers: Vec<GateModifier>,
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct CircuitDefinition {
+    pub name: String,
+    pub parameters: Vec<String>,
+    // These cannot be fixed qubits and thus are not typed as `Qubit`
```
genos, line 228: Would splitting the `Qubit` type into fixed & variable be useful? Or just more clutter?

src/program/graph.rs
```diff
@@ -21,7 +21,13 @@ use std::fmt;
 use petgraph::graphmap::GraphMap;
 use petgraph::Directed;
 
-use crate::instruction::{FrameIdentifier, Instruction, MemoryReference};
+use crate::instruction::{
```
genos, line 24: At this point, maybe `use crate::instruction::*`?

genos (approved):
LGTM, with a few questions for thought (no blockers whatsoever). I hope this was a largely mechanical process aided by the editor/IDE, and you didn't make all these changes by hand!

src/instruction.rs
```diff
@@ -363,38 +459,40 @@ pub enum InstructionRole {
...
-            | Instruction::SetScale { .. }
-            | Instruction::ShiftFrequency { .. }
-            | Instruction::ShiftPhase { .. }
-            | Instruction::SwapPhases { .. } => InstructionRole::RFControl,
-            Instruction::Arithmetic { .. }
-            | Instruction::Move { .. }
-            | Instruction::Exchange { .. }
-            | Instruction::Load { .. }
+            Instruction::CalibrationDefinition(CalibrationDefinition(_))
+            | Instruction::CircuitDefinition(CircuitDefinition { .. })
```
dbanty, line 463: If ignoring the interior, couldn't you still have this for brevity?

```suggestion
            Instruction::CalibrationDefinition(_)
            | Instruction::CircuitDefinition(_)
```

src/instruction.rs
```diff
@@ -213,143 +213,239 @@ impl fmt::Display for MemoryReference {
...
-        target: String,
-        condition: MemoryReference,
-    },
+    Gate(Gate),
+    CircuitDefinition(CircuitDefinition),
+    GateDefinition(GateDefinition),
+    Declaration(Declaration),
+    Measurement(Measurement),
+    Reset(Reset),
+    CalibrationDefinition(CalibrationDefinition),
```
dbanty, line 423: If this one was already containing a single field, should we just leave it as is? Seems like an unnecessary extra layer.

src/program/graph.rs
```diff
@@ -503,48 +509,54 @@ impl ScheduledProgram {
...
         for instruction in instructions {
             match instruction {
-                Instruction::Arithmetic { .. }
-                | Instruction::Capture { .. }
-                | Instruction::Delay { .. }
-                | Instruction::Fence { .. }
-                | Instruction::Move { .. }
-                | Instruction::Exchange { .. }
-                | Instruction::Load { .. }
+                Instruction::Arithmetic(Arithmetic { .. })
```
dbanty, line 512: Same note here as previously about the empty pattern `_`

src/instruction.rs
```diff
@@ -213,143 +213,239 @@ impl fmt::Display for MemoryReference {
...
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct Arithmetic {
+    pub operator: ArithmeticOperator,
+    pub destination: ArithmeticOperand,
+    pub source: ArithmeticOperand,
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct Halt;
```
dbanty, line 367: An empty type for an empty enum is definitely excessive. If the worry is about future compatibility, changing any one of these structs is a breaking change anyway because everything is pub.

dbanty (dismissed):
The main downside is (clearly) verbosity, but I don't think the current interface is particularly useful as a builder-style UX from Rust anyway.

I left a couple notes around where I think things could be less verbose without sacrificing usability.

dbanty:
Rebased on and updated for the changes on `main` with permission from Mark 🙂

src/instruction.rs
kalzoo: Since each just wraps a primitive type, I feel like this would just be clutter, but I'm not fixed on that

src/instruction.rs
```diff
@@ -213,143 +213,239 @@ impl fmt::Display for MemoryReference {
...
+    pub target: Option<MemoryReference>,
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct Reset {
+    pub qubit: Option<Qubit>,
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct CalibrationDefinition(pub Calibration);
```
kalzoo, line 260: Agreeing with what @dbanty wrote below; this does not need a double nesting. The reason it was like this was because I needed the functionality which we're extending here to all instructions; I needed to be able to pull out the contents and use them independently from `Instruction`.

`Calibration` can just become `CalibrationDefinition` here.

src/program/graph.rs
kalzoo: Not a fan of the implicitness/"magic" of `*`, and since `rust-analyzer` helps maintain the import list I'd say it's not bad

kalzoo (changes requested):

kalzoo (approved):
lgtm! Thanks @notmgsk and @dbanty !

PR #25, opened 2021-08-11
Upgrade: impl Hash and Eq for Expression for use in deduplication.

dbanty:
This is to allow the pyQuil style of expression de-duplication by substitution. Allowing for use in a HashSet directly feels a lot better than a map of Expression.to_string() => Expression, and this way we can tweak how we differentiate between expressions (e.g. evaluating/simplifying automatically) at some point if we want to.

Note that Clippy is rightly angry about implementing `Hash` on something which has derived `PartialEq`. This is ultimately because we store floats in expressions and floats are not hashable. That problem can also lead to issues with equality anyway so I just ignored for now. The recommendation is to create a `PartialEq` implementation by hand that is guaranteed to have the same results as `Hash`.

We probably also want to do a custom PartialEq/Hash at some point which allows for the commutative property in expressions (e.g. 1 + 2 == 2 + 1)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,3 +15,5 @@ serde = { version = "1.0.125", features = ["derive"] }
 
 [dev-dependencies]
 insta = "1.7.1"
+proptest = "1.0.0"
+proptest-derive = "0.3.0"
--- a/src/expression.rs
+++ b/src/expression.rs
@@ -17,6 +17,7 @@ use crate::{imag, instruction::MemoryReference, real};
 use std::collections::HashMap;
 use std::f64::consts::PI;
 use std::fmt;
+use std::hash::{Hash, Hasher};
 
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub enum EvaluationError {
@@ -43,6 +44,15 @@ pub enum Expression {
     },
     Variable(String),
 }
+
+#[allow(clippy::derive_hash_xor_eq)]
+impl Hash for Expression {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.to_string().hash(state);
+    }
+}
+
+impl Eq for Expression {}
 
 /// Compute the result of an infix expression where both operands are complex.
 fn calculate_infix(
--- a/src/expression.rs
+++ b/src/expression.rs
@@ -308,6 +308,7 @@ mod tests {
         real,
     };
     use num_complex::Complex64;
+    use std::collections::HashSet;
     use std::{collections::HashMap, f64::consts::PI};
 
     #[test]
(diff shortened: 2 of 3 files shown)


genos:
> We probably also want to do a custom PartialEq/Hash at some point which allows for the commutative property in expressions (e.g. 1 + 2 == 2 + 1)

Another case for property testing!

genos:
Could we consider bringing in the [`ordered_float`](https://docs.rs/crate/ordered-float/2.7.0) crate? It provides us "A wrapper around floats providing implementations of `Eq`, `Ord`, and `Hash`," so we could rely on `derive` to get them for Expression.

kalzoo:
> We probably also want to do a custom PartialEq/Hash at some point which allows for the commutative property in expressions (e.g. 1 + 2 == 2 + 1)

Definitely. FWIW in Python we effect this by use of a function which makes these expressions "canonical" by reordering operands in a deterministic way.

kalzoo:
Also - would you mind adding simple tests to assert the expected behavior?

genos:
@kalzoo et. al. what do you think about adding:

- `ordered_float` to the main dependencies to wrap our floats and give us this behavior
- `proptest` to the dev dependencies to give us property testing for things like commutativity

?

dbanty:
> Could we consider bringing in the [`ordered_float`](https://docs.rs/crate/ordered-float/2.7.0) crate? It provides us "A wrapper around floats providing implementations of `Eq`, `Ord`, and `Hash`," so we could rely on `derive` to get them for Expression.

@genos so we would store `Complex<OrderedFloat<64>>` instead of `Complex<64>`? Definitely will be worth it to add some helpers to construct from 1 or 2 floats in that case.

dbanty:
> Great callout though - could you add an issue for this?

Added #27

dbanty:
> Also - would you mind adding simple tests to assert the expected behavior?

Done!

kalzoo:
@genos I’m not sure of all the implications of using a custom type in place of a primitive. It feels like something we’d want to avoid, but if others are in favor we could see how it goes.

A compromise option could be to just add its trait implementations on Expression instead of the float value itself.

I’d be happy with proptest, but ideally in another PR (maybe revive & rebase the previous one)

dbanty:
> @dbanty yup, we could adapt the [macros](https://github.com/rigetti/quil-rust/blob/main/src/macros.rs) for the helpers.

We could also do something like:

```rust
impl From<f64> for Expression {}
```

so you can just do Expression::from(1.0). I'd really like to do `FromStr` for `Expression` as well in the future to make testing easier, but that's going to involve exposing some parsing stuff where it isn't today.

dbanty:
> A compromise option could be to just add its trait implementations on Expression instead of the float value itself.

I think we need custom `PartialEq` and `Hash` impls anyway in order to make commutation work, so that might eliminate some of the benefits of a custom type. I'd say let's do it ourselves and if it makes sense to pull in helpers to make it less cumbersome then we can do that as well.

There's also [float_cmp](https://docs.rs/float-cmp/0.9.0/float_cmp/) and similar which we could use ourselves if we wanted to.

genos:
> I think we need custom PartialEq and Hash impls anyway in order to make commutation work, so that might eliminate some of the benefits of a custom type. I'd say let's do it ourselves and if it makes sense to pull in helpers to make it less cumbersome then we can do that as well.

Excellent point; agreed!

genos:
> I’d be happy with proptest, but ideally in another PR (maybe revive & rebase the previous one)

The previous PR didn't have much beyond the roundtripping (with which we were unimpressed). I think it would work nicely for handling #27, though so I'll keep it in mind for that once this this PR merges.

genos:
> FWIW in Python we effect this by use of a function which makes these expressions "canonical" by reordering operands in a deterministic way.

Is that why `pyquil` things test equality by checking for same class + same `self.out()` values?

genos:
> FWIW in Python we effect this by use of a function which makes these expressions "canonical" by reordering operands in a deterministic way.

@kalzoo Where is this? I seem to be able to get the same problem with `pyquil`:

```python
>>> from pyquil.quilatom import Add
>>> Add(1, 2) == Add(2, 1)
False
```

genos:
See #29 for merging into _here_ (**not** `main` yet!) the requested commutativity handling from #27.

genos:
Just merged #29 into here

genos:
*Pinging*

Any movement on this?

kalzoo:
```
PING kalzoo: 3 PRs
64 bytes from kalzoo: icmp_seq=0 ttl=big time=2 hours
```

> @kalzoo Where is this? I seem to be able to get the same problem with pyquil:

What I was thinking of is in internal code, and also to my surprise does not actually handle commutation. tl;dr disregard what I said

Otherwise - just reviewed it - I like it! A quality change, especially with proptest in there.

kalzoo (approved):

genos:
Huzzah! Thanks @kalzoo!

@dbanty, would you like to do the honors and 🚨 **push the button** 🚨 ?

PR #26, opened 2021-08-11
New: Support parsing SET-SCALE, SET-FREQUENCY, and SHIFT-FREQUENCY.

dbanty:
For the two frequency commands, [the spec] agrees with the previous structs in that they should take floats, but [pyQuil] allows expressions and does conversion on them, so I assume we should stick with that implementation.

[the spec]: https://github.com/quil-lang/quil/blob/master/rfcs/analog/spec_changes.md#frequency
[pyQuil]: https://github.com/rigetti/pyquil/blob/9545776eff57362aed18b619a36182c556b2296f/pyquil/api/_rewrite_arithmetic.py#L107

### DIFF
--- a/src/parser/command.rs
+++ b/src/parser/command.rs
@@ -31,9 +31,7 @@ use super::{
 use crate::parser::common::parse_variable_qubit;
 use crate::parser::instruction::parse_block;
 use crate::{
-    instruction::{
-        ArithmeticOperand, ArithmeticOperator, Calibration, FrameIdentifier, Instruction, Waveform,
-    },
+    instruction::{ArithmeticOperand, ArithmeticOperator, Calibration, Instruction, Waveform},
     parser::{
         error::{Error, ErrorKind},
         lexer::Token,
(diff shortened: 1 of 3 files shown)


genos (approved):
👍

dbanty:
Thanks for the quick review @genos! I dismissed because I shoved a couple extra commands into this PR since they're closely related 😅

src/parser/instruction.rs
```diff
@@ -465,4 +467,85 @@ mod tests {
             source: ArithmeticOperand::LiteralReal(1.0)
         }]
     );
+
+    #[test]
+    fn parse_set_scale() {
+        let tokens = lex(r#"SET-SCALE 0 "rf" 1.0; SET-SCALE 0 1 "rf" theta"#);
```
kalzoo, line 473: I should have used raw strings everywhere

src/parser/instruction.rs
```diff
@@ -465,4 +467,85 @@ mod tests {
...
+        let (remainder, parsed) = parse_instructions(&tokens).unwrap();
+        let expected = vec![
+            Instruction::SetScale {
+                frame: FrameIdentifier {
+                    name: String::from("rf"),
+                    qubits: vec![Qubit::Fixed(0)],
+                },
+                scale: Expression::Number(Complex64::from(1.0)),
```
kalzoo, line 481: I've defined handy `real` and `imag` macros in this crate to cut down on a little bit of verbosity when instantiating `Complex64`s. I don't they should be exported, but they're useful to keep at the crate level and use in spots like this. What do you think?

src/parser/command.rs
```diff
@@ -323,16 +321,15 @@ pub fn parse_pragma<'a>(input: ParserInput<'a>) -> ParserResult<'a, Instruction>
 }
 
 /// Parse a pulse instruction, **including the PULSE keyword**.
```
kalzoo, line 323: My fault - but this comment is out of date and should read like all the others. PULSE keyword is no longer included.

kalzoo (approved):

src/parser/command.rs
dbanty: Fixed!

src/parser/instruction.rs
dbanty: Nice, swapped that out.

PR #28, opened 2021-08-13
New: Parse `SET-PHASE` and `SHIFT-PHASE`.

dbanty:
A lot of these instructions are parsed exactly the same, so it feels kinda silly to have so many different functions to do it. Not sure how to deduplicate without just making it harder to use though so 🤷

### DIFF
--- a/src/parser/command.rs
+++ b/src/parser/command.rs
@@ -360,6 +360,14 @@ pub fn parse_set_frequency(input: ParserInput) -> ParserResult<Instruction> {
     Ok((input, Instruction::SetFrequency { frame, frequency }))
 }
 
+/// Parse the contents of a `SET-PHASE` instruction.
+pub fn parse_set_phase(input: ParserInput) -> ParserResult<Instruction> {
+    let (input, frame) = parse_frame_identifier(input)?;
+    let (input, phase) = parse_expression(input)?;
+
+    Ok((input, Instruction::SetPhase { frame, phase }))
+}
+
 /// Parse the contents of a `SET-SCALE` instruction.
 pub fn parse_set_scale(input: ParserInput) -> ParserResult<Instruction> {
     let (input, frame) = parse_frame_identifier(input)?;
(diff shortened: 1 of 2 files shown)


genos (approved):
👍

Agreed about it not being the DRY-est code, but I don't see a great way around it.

src/parser/command.rs
```diff
@@ -360,6 +360,14 @@ pub fn parse_set_frequency(input: ParserInput) -> ParserResult<Instruction> {
     Ok((input, Instruction::SetFrequency { frame, frequency }))
 }
 
+/// Parse the contents of a `SET-PHASE` instruction.
+pub fn parse_set_phase(input: ParserInput) -> ParserResult<Instruction> {
+    let (input, frame) = parse_frame_identifier(input)?;
+    let (input, phase) = parse_expression(input)?;
```
notmgsk, line 366: I'm not totally sure why, but the [Quil-T proposal](https://github.com/quil-lang/quil/blob/master/rfcs/analog/spec_changes.md#phase) requires that the phase be an immediate float (i.e. not a memory reference which `parse_expression` might return).

notmgsk (changes requested):

notmgsk:
_DRY_ but also _clear_ is better than _clever_.

src/parser/command.rs
dbanty: Yeah, the current master branch disagrees with that as well as expressions in set/shift frequency. There is an open but stale PR (https://github.com/quil-lang/quil/pull/46/files) which seeks to address it.

I'm writing under the assumption that the pyQuil implementation is correct, which expects expressions in all three of those locations. I suppose `quilc` should be the real arbiter of truth here...

kalzoo: Yes, best to model around pyQuil for now until the Quil spec is fixed.

kalzoo (approved):

src/parser/instruction.rs
```diff
@@ -466,6 +466,33 @@ mod tests {
...
     );
 
+    #[test]
+    fn parse_set_phase() {
+        let tokens = lex(r#"SET-PHASE 0 "rf" 1.0; SET-PHASE 0 1 "rf" theta"#);
+        let (remainder, parsed) = parse_instructions(&tokens).unwrap();
+        let expected = vec![
+            Instruction::SetPhase {
+                frame: FrameIdentifier {
+                    name: String::from("rf"),
```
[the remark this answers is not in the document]
dbanty: Ah yes, the ol' "how to create a `String` from a `str`" conversation 😆. The main options are, I believe:

1. `String::from("str")`
2. `"str".to_owned()`
3. `"str".to_string()`

I'm definitely guilty of switching between them a bit, but I prefer the first one because it's more obvious when skimming that you've just allocated some new memory on the heap- which makes it easier to make tiny optimizations later. [The book](https://doc.rust-lang.org/stable/book/ch04-01-what-is-ownership.html#the-string-type), [Rust By Example](https://doc.rust-lang.org/stable/rust-by-example/std/str.html#strings), [the `String` docs](https://doc.rust-lang.org/stable/std/string/struct.String.html), and [the `str` docs](https://doc.rust-lang.org/std/primitive.str.html#method.get) all default to `String::from`, so that seems to be the recommended method.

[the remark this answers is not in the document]
dbanty: I created an internal Runbook issue to track adding to our internal style guide.

PR #30, opened 2021-08-13
Evaluate patched memory addresses

dbanty:
Two changes in here:

## Breaking: `Expression::evaluate` and `evaluate_to_complex` now take in a `patch_values` parameter.

`patch_values` allows for evaluating expressions that contain memory references in the same format used to send executables to QPUs.

##  New: Implement `FromStr` for `Expression`.

Enables creation of expressions from strings for easier testing (e.g. `Expression::from_str("theta" + "beta")`).

## Potential Issues

1. Obviously this is a breaking change, so downstream consumers will have to deal with that.
2. The error messaging on `FromStr` isn't great, but I think we intend on revamping errors overall in this crate so I didn't pull in `thiserror` or anything for it.

### DIFF
--- a/src/expression.rs
+++ b/src/expression.rs
@@ -13,10 +13,13 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  **/
-use crate::{imag, instruction::MemoryReference, real};
 use std::collections::HashMap;
 use std::f64::consts::PI;
 use std::fmt;
+use std::str::FromStr;
+
+use crate::parser::{lex, parse_expression};
+use crate::{imag, instruction::MemoryReference, real};
 
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub enum EvaluationError {
@@ -81,14 +84,18 @@ impl Expression {
     /// Consume the expression, simplifying it as much as possible using the values provided in the environment.
     /// If variables are used in the expression which are not present in the environment, evaluation stops there,
     /// returning the possibly-simplified expression.
-    pub fn evaluate(self, environment: &EvaluationEnvironment) -> Self {
+    pub fn evaluate(
+        self,
+        environment: &EvaluationEnvironment,
+        patch_values: Option<&HashMap<&str, Vec<f64>>>,
+    ) -> Self {
         use Expression::*;
         match self {
             FunctionCall {
                 function,
                 expression,
             } => {
-                let evaluated = (*expression).evaluate(environment);
+                let evaluated = (*expression).evaluate(environment, patch_values);
                 match &evaluated {
                     Number(value) => Number(calculate_function(&function, value)),
                     PiConstant => Number(calculate_function(&function, &real!(PI))),
@@ -103,8 +110,8 @@ impl Expression {
                 operator,
                 right,
             } => {
-                let left_evaluated = (*left).evaluate(environment);
-                let right_evaluated = (*right).evaluate(environment);
+                let left_evaluated = (*left).evaluate(environment, patch_values);
+                let right_evaluated = (*right).evaluate(environment, patch_values);
 
                 match (&left_evaluated, &right_evaluated) {
                     (Number(value_left), Number(value_right)) => {
@@ -143,7 +150,17 @@ impl Expression {
                 Some(value) => Number(*value),
                 None => Variable(identifier),
             },
-            _ => self,
+            Address(memory_reference) => {
+                let number = patch_values.and_then(|patch_values| {
+                    let values = patch_values.get(memory_reference.name.as_str())?;
+                    let value = values.get(memory_reference.index as usize)?;
+                    Some(real!(*value))
+                });
+
+                number.map_or(Address(memory_reference), Number)
+            }
+            PiConstant => PiConstant,
+            Number(number) => Number(number),
         }
     }
 
@@ -152,10 +169,11 @@ impl Expression {
     pub fn evaluate_to_complex(
         self,
         environment: &EvaluationEnvironment,
+        patch_values: Option<&HashMap<&str, Vec<f64>>>,
     ) -> Result<num_complex::Complex64, EvaluationError> {
         use Expression::*;
 
-        let result = self.evaluate(environment);
+        let result = self.evaluate(environment, patch_values);
         match result {
             Number(value) => Ok(value),
             PiConstant => Ok(real!(PI)),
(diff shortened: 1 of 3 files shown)


src/expression.rs
```diff
@@ -13,10 +13,13 @@
...
  * limitations under the License.
  **/
-use crate::{imag, instruction::MemoryReference, real};
 use std::collections::HashMap;
 use std::f64::consts::PI;
 use std::fmt;
+use std::str::FromStr;
+
+use crate::parser::{lex, parse_expression};
+use crate::{imag, instruction::MemoryReference, real};
```
genos, line 22: 👍 Reminds me of the `flake-import-order` setup for Python: standard imports before third-party before local. I wonder if we can turn this into a `clippy` lint?

src/expression.rs
```diff
@@ -81,14 +84,18 @@ impl Expression {
     /// Consume the expression, simplifying it as much as possible using the values provided in the environment.
     /// If variables are used in the expression which are not present in the environment, evaluation stops there,
     /// returning the possibly-simplified expression.
-    pub fn evaluate(self, environment: &EvaluationEnvironment) -> Self {
+    pub fn evaluate(
+        self,
+        environment: &EvaluationEnvironment,
+        patch_values: Option<&HashMap<&str, Vec<f64>>>,
```
genos, line 90: I don't normally like optional-containers-of-possibly-empty-things (does `None` mean something different from `empty`?), but I think this is about as clean as it can be without Rust allowing default keyword arguments; passing an empty `HashMap::new()` instead of `None` wouldn't exactly be zero-cost.

src/expression.rs
```diff
@@ -143,7 +150,17 @@ impl Expression {
                 Some(value) => Number(*value),
                 None => Variable(identifier),
             },
-            _ => self,
+            Address(memory_reference) => {
+                let number = patch_values.and_then(|patch_values| {
```
genos, line 154: Maybe don't shadow the `patch_values` name inside the `and_then`? Matter of taste more than anything.

genos (commented):
Nice! A few comments/questions here & there, but nothing blocking.

Should we put a "more robust error handling" issue on the backlog?

src/expression.rs
dbanty: JetBrains IDEs do that for you 😁. Looks like rustfmt can be configured to do it, just doesn't by default, search for "group_imports": https://rust-lang.github.io/rustfmt/

src/expression.rs
dbanty: That is why, it's just because `None` is easier to type and also zero-cost.

src/expression.rs
dbanty: Yeah I don't like shadowing either but I struggle with what to call `Option<Something>` once I've verified that it's `Some(something)`.

dbanty:
@genos I've stumbled across another issue, the way I want to use evaluate, it doesn't make sense to consume the expressions because I need to recalculate for multiple sets of parameters and cloning feels wasteful. I'm trying to sort out if making this take a `&self` makes any kind of sense. I'm gonna try and fiddle with the API to both keep the "consume this expression and simplify it" behavior and prevent cloning when just asking to evaluate or fail. Totally unsure how that's going to work 😅.

src/expression.rs
genos: How about just `pv`? Small variable names seem less of an issue if it’s referring to the inner thing.

src/expression.rs
genos: I defo get it! I’m just still scarred from working with a Scala `Option[Seq[Option[String]]]` at a previous job 😭

dbanty:
Alright @genos & @kalzoo—tell me what you think about this update. `simplify` only simplifies complex expressions into more basic ones with no additional context. `evaluate` only checks if the expression can be evaluated when provided context without any simplification happening.

In theory, this makes evaluation faster because it won't be trying to simplify as part of the process _and_ allows it to be called multiple times on an expression without cloning.

The main downside I see to this change is that you can't partially evaluate an expression, I'm not sure if that is a capability that's being used anywhere that will be missed.

genos:
> The main downside I see to this change is that you can't partially evaluate an expression

Aw man, I was going to throw the Yoneda lemma at this (j/k j/k)

src/expression.rs
```diff
@@ -75,91 +78,123 @@ fn calculate_function(
...
 }
 
-pub type EvaluationEnvironment = HashMap<String, num_complex::Complex64>;
-
 impl Expression {
     /// Consume the expression, simplifying it as much as possible using the values provided in the environment.
-    /// If variables are used in the expression which are not present in the environment, evaluation stops there,
-    /// returning the possibly-simplified expression.
-    pub fn evaluate(self, environment: &EvaluationEnvironment) -> Self {
+    pub fn simplify(self) -> Self {
```
genos, line 83: Probably want to update this docstring

src/expression.rs
```diff
@@ -75,91 +78,123 @@ fn calculate_function(
...
-                    PiConstant => Number(calculate_function(&function, &real!(PI))),
-                    _ => FunctionCall {
-                        function,
-                        expression: Box::new(evaluated),
-                    },
-                }
-            }
+            } => FunctionCall {
+                function,
+                expression: Box::new(expression.simplify()),
```
genos, line 92: Maybe include the `calculate_function(&function, expression.simplify())` in here?

genos: As in: if the expression is/simplifies to a number or π

genos (changes requested):
I like the separation of concerns a lot!

A few comments/suggestions here and there. Most importantly: I think we probably still want the `calculate_function` inside `simplify` for those instances where it can be simplified.

src/expression.rs
dbanty: Do Scala Options implement from iterator? Meaning, can you you do like in Rust and easily turn `Vec<Option<Thing>>` into `Option<Vec<Thing>>`?

Between that and `map`/`and_then` I rarely have issues making Optional code both readable and efficient 😁.

In any event, the refactor of `evaluate`/`simplify` eliminated this `Option`.

genos: It does have a `flatMap`, which means you can use them in `for` comprehensions (FYI `for` in Scala is a monad comprehension, _not_ a for loop!) but the standard library lacks the traverse/sequence that [`Futures`](https://www.scala-lang.org/api/2.13.5/scala/concurrent/Future$.html) have. Luckily, the [`cats`](https://typelevel.org/cats/) library adds that good stuff in, but that leads to completely different programming idioms.

Screen Shot 2021-08-19 at 9 43 50 PM

genos (approved):
👍

PR #37, opened 2022-03-07
Fix: `Program::from_str` and `Expression::from_str` will no longer panic on bad input.

dbanty:


dbanty:
cargo-deny issues fixed in #38

kalzoo (approved):

PR #40, opened 2022-03-07
Fix: DEFWAVEFORM and waveform name parsing

kalzoo:
This alters quil-rust to match the pyQuil "defacto" spec and support what's present in live calibrations returned from Rigetti QCS.

While this is technically a breaking change, the fact is that the existing parser could not have been effectively used with any live systems; thus we consider it a bug.


genos (approved):

dbanty (approved):
LGTM though I have not independently verified the pyQuil vs Quil spec thing. Agreed re: bug vs breaking change.

PR #42, opened 2022-03-07
Update: add ScheduleError#instruction_index

### LINKED ISSUE #41: Include program instruction index in ScheduleError
Currently, diagnosing the cause of a `ScheduleError` is difficult given that the instruction which causes it may be present anywhere in a long program.

AC:
- `ScheduleError` has an `instruction_index` field which (optionally) contains the index of the instruction which caused an error. This field should be `None` when this information is not relevant.

kalzoo:
Closes #41


genos:
👍

dbanty (approved):

PR #47, opened 2022-03-07
Fix: lex keywords from identifiers

### LINKED ISSUE #44: Bug: Identifiers which include keywords are incorrectly lexed
Case in point: `LOAD-MEMORY`  lexed as the keyword `LOAD`, a minus operator, and then an identifier `MEMORY`. It should be lexed as a single identifier. This is due to the ordering of the lexer combinators.

To avoid this, because all keywords are themselves valid identifiers, we should only attempt to lex keywords from completed identifiers rather than from the raw string. While we do not yet have benchmarking (#36 ), I'd expect that this will also significantly speed up lexing, because we do not need to branch on all possible keywords at the beginning of each token.

kalzoo:
Closes #44


src/parser/lexer.rs
```diff
@@ -189,69 +187,65 @@ fn lex_comment(input: &str) -> LexResult {
...
+        "SET-FREQUENCY" => Some(SetFrequency),
+        "SET-PHASE" => Some(SetPhase),
+        "SET-SCALE" => Some(SetScale),
+        "SHIFT-FREQUENCY" => Some(ShiftFrequency),
+        "SHIFT-PHASE" => Some(ShiftPhase),
+        "LABEL" => Some(Label),
+        _ => None,
+    };
+
+    match command {
```
genos, line 245: Probably fine, but shadowing the word `command` here took me a second to mentally parse

src/parser/lexer.rs
```diff
@@ -272,8 +266,10 @@ fn lex_identifier_raw(input: &str) -> IResult<&str, String> {
     )(input)
 }
 
-fn lex_identifier(input: &str) -> LexResult {
-    lex_identifier_raw(input).map(|(input, ident)| (input, Token::Identifier(ident)))
+fn lex_command_or_identifier(input: &str) -> LexResult {
+    let (input, identifier) = lex_identifier_raw(input)?;
+    let token = recognize_command_or_identifier(identifier);
+    Ok((input, token))
```
dbanty, line 272:
```suggestion
    let (remainder, identifier) = lex_identifier_raw(input)?;
    let token = recognize_command_or_identifier(identifier);
    Ok((remainder, token))
```

Just assuming that that's what that function is returning because of how nom usually works. More obvious to me if that's the case (vs returning the same thing as the input).

src/parser/lexer.rs
```diff
@@ -189,69 +187,65 @@ fn lex_comment(input: &str) -> LexResult {
...
+        "SHIFT-FREQUENCY" => Some(ShiftFrequency),
+        "SHIFT-PHASE" => Some(ShiftPhase),
+        "LABEL" => Some(Label),
+        _ => None,
+    };
+
+    match command {
+        Some(command) => Token::Command(command),
+        None => Token::Identifier(identifier),
+    }
```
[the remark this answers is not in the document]
dbanty: Hehe 😅. Can you tell what kind of projects I've been working on recently?

src/parser/lexer.rs
kalzoo: Yup, see my response on #48 - but tl;dr not a bad idea but I'll favor consistency for the time being, and `input` is used consistently across the parser here.

src/parser/lexer.rs
```diff
@@ -189,69 +187,65 @@ fn lex_comment(input: &str) -> LexResult {
     Ok((input, Token::Comment(content.to_owned())))
 }
 
-fn lex_instruction(input: &str) -> LexResult {
+// If the given identifier string matches a command keyword, return the keyword;
+// otherwise, return the original identifier as a token.
```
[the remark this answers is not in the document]
kalzoo: whoops, thanks

src/parser/lexer.rs
kalzoo: True - agreed that this was confusing, applied a simple fix to use a single `match`.

kalzoo:
Thank you to @genos and @dbanty for the reviews! Merging now but happy to apply retroactive fixes for anything else you see.

PR #49, opened 2022-03-07
feat: update program to use btreemap for deterministic ordering

nilslice:
This was causing snapshot test failures when a program was serialized and compared to a cached version of itself from a prior run. Specifically, the memory regions of a program would serialize out-of-order & while that should not have any impact on the program itself, it creates false-positives when testing. 

The use of a `HashMap` is replaced with a `BTreeMap` throughout a `Program`, including the `waveforms` field. The test included in this PR does not necessarily help differentiate or prove the new behavior compared to the `HashMap` version, but it does help to ensure that we see the expected behavior.

This change fixes the broken test (in another repo) which lead to this PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -324,7 +324,7 @@ checksum = "a993555f31e5a609f617c12db6250dedcac1b0a85076912c436e6fc9b2c8e6a3"
 
 [[package]]
 name = "quil-rs"
-version = "0.8.1"
+version = "0.8.2"
 dependencies = [
  "indexmap",
  "insta",
--- a/src/program/mod.rs
+++ b/src/program/mod.rs
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  **/
-use std::collections::HashMap;
+use std::collections::BTreeMap;
 use std::str::FromStr;
 
 use crate::instruction::{
@@ -41,8 +41,8 @@ mod memory;
 pub struct Program {
     pub calibrations: CalibrationSet,
     pub frames: FrameSet,
-    pub memory_regions: HashMap<String, MemoryRegion>,
-    pub waveforms: HashMap<String, Waveform>,
+    pub memory_regions: BTreeMap<String, MemoryRegion>,
+    pub waveforms: BTreeMap<String, Waveform>,
     pub instructions: Vec<Instruction>,
 }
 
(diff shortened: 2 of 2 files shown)


genos (commented):
Nice catch! LGTM, except for the unused imports in L243 (in `tests`).

.github/workflows/release.yml
```diff
@@ -28,6 +28,12 @@ jobs:
           profile: minimal
           override: true
 
+      - name: Update Cargo.lock
+        uses: actions-rs/cargo@v1
+        with:
+          command: update
+          args: -w
+
```
dbanty, line 36: The version in `Cargo.toml` won't actually be updated until `npx semantic-release` has run. Likewise, the `git commit` which consumes the list of files in `assets` in `package.json` is run _within_ `semantic-release`, so moving this command to after that one wouldn't have the desired effect either.

I _think_ what needs to happen is putting this command inline in the `@semantic-release/exec` plugin.

package.json
```diff
@@ -10,32 +10,68 @@
...
+        "@semantic-release/release-notes-generator",
+        {
+          "preset": "eslint"
+        }
+      ],
+      [
+        "@semantic-release/exec",
+        {
+          "verifyConditionsCmd": "semantic-release-rust verify-conditions",
+          "prepareCmd": "semantic-release-rust prepare ${nextRelease.version}",
```
dbanty, line 59: This is almost-definitely what is bumping the version in `Cargo.toml`. Maybe `semantic-release-rust` has an option for updating the `.lock` file? If not, this is just a bash script, so we could do this:

```suggestion
          "prepareCmd": "semantic-release-rust prepare ${nextRelease.version} && cargo update -w",
```

Or, as we've done in all [our Python projects](https://github.com/rigetti/pyquil/blob/497791e8108d8780109d75410be786c5f6e590ea/package.json#L29) create a `prepare_release.sh` and invoke that here.

dbanty (changes requested):

package.json
nilslice: Good call -- I'm good inlining this at first and extracting out to a script if it gets more complex down the line. Thanks!

genos: Looks like `semantic-release-rust` only touches the `Cargo.toml`, not the lockfile. What's more, the documentation has a number of unfortunate typos 😬

So! I support the `&& cargo update -w` maneuver!

kalzoo (approved):

Issue #36, opened 2022-03-03
Benchmark vs existing parser


kalzoo:
We can't currently gauge relative performance between this library and the two existing parsers:

- `quilc`
- `pyQuil`

To ensure a good UX in Rust, we should benchmark parsing in this library in an analogous way (i.e. a similar program) to one of those parsers and compare the results.

Results here should be stored in a way that allows us to view results by git ref and thus evolution/regression over time. This may look like a CI task to commit automated results back into the repository.

`quilc` has a benchmarking script [here](https://github.com/quil-lang/quilc/blob/master/benchmarking/quilc-perf.lisp). `pyQuil` does [not](https://github.com/rigetti/pyquil/blob/master/test/unit/test_parser.py), but could use one.

We only need to benchmark against one of these existing libraries; which one is the implementer's choice. Ensure that the same program(s) are used under the same conditions for the purposes of comparison.

kalzoo:
Given that the grammar supported by pyQuil is closer to the intent of this library (namely Quil-T but not QASM), it may make sense to compare to pyQuil.

genos:
If possible, I think it'd be great to use [`criterion.rs`](https://github.com/bheisler/criterion.rs) for this.

genos:
Alternatively/in addition, since the built-in benchmark tests are on unstable only, we could use [`bencher`](https://github.com/bluss/bencher) and maybe compare via [`benchcmp`](https://github.com/BurntSushi/cargo-benchcmp).

nilslice:
@genos - `criterion` would be nice, and was the first thing that came to mind. I need to take a closer look at the benchmarks in `quilc` as well as what is available in the python parser to get a better idea of how they can all be compared against one another in some programmatic way. 

Manually observing these results is fine for the time being though (@kalzoo check me on this though if you disagree!)... and so we defer anything fancy for a while. I think we just want to know we are in the same ballpark and take note of any major regressions if they occur. 

Initially I was thinking I could execute the benchmarks for lisp & python from a `criterion` bench, but am concerned about the non-negligible time recorded by the framework to shell out or otherwise spin up the other language benches.

genos:
[...]

PR #53, opened 2022-03-09
Update: add automated benchmarks for parsing quilc test programs

nilslice:
Partially satisfying #36, this PR adds automated benchmarks running `criterion.rs` harness to parse several `quil` programs used as tests in the `quilc` repository. `quilc` has been included as a git submodule of this repository.

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -4,11 +4,6 @@ on: [ pull_request ]
 
 name: Benchmark
 
-env:
-  MAIN_SHA: $(git rev-parse main)
-  MAIN_BENCH: ${{ env.MAIN_SHA:0:7 }}.bench
-  THIS_BENCH: ${{ GITHUB_SHA:0:7 }}.bench
-
 jobs:
   benchmark:
     name: Run benchmarks
@@ -18,6 +13,11 @@ jobs:
         rust:
           - stable
     steps:
+      - name: Checkout sources
+        uses: actions/checkout@v2
+        with:
+          submodules: recursive
+
       - name: Set env vars
         run: |
           THIS_SHA=$(echo $GITHUB_SHA | cut -c 1-7)
@@ -25,16 +25,6 @@ jobs:
 
           echo "THIS_BENCH=${THIS_SHA}.bench" >> $GITHUB_ENV
           echo "MAIN_BENCH=${MAIN_SHA}.bench" >> $GITHUB_ENV
-
-      - name: Test env vars
-        run: |
-          echo "MAIN_BENCH=${MAIN_BENCH}"
-          echo "THIS_BENCH=${THIS_BENCH}"
-
-      - name: Checkout sources
-        uses: actions/checkout@v2
-        with:
-          submodules: recursive
 
       - name: Install toolchain
         uses: actions-rs/toolchain@v1
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -17,6 +17,7 @@ jobs:
         uses: actions/checkout@v2
         with:
           submodules: recursive
+          fetch-depth: 0
 
       - name: Set env vars
         run: |
@@ -31,6 +32,8 @@ jobs:
         with:
           toolchain: ${{ matrix.rust }}
           override: true
+        
+      - uses: Swatinem/rust-cache@v1
         
       - name: Install critcmp
         run: cargo install critcmp
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -1,0 +1,73 @@
+# Based on https://github.com/actions-rs/meta/blob/master/recipes/msrv.md
+
+on: [ pull_request ]
+
+name: Benchmark
+
+env:
+  MAIN_SHA: $(git rev-parse main)
+  MAIN_BENCH: ${{ env.MAIN_SHA:0:7 }}.bench
+  THIS_BENCH: ${{ GITHUB_SHA:0:7 }}.bench
+
+jobs:
+  benchmark:
+    name: Run benchmarks
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        rust:
+          - stable
+    steps:
+      - name: Set env vars
+        run: |
+          THIS_SHA=$(echo $GITHUB_SHA | cut -c 1-7)
+          MAIN_SHA=$(echo $(git rev-parse main) | cut -c 1-7)
+
+          echo "THIS_BENCH=${THIS_SHA}.bench" >> $GITHUB_ENV
+          echo "MAIN_BENCH=${MAIN_SHA}.bench" >> $GITHUB_ENV
+
+      - name: Test env vars
+        run: |
+          echo "MAIN_BENCH=${MAIN_BENCH}"
+          echo "THIS_BENCH=${THIS_BENCH}"
+
+      - name: Checkout sources
+        uses: actions/checkout@v2
+        with:
+          submodules: recursive
+
+      - name: Install toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: ${{ matrix.rust }}
+          override: true
+        
+      - name: Install critcmp
+        run: cargo install critcmp
+
+      - name: Run cargo check
+        uses: actions-rs/cargo@v1
+        with:
+          command: check
+
+      # - name: Download benchmark artifacts
+      #   uses: actions/download-artifact@v3
+      #   with:
+      #     name: benchmarks
+      
+      - name: Run cargo benchmark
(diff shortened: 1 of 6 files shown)


.github/workflows/bench.yml
```diff
@@ -0,0 +1,63 @@
...
+          echo "MAIN_BENCH=${MAIN_SHA}.bench" >> $GITHUB_ENV
+
+      - name: Install toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: ${{ matrix.rust }}
+          override: true
+        
+      - name: Install critcmp
+        run: cargo install critcmp
```
dbanty, line 36: I would add the `rust-cache` action in, it should cache cargo binaries as well as dependencies to really speed things along.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "benches/quilc"]
+	path = benches/quilc
+	url = git@github.com:quil-lang/quilc.git
```
dbanty, line 3: Not a fan of submodules personally, I find they tend to be hard to maintain. That said—I don't think there's an easier way to get `quilc` other than Docker and Docker would make benchmarking harder 🤷.

benches/parser.rs
```diff
@@ -0,0 +1,60 @@
...
+        });
+    })
+}
+
+struct QuilBenchConfig {
+    name: String,
+    program: String,
+}
+
+fn from_corpus() -> Vec<QuilBenchConfig> {
```
dbanty, line 19: `from_corpus`? 👀

benches/parser.rs
```diff
@@ -0,0 +1,60 @@
...
+    const PATH_SRC: &str = "benches/quilc/tests/good-test-files";
+
+    // collect valid quil programs
+    let mut programs = vec![];
+    let mut corpus_dir = PathBuf::new();
+    PATH_SRC.split('/').for_each(|p| corpus_dir.push(p));
+    let dir = fs::read_dir(corpus_dir).expect("failed to locate quil corpus directory");
```
dbanty, line 26: I think you can do this:

```suggestion
    const PATH_SRC: &str = "benches/quilc/tests/good-test-files";

    // collect valid quil programs
    let mut programs = vec![];
    let mut corpus_dir = PathBuf::from(PATH_SRC);
    PATH_SRC.split('/').for_each(|p| corpus_dir.push(p));
    let dir = fs::read_dir(corpus_dir).expect("failed to locate quil corpus directory");
```

Or, better yet, this:

```suggestion
    const PATH_SRC: &str = "benches/quilc/tests/good-test-files";

    // collect valid quil programs
    let mut programs = vec![];
    let dir = fs::read_dir(PATH_SRC).expect("failed to locate quil corpus directory");
```

benches/parser.rs
```diff
@@ -0,0 +1,60 @@
...
+
+    dir.filter_map(|res| {
+        if res.is_ok() {
+            Some(res.unwrap())
+        } else {
+            None
+        }
+    })
```
dbanty, line 34:
```suggestion
    dir.filter_map(Result::ok)
```

benches/parser.rs
```diff
@@ -0,0 +1,60 @@
...
+                .to_str()
+                .expect("bad filename")
+                .to_string();
+
+            // attempt to parse the quil once, ignoring unparsable input (only benchmark parsable code)
+            if quil_rs::Program::from_str(&program).is_ok() {
+                programs.push(QuilBenchConfig { name, program });
+            }
+        }
+    });
```
dbanty, line 54: Move the for_each predicate into a filter. Spacing looks weird in suggestion-mode, I might have messed something up 😬.

```suggestion
    .filter(|entry| {
        entry
            .metadata()
            .expect("failed to read file metadata")
            .is_file()
    })
    .for_each(|entry| {
        let program =
            fs::read_to_string(entry.path()).expect("failed to read quil program file");
        let name = entry
            .file_name()
            .to_str()
            .expect("bad filename")
            .to_string();

        // attempt to parse the quil once, ignoring unparsable input (only benchmark parsable code)
        if quil_rs::Program::from_str(&program).is_ok() {
            programs.push(QuilBenchConfig { name, program });
        }
    });
```

benches/parser.rs
```diff
@@ -0,0 +1,60 @@
...
+            let name = entry
+                .file_name()
+                .to_str()
+                .expect("bad filename")
+                .to_string();
+
+            // attempt to parse the quil once, ignoring unparsable input (only benchmark parsable code)
+            if quil_rs::Program::from_str(&program).is_ok() {
+                programs.push(QuilBenchConfig { name, program });
+            }
```
dbanty, line 52: Should probably add an issue to catch anything `quil-rs` can't parse, right? Might give us a better idea of what is left to implement to be on-par with `quilc`.

dbanty (approved):
LGTM. A couple suggestions, nothing blocking though.

.gitmodules
nilslice: Yea, generally I agree.. avoid if possible. We can come back to this later if it starts to make things gross!

benches/parser.rs
nilslice: Haha a "corpus" as in: 

_noun_

1. a collection of written texts, especially the entire works of a particular author or a body of writing on a particular subject.

benches/parser.rs
nilslice: d'oh good catch! going with `PathBuf::from`, not sure if `read_dir` is platform independent.

benches/parser.rs
```diff
@@ -0,0 +1,53 @@
...
+        })
+        .for_each(|entry| {
+            let program =
+                fs::read_to_string(entry.path()).expect("failed to read quil program file");
+            let name = entry
+                .file_name()
+                .to_str()
+                .expect("bad filename")
+                .to_string();
+
```
genos, line 42: Perhaps it's the code golfer in me, but I feel like this `filter_map().filter().for_each()` could probably be collapsed.

genos (commented):
Thanks for jumping on this, it's very exciting!

benches/parser.rs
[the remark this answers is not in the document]
genos: Outdated, plz ignore.

benches/parser.rs
```diff
@@ -0,0 +1,60 @@
+use criterion::{criterion_group, criterion_main, Criterion};
+use std::{fs, path::PathBuf, str::FromStr};
+
+fn benchmark_quil_corpus(c: &mut Criterion) {
+    from_corpus().iter().for_each(|cfg| {
+        c.bench_function(&cfg.name, |b| {
+            b.iter(|| {
+                let _ = quil_rs::Program::from_str(&cfg.program);
```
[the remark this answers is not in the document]
genos: Oh! Probably want to use `criterion::black_box` as well, no?

benches/parser.rs
nilslice: yep, definitely something top of mind!

benches/parser.rs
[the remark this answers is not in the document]
nilslice: Good question -- in this case, I don't think there is any optimization (I have not checked the optimized IR, though!). 

I initially used `black_box`, and found that it actually made the average _faster_. I assume this is because Rust no longer needs to `drop` the value, which isn't really a real-world characteristic of a program that would use `quil-rust`. 

The benchmarks have also been run with the `Result` unwrapped, an actual variable binding, as well as an additional use of `black_box` just to use the unwrapped `Program` from the `Result`. They didn't have much impact, so I figured I'd use the simplest bit in the benchmark. Please feel free to tinker though, or I can try other suggestions - definitely don't want these results to be misleading 👍

benches/parser.rs
```diff
@@ -0,0 +1,60 @@
...
+    })
+}
+
+struct QuilBenchConfig {
+    name: String,
+    program: String,
+}
+
+fn from_corpus() -> Vec<QuilBenchConfig> {
+    const PATH_SRC: &str = "benches/quilc/tests/good-test-files";
```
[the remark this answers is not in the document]
nilslice: In other cases, it probably could make sense to use that, but `benches` is a required path for benchmarks to be within - so there is already a base requirement that this path exists.

benches/parser.rs
nilslice: Fore!

Maybe it could? Would be happy to see if you have anything in mind. TBH it's good enough for me, and probably compiles down to the same thing a more visually compact version would. On the plus side, readability is still high enough here, where more combinator stuff might cut into that margin.

benches/parser.rs
nilslice: *update*: I reverted the change to use `PathBuf::from` here since I am pretty sure both are dependent on the running platform.

PR #56, opened 2022-03-10
test use of criterion default baselines with archive

nilslice:
Simplify & correct the download/archive flow for benchmark results. Each PR will be benchmarked against the current `base` result. When the branch is `main`, the `base` result is archived for use in the next workflow run. 

Upon merging to `main`, the benchmark should run again and archive its results.


dbanty (approved):
LGTM

PR #57, opened 2022-03-11
Fix: change quilc submodule to use HTTPS

notmgsk:
This caused an issue in CI, which doesn't have an SSH key to provide for github auth. Unnecessary anyway -- the quilc repo is public.


nilslice:
Good call, thank you!

nilslice (approved):

PR #58, opened 2022-03-14
Update: parse ternary, binary, and unary logic

nilslice:
Parses:

```quil
EQ dest ro 0
LT dest ro[1] -1
LE dest ro 1.2
GT dest ro[2] 1e-6
GE dest ro x
AND ro ro[1]
IOR ro[1] ro[2]
XOR ro[1] ro
NOT ro
NEG ro[1]
```

TODO: 
- [x] https://github.com/rigetti/quil-rust/pull/58#discussion_r825672739
- [x] https://github.com/rigetti/quil-rust/pull/58#pullrequestreview-908406183
- [x] https://github.com/rigetti/quil-rust/pull/58#pullrequestreview-908406896
- [x] https://github.com/rigetti/quil-rust/pull/58#issuecomment-1066493020


src/instruction.rs
```diff
@@ -364,6 +420,13 @@ pub struct Arithmetic {
     pub source: ArithmeticOperand,
 }
 
+#[derive(Clone, Debug, PartialEq)]
+pub struct Logic {
+    pub operator: LogicalOperator,
+    pub operands: [MemoryReference; 2],
```
nilslice, line 426: Quil specfies `logicalBinaryOp     : ( AND | OR | IOR | XOR ) addr ( addr | INT ) ;` 

TODO: add support for `addr ( addr | INT )` operands, not `addr addr`

src/parser/command.rs
```diff
@@ -57,6 +58,24 @@ pub fn parse_arithmetic(
     ))
 }
 
+/// Parse an arithmetic instruction of the form `destination source`.
+/// Called using the arithmetic operator itself (such as `ADD`) which should be previously parsed.
```
nilslice, line 62: TODO: fix doc comment here

src/instruction.rs
```diff
@@ -801,6 +866,9 @@ impl fmt::Display for Instruction {
                 write!(f, "JUMP-WHEN @{} {}", target, condition)
             }
             Instruction::Label(Label(label)) => write!(f, "LABEL @{}", label),
+            Instruction::Logic(Logic { operator, operands }) => {
+                write!(f, "{} {} {}", operator, operands[0], operands[1])
+            }
```
nilslice, line 871: TODO: test

nilslice:
Account for https://github.com/quil-lang/quil/blob/master/grammars/Quil.g4#L172-L178

src/instruction.rs
```diff
@@ -57,6 +57,62 @@ impl fmt::Display for ArithmeticOperator {
...
+    Ior,
+    Xor,
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum UnaryLogic {
+    Neg,
+    Not,
+    True,
+    False,
```
genos, line 94: The [grammar](https://github.com/quil-lang/quil/blob/master/grammars/Quil.g4#L172-L178) marks `True` and `False` as deprecated; should we still include them?

src/instruction.rs
```diff
@@ -57,6 +57,62 @@ impl fmt::Display for ArithmeticOperator {
...
+#[derive(Clone, Debug, PartialEq)]
+pub enum LogicalOperator {
+    Binary(BinaryLogic),
+    Unary(UnaryLogic),
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum BinaryLogic {
+    And,
+    Or,
```
genos, line 84: The [grammar](https://github.com/quil-lang/quil/blob/master/grammars/Quil.g4#L172-L178) marks `Or` as deprecated (presumably in favor of `Ior` and `Xor`). Should we still include it?

src/parser/lexer.rs
```diff
@@ -343,6 +345,13 @@ fn lex_variable(input: &str) -> LexResult {
     })(input)
 }
 
+// fn lex_logical_binary_op(input: &str) -> LexResult {
```
genos, line 348: Commented out code

src/instruction.rs
genos: What's more, I don't see them being used elsewhere…

src/parser/command.rs
```diff
@@ -57,6 +58,24 @@ pub fn parse_arithmetic(
...
+
+    Ok((
+        input,
+        Instruction::Logic(Logic {
+            operator,
+            operands: [left, right],
+        }),
+    ))
+}
+
```
genos, line 78: Any plans for a `parse_logical_unary`?

src/program/memory.rs
```diff
@@ -90,6 +90,16 @@ impl Instruction {
...
         match self {
+            Instruction::Logic(Logic { operands, operator }) => match operator {
+                LogicalOperator::Binary(_) => MemoryAccesses {
+                    reads: operands
+                        .iter()
+                        .map(|memref| memref.name.clone())
+                        .collect::<HashSet<String>>(),
+                    ..Default::default()
+                },
+                LogicalOperator::Unary(_) => todo!("implement memory access for unary logic"),
```
genos, line 101: TODO

genos (commented):
A great start! Looking forward to seeing where this goes.

src/instruction.rs
nilslice: ditto, above 😄

src/parser/lexer.rs
nilslice: good catch, thanks!

src/parser/command.rs
nilslice: Yep! Just not implemented yet 😄

src/instruction.rs
notmgsk: IMO, no. The [live version of the spec](https://quil-lang.github.io/) doesn't include it (or the others).

notmgsk (changes requested):
Aside from the clean up needed (commented code, TODOs) and removing the deprecated instructions, LGTM!

I'll defer approval until it's no longer marked `draft`.

src/instruction.rs
nilslice: rolling up to #59

src/parser/command.rs
nilslice: Updated w/ implementation 👍

src/instruction.rs
nilslice: I have removed all deprecated commands

src/program/memory.rs
```diff
@@ -90,6 +90,22 @@ impl Instruction {
...
     pub fn get_memory_accesses(&self) -> MemoryAccesses {
         match self {
+            Instruction::BinaryLogic(BinaryLogic { operands, .. }) => {
+                let mut memories = HashSet::new();
+                memories.insert(operands.0.name.clone());
+                if let LogicalOperand::MemoryReference(mem) = &operands.1 {
+                    memories.insert(mem.name.clone());
+                }
+
+                MemoryAccesses {
```
kalzoo, line 100: Don't the binary commands both read from and write to the first operand?

This wouldn't cause issues in the current system because reads will block all captures and writes, and there couldn't be a concurrent write happening, but it's worth keeping things correct in case that changes.

src/program/memory.rs
```diff
@@ -90,6 +90,22 @@ impl Instruction {
...
+                    reads: memories,
+                    ..Default::default()
+                }
+            }
+            Instruction::UnaryLogic(UnaryLogic { operand, .. }) => MemoryAccesses {
+                reads: HashSet::from([operand.name.clone()]),
```
kalzoo, line 106: Likewise, isn't this a read and a write?

kalzoo (changes requested):
Small Q / fix - but otherwise lgtm 👍

src/program/memory.rs
nilslice: Updated. I was a little murky on this section, and forgot to call-out a request for assist here. Would be great to also cover this in 1x1.

nilslice:
@notmgsk @kalzoo - I've merged #60 into this branch, which adds support for ternary logic operations. Would appreciate a review when you have a moment. If you want to look at _only_ the new ternary support, you can see a clean diff in #60.

src/instruction.rs
```diff
@@ -57,6 +57,90 @@ impl fmt::Display for ArithmeticOperator {
...
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum TernaryOperator {
+    Equal,
+    GreaterThanOrEqual,
+    GreaterThan,
+    LessThanOrEqual,
+    LessThan,
+}
```
nilslice, line 130: Open to making these more concise, e.g. `Eq`, `Gt` etc. I sort of prefer them in long-form, but only because I didn't feel strongly about the equivalent `Le` vs. `Lte`-style abbreviations, and at least in long-form there is no variation.

nilslice:
Another note, more generally speaking, open to renaming things like `TernaryOperator` to `ComparisonOperator` or similar. The grouping of these follows the Quil spec, but the naming is a bit more broad. For example, a `LOAD` is also ternary but is separated out on its own.

src/instruction.rs
kalzoo: My vote is for the long form like you have it here

src/instruction.rs
```diff
@@ -57,6 +57,90 @@ impl fmt::Display for ArithmeticOperator {
...
+        }
+    }
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum BinaryOperator {
+    And,
+    Ior,
+    Xor,
+}
```
kalzoo, line 80: Nit: looks like this needs a `cargo fmt` and perhaps a GH action to match?

src/instruction.rs
```diff
@@ -57,6 +57,90 @@ impl fmt::Display for ArithmeticOperator {
...
+        match &self {
+            TernaryOperand::LiteralInteger(value) => write!(f, "{}", value),
+            TernaryOperand::LiteralReal(value) => write!(f, "{}", value),
+            TernaryOperand::MemoryReference(value) => write!(f, "{}", value),
+        }
+    }
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum TernaryOperator {
```
kalzoo, line 124: Not 100% on this - but like in your own comment, I wonder if this would be more intuitive as `ComparisonOperator`. The pattern of grouping multiple instructions by common function and syntax was meant to reduce duplicate code and match arms, but my goal was still to keep it intuitive to read in `Debug` format or in code. If I were looking to add an `LT` instruction to a program, I might not look for `TernaryInstruction`. This code is not just used to parse text, but also to build up programs at runtime.

src/instruction.rs
```diff
@@ -364,6 +448,24 @@ pub struct Arithmetic {
...
 }
 
+#[derive(Clone, Debug, PartialEq)]
+pub struct TernaryLogic {
+    pub operator: TernaryOperator,
+    pub operands: (MemoryReference, MemoryReference, TernaryOperand),
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct BinaryLogic {
```
kalzoo, line 458: My comment above about `TernaryLogic` aside, I think `BinaryLogic` and `UnaryLogic` are more intuitive. I'm not sure if that's arbitrary or inconsistent.

That is, if I wanted an `LT`, I'd look for `Comparison`. If I wanted a `NEG`, `UnaryLogic` seems like it would make sense.

That being said, I wonder if we'd want a suite of macros where the names are the Quil instruction names themselves, so there's a part of the code that maps 1<>1 with Quil, while not requiring all users of quil-rust to match against so many variants of `Instruction` (as they would have to if `LT`, `GT`, etc got their own variants)

kalzoo (commented):
Aside from the possible naming change (and a `fmt`), looks great!

src/instruction.rs
nilslice: Yep, I think switching the name makes sense here. I will also do some thinking on addressing:

> That being said, I wonder if we'd want a suite of macros where the names are the Quil instruction names themselves, so there's a part of the code that maps 1<>1 with Quil, while not requiring all users of quil-rust to match against so many variants of Instruction (as they would have to if LT, GT, etc got their own variants)

Definitely could add something to help a user construct programs especially when using the grouped instructions I've included in this PR.

genos:
LGTM! 👍

src/parser/instruction.rs
```diff
@@ -242,6 +246,211 @@ mod tests {
...
+            "EQ ro 1 1",
+            "LT 1 1 1",
+            "LE 1 x ro",
+            "GT 1 ro x",
+            "GE dest 0.3 4",
+        ]
+        .iter()
+        .for_each(|input| {
+            let tokens = lex(input).unwrap();
+            assert!(parse_instructions(&tokens).is_err())
```
kalzoo, line 341: Might be worth adding an assertion message with the input string so it's clear which one fails

kalzoo (approved):
Looks good, thanks!

PR #60, opened 2022-03-15
feat: support ternary logic instructions

nilslice:
Adds support for `EQ, LT, LE, GT, GE` instructions as per [Quil spec #6-5Classical-Instructions](https://quil-lang.github.io/#6-5Classical-Instructions)


nilslice:
Going to self-merge this into #58, and a consolidated review can be conducted there. I think that is better than 2 separate releases for these PRs.

PR #63, opened 2022-04-12
Add Expression.to_real()

genos:
In using this library elsewhere, I've noticed folks doing something like

```rust
match x {
    case Expression::Number(x) if x.im == 0.0 => {
      ...
    },
    ...
}
```

which to me seems

1. repetitive, since we'd like to just match on a real number and move on with our lives; and
2. error-prone, since it's conceivable (however unlikely) that we might have an expression that algebraically _should_ be a real number but has a non-zero but _wicked_ tiny imaginary part due to rounding issues from other computation.

This PR creates `Expression.to_real() -> Result<f64, EvaluationError>` and adds two descriptive types of `EvaluationError`.

### DIFF
--- a/src/expression.rs
+++ b/src/expression.rs
@@ -14,6 +14,7 @@
  * limitations under the License.
  **/
 use std::collections::{hash_map::DefaultHasher, HashMap};
+use std::convert::TryInto;
 use std::f64::consts::PI;
 use std::fmt;
 use std::hash::{Hash, Hasher};
@@ -330,6 +331,15 @@ impl Expression {
             Expression::Number(_) => Err(EvaluationError::NumberNotReal),
             _ => Err(EvaluationError::NotANumber),
         }
+    }
+}
+
+/// If this is a number with imaginary part "equal to" zero (of _small_ absolute value), return
+/// that number. Otherwise, error with an evaluation error of a descriptive type.
+impl TryInto<f64> for Expression {
+    type Error = EvaluationError;
+    fn try_into(self) -> Result<f64, Self::Error> {
+        self.to_real()
     }
 }
 
--- a/src/expression.rs
+++ b/src/expression.rs
@@ -28,6 +28,8 @@ use crate::{imag, instruction::MemoryReference, real};
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub enum EvaluationError {
     Incomplete,
+    NumberNotReal,
+    NotANumber,
 }
 
 #[derive(Clone, Debug)]
@@ -172,6 +174,12 @@ fn calculate_function(
     }
 }
 
+/// Is this a small floating point number?
+#[inline(always)]
+fn _is_small(x: f64) -> bool {
+    x.abs() < 1e-16
+}
+
 impl Expression {
     /// Consume the expression, simplifying it as much as possible.
     ///
@@ -310,6 +318,16 @@ impl Expression {
                 .ok_or(EvaluationError::Incomplete),
             PiConstant => Ok(real!(PI)),
             Number(number) => Ok(*number),
+        }
+    }
+
+    /// If this is a number with imaginary part "equal to" zero (of _small_ absolute value), return
+    /// that number. Otherwise, error with an evaluation error of a descriptive type.
+    pub fn to_real(self) -> Result<f64, EvaluationError> {
+        match self {
+            Expression::Number(x) if _is_small(x.im) => Ok(x.re),
+            Expression::Number(_) => Err(EvaluationError::NumberNotReal),
+            _ => Err(EvaluationError::NotANumber),
         }
     }
 }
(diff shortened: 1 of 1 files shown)


src/expression.rs
```diff
@@ -172,6 +174,12 @@ fn calculate_function(
     }
 }
 
+/// Is this a small floating point number?
+#[inline(always)]
+fn _is_small(x: f64) -> bool {
```
nilslice, line 179: I think I'd `s/_is_small/is_small`, as it's already private unless marked `pub` otherwise. But maybe you're trying to indicate something else?

genos: Great point! I've been writing a bunch of Python recently. I'll update `_hash_to_u64` as well (since I'm the guilty party there as well).

src/expression.rs
```diff
@@ -674,5 +692,29 @@ mod tests {
...
+
+        #[test]
+        fn no_other_exps_are_real(expr in arb_expr().prop_filter("Not numbers", |e| match e {
+            Expression::Number(_) => false,
+            _ => true,
+        }
+            )) {
+            prop_assert_eq!(expr.to_real(), Err(EvaluationError::NotANumber))
+        }
+
```
nilslice, line 718: I'm not super familiar with the way `prop_test` picks values, but should there maybe be a concrete test with specifically `0.0` or something that might be commonly used in a Quil program?

genos: Thanks! Reminds me that I should probably handle π as well.

nilslice: yum.. 🥧

nilslice (approved):
Seems like a good idea to me! I'd await a review from one of @kalzoo / @dbanty / @notmgsk though as well.

genos:
I suppose we could also try something along the lines of `impl TryInto<f64> for Expression`, if folks think that'd be more idiomatic/ergonomic/etc.

nilslice:
> I suppose we could also try something along the lines of impl TryInto<f64> for Expression, if folks think that'd be more idiomatic/ergonomic/etc

Suppose both wouldn't hurt... then you have the impl in case some function expects one, and the added benefit of an obvious method to call.

src/expression.rs
```diff
@@ -28,6 +29,8 @@ use crate::{imag, instruction::MemoryReference, real};
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub enum EvaluationError {
     Incomplete,
+    NumberNotReal,
```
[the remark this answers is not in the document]
kalzoo: (we can accommodate that by squashing on merge with a `Breaking:` prefix)

src/expression.rs
```diff
@@ -312,6 +321,26 @@ impl Expression {
             Number(number) => Ok(*number),
         }
     }
+
+    /// If this is a number with imaginary part "equal to" zero (of _small_ absolute value), return
+    /// that number. Otherwise, error with an evaluation error of a descriptive type.
+    pub fn to_real(self) -> Result<f64, EvaluationError> {
```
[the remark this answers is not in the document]
genos: be rational get real

src/expression.rs
```diff
@@ -312,6 +321,26 @@ impl Expression {
...
+            Expression::Number(x) if is_small(x.im) => Ok(x.re),
+            Expression::Number(_) => Err(EvaluationError::NumberNotReal),
+            _ => Err(EvaluationError::NotANumber),
+        }
+    }
+}
+
+/// If this is a number with imaginary part "equal to" zero (of _small_ absolute value), return
+/// that number. Otherwise, error with an evaluation error of a descriptive type.
+impl TryInto<f64> for Expression {
```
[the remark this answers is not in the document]
genos: Point well taken! Removing it.

src/expression.rs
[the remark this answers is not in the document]
genos: I'm adding some documentation to them, and will mark the incoming commit at `Breaking:`.

genos:
I've just hit a testing error via `proptest` that has me stumped!

```
failures:

---- expression::tests::eq_implies_hash_eq stdout ----
thread 'expression::tests::eq_implies_hash_eq' panicked at 'Test failed: assertion failed: `(left == right)`
  left: `true`,
 right: `false` at src/expression.rs:697; minimal failing input: x = FunctionCall { function: Cosine, expression: Prefix { operator: Plus, expression: PiConstant } }, y = FunctionCall { function: Cosine, expression: Prefix { operator: Plus, expression: PiConstant } }
        successes: 0
        local rejects: 0
        global rejects: 0
', src/expression.rs:628:5
```

For some reason, the expression 

```rust
Expression::FunctionCall {
    function: ExpressionFunction::Cosine,
    expression: Box::new(Expression::Prefix {
        operator: PrefixOperator::Plus,
        expression: Box::new(Expression::PiConstant),
    }),
};
```

evaluates as equal to itself, but is hashing to two different things? I'll open a new issue.

kalzoo (approved):
lgtm, thanks @genos !

genos:
Thanks, @kalzoo & @nilslice! If it's OK with you, I'll leave y'all to do the merging to ensure we get the proper message prefixes and semver behavior.

dbanty:
@genos it's not clear to me at a glance why this was a breaking change. Can you describe what the breaking piece is so I can add to the release notes? Want to make sure nothing breaks when I update dependents 😅

genos:
Thanks @dbanty! @kalzoo said that the additions to the `EvaluationError` enum made it breaking: https://github.com/rigetti/quil-rust/pull/63#discussion_r848956297

dbanty:
Thanks @genos! I'll stick that note in the release notes.

PR #66, opened 2022-05-04
Update: Make Expression.to_real take a reference

genos:
In using `Expression.to_real()` from 0.9.0 in a downstream project, I found myself tripped up by needing to take ownership over the expression rather than just borrowing a reference. This hopefully fixes that.


dbanty (approved):
LGTM. This seems like the sort of thing Clippy should have warned us about. I thought I'd seen a "you can pass a reference instead" lint before.

PR #68, opened 2022-05-26
New: Program::get_used_qubits()

notmgsk:


src/program/mod.rs
```diff
@@ -367,4 +387,29 @@ DECLARE ec BIT
...
+    1,2
+I 0
+";
+        let program = Program::from_str(input).unwrap();
+        let expected = vec![Qubit::Fixed(0), Qubit::Variable("q".to_string())]
+            .into_iter()
+            .collect::<HashSet<_>>();
+        let actual = program.get_used_qubits();
+        assert_eq!(expected, actual);
+    }
```
nilslice, line 414: I may be missing it, but consider also include a qubit operation in the test Quil program that would _not_ be returned from the `get_used_qubits` method to check that this behavior is correct and continues to be later.

nilslice (changes requested):

src/program/mod.rs
nilslice: Thinking back on this, I'm not sure my request makes sense! Happy to connect on this offline 😆

kalzoo:
I fumbled the merge-and-delete of `feat/improve-dot-graphs` - @notmgsk if you'd like to see this merged please just open a new MR against `main`.

src/program/mod.rs
notmgsk: Yeah, not sure I can think of a case where that would happen.

PR #73, opened 2022-06-01
Fix how an instruction's used/blocked frames are calculated

### LINKED ISSUE #72: Bug in FENCE / NONBLOCKING behavior
From the [quil spec](https://github.com/quil-lang/quil/blob/master/rfcs/analog/proposal.md#nonblocking):

> A pulse operation (PULSE, CAPTURE, and RAW-CAPTURE) with the NONBLOCKING modifier does not exclude **pulse operations on other frames** [emphasis added]

> The FENCE instruction provides a means of synchronization of all frames involving a set of qubits. In particular, it guarantees that all instructions involving any of the fenced qubits preceding the FENCE are completed before any instructions involving the fenced qubits which follow the FENCE.

Currently, when constructing the timing graph/vector clock, FENCE instructions which follow NONBLOCKING instructions do not register a timing/frame dependency on the latter, which means that instructions may be compiled out-of-sequence.

kalzoo:
Closes #72 

See issue for context.

Reworks the way that frame-blocking is computed for an instruction within the scope of a program. The previous way was to perform that work directly in `Program::get_used_frames` and then just pull those frames out of the program's `FrameSet`. This was brittle and harder to understand in the light of the Quil spec.

This new approach is simpler in that it takes place in two steps:

- First, a `FrameMatchCondition` is computed for an instruction, which is independent of the program in which that instruction lives.
- Then, that condition is evaluated against the program's `FrameSet`, which converts the "abstract" frame requirements of the instruction to the concrete frames available in the program.

It also adds a new private method, `Instruction::parse`, for the convenience of the tests, so that single instructions can easily be used as fixtures.

I considered using `rstest` for the table-driven test, but I'm not sure if we need that yet, and it will be easy to change over if we do.

**Notes for reviewers**:

* Please carefully compare the tests added in this MR to the [quil spec](https://github.com/quil-lang/quil/blob/master/rfcs/analog/proposal.md) and note/question where they disagree. Note that `DELAY`'s behavior is not fully specified in the spec, and so I kept its behavior consistent with how it was already enforced here. Thoughts welcome on that.
* Please consider if there is a reason that `include_blocking` is not enough information to distinguish all use cases to know the frames for an instruction. Currently, this allows for returning either _used_ frames (those which the instruction will actively use, and for which it will take a dependency on instructions which earlier block it) or _blocked frames_ (instruction may or may not use it, but it prevents other instructions from using it until it's done). Is there any other category of frames that we'd want to query for, or is this boolean argument enough?
* There are some inelegant fixes here (such as querying the used frames and then the blocking frames separately). I don't want inelegance to block this PR (which itself is blocking other work), but if you see a quick, neat fix - please suggest it.


src/program/mod.rs
```diff
@@ -369,4 +387,116 @@ DECLARE ec BIT
...
+                "CAPTURE 0 \"a\" custom_waveform ro[0]",
+                vec!["0 \"a\""],
+                vec!["0 \"a\"", "0 \"b\"", "0 1 \"2q\""],
+            ),
+            (
+                "CAPTURE 1 \"c\" custom_waveform ro[0]",
+                vec!["1 \"c\""],
+                vec!["1 \"c\"", "0 1 \"2q\""],
+            ),
+            ("CAPTURE 2 \"a\" custom_waveform ro[0]", vec![], vec![]),
```
notmgsk, line 433: Include `RAW-CAPTURE` for completeness?

src/program/frame.rs
```diff
@@ -87,3 +126,27 @@ impl FrameSet {
             .collect()
     }
 }
+
+pub(crate) enum FrameMatchCondition<'a> {
+    /// Match all frames in the set
+    All,
+
+    /// Match all frames which shares any one of these names
+    AnyOfNames(&'a [String]),
```
genos, line 135: Why slices here? You're using vectors below (because of needing to put recursive ones on the heap), and if these get large enough, I imagine a set-based solution would be faster.

src/program/frame.rs
```diff
@@ -87,3 +126,27 @@ impl FrameSet {
...
+    ExactQubits(&'a [Qubit]),
+
+    /// Return these specific frames, if present in the set
+    Specific(&'a FrameIdentifier),
+
+    /// Return all frames which match all of these conditions
+    And(Vec<FrameMatchCondition<'a>>),
+
+    /// Return all frames which match any of these conditions
+    #[allow(dead_code)]
```
genos, line 150: Why does dead code need to be allowed here? I see `FrameMatchCondition::Or` above on line 78.

src/program/frame.rs
genos: Removing the lifetime stuff would probably simplify downstream code as well.

src/program/frame.rs
```diff
@@ -35,18 +35,57 @@ impl FrameSet {
...
+                if let Some((frame, _)) = self.frames.get_key_value(frame) {
+                    vec![frame].into_iter().collect()
+                } else {
+                    HashSet::new()
+                }
+            }
+            FrameMatchCondition::And(conditions) => conditions
+                .into_iter()
+                .map(|c| self.get_matching_keys(c))
+                .reduce(|acc, el| acc.into_iter().filter(|&v| el.contains(v)).collect())
```
genos, line 76: I suppose my go-to reaction shouldn't be "maybe we should add another dependency!" but I think `any` and `all` from [`itertools`](https://docs.rs/itertools/latest/itertools/) might be useful.

src/program/graph.rs
```diff
@@ -234,19 +234,30 @@ impl InstructionBlock {
                     Ok(())
                 }
                 InstructionRole::RFControl => {
-                    let frames = match program.get_frames_for_instruction(instruction, true) {
+                    let used_frames = match program.get_frames_for_instruction(instruction, false) {
```
genos, line 237: [`Option::unwrap_or_default`](https://doc.rust-lang.org/std/option/enum.Option.html#method.unwrap_or_default)

genos: Same again for `blocked_frames`

src/program/frame.rs
notmgsk: But it's never constructed anywhere I think is the point. Similar to how if you have `f(g(x))` and `f()` is never called, `g()` is considered dead.

genos: Can we remove it then? Or is it here for future-proofing?

src/program/mod.rs
```diff
@@ -369,4 +387,116 @@ DECLARE ec BIT
...
+                "FENCE",
+                vec!["0 \"a\"", "0 \"b\"", "1 \"c\"", "0 1 \"2q\""],
+                vec!["0 \"a\"", "0 \"b\"", "1 \"c\"", "0 1 \"2q\""],
+            ),
+            (
+                "DELAY 0 1.0",
+                vec!["0 \"a\"", "0 \"b\""],
+                vec!["0 \"a\"", "0 \"b\""],
+            ),
+            ("DELAY 1 1.0", vec!["1 \"c\""], vec!["1 \"c\""]),
```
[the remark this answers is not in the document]
kalzoo: Good point, although how I read the spec it should only be the last of these.

src/program/frame.rs
kalzoo: Removed

src/program/frame.rs
kalzoo: Yeah I think for only a single use case here I might prefer to keep out the added dep

src/program/frame.rs
kalzoo: I suppose it makes sense if you see them all as slices but the Vec as just a `Box<&[_]>` due to the type recursion

src/program/frame.rs
genos: For sure.

genos:
LGTM 🎉

kalzoo:
Note: this fails the example cited in [this issue](https://github.com/quil-lang/quil/issues/59); if the spec remains unchanged, the approach here will need to be modified.

PR #76, opened 2022-06-02
Fix: frame dependency calculation

### LINKED ISSUE #75: Bug in Timing/Frame Dependencies
Hot on the heels of #73 , another bug has been found in how frame blocking is calculated. Two examples are given here, with further explanation below. Can you spot the bug?

# Example 1

Hint: there is an additional frame defined in the program, `0 1 "cz"`, which is not used by any of the instructions.

graphviz (4)

# Example 2

graphviz (5)

The problem here is that the change in #73 properly fixed how dependencies are calculated _after_ a blocking instruction, but that blocking instruction _must itself block on previous instructions which overlap in frames_, and that is not happening.

Specifically/formally:

1. An instruction which _uses_ a frame must depend on the most-recent _user_ **and** _blockers_ of the frame, if any. Doing so clears all blockers of the frame and this instruction becomes the most-recent user.
2. An instruction which _blocks_ but does not _use_ a frame must depend on the most-recent user of the frame. Doing so leaves the most-recent user in place and adds this instruction to a list of most-recent blockers.

Blocking prevents simultaneous use but allows simultaneous blocking.

This approach is akin to a single-writer, multiple-reader lock or queue, where blocking:reading::using:writing.

kalzoo:
Another fix, back to back with #73.

Closes #75 

This takes the approach of tracking the instructions which _use_ and _block_ a frame separately, solving the problems described in #75. It's more complex, but should still remain readable and maintainable. This PR includes a couple more snapshot test cases to illustrate the differences and prevent regression.

### DIFF
--- a/src/program/graph.rs
+++ b/src/program/graph.rs
@@ -301,19 +301,17 @@ impl InstructionBlock {
                     Ok(())
                 }
                 InstructionRole::RFControl => {
-                    let used_frames: HashSet<&FrameIdentifier> = program
+                    let used_frames = program
                         .get_frames_for_instruction(instruction, false)
-                        .unwrap_or_default()
-                        .into_iter()
-                        .collect();
-                    let blocked_frames: HashSet<&FrameIdentifier> = program
-                        .get_frames_for_instruction(instruction, true)
-                        .unwrap_or_default()
-                        .into_iter()
-                        .filter(|f| !used_frames.contains(f))
-                        .collect();
+                        .unwrap_or_default();
 
-                    for frame in used_frames {
+                    let blocked_frames = program
+                        .get_frames_for_instruction(instruction, true)
+                        .unwrap_or_default();
+
+                    let blocked_but_not_used_frames = blocked_frames.difference(&used_frames);
+
+                    for frame in &used_frames {
                         let previous_node_ids = last_instruction_by_frame
                             .entry(frame.clone())
                             .or_insert(PreviousNodes {
@@ -329,7 +327,7 @@ impl InstructionBlock {
                         }
                     }
 
-                    for frame in blocked_frames {
+                    for frame in blocked_but_not_used_frames {
                         if let Some(previous_node_id) = last_instruction_by_frame
                             .entry(frame.clone())
                             .or_insert(PreviousNodes {
--- a/src/program/graph.rs
+++ b/src/program/graph.rs
@@ -240,7 +240,7 @@ impl PreviousNodes {
     /// Register a node as using a frame, and return the instructions on which it should depend/wait for scheduling (if any).
     ///
     /// A node which uses a frame will block on any previous user or blocker of the frame, much like a writer in a read-write lock.
-    pub fn get_dependencies_for_next_user(
+    fn get_dependencies_for_next_user(
         &mut self,
         node: ScheduledGraphNode,
     ) -> HashSet<ScheduledGraphNode> {
@@ -258,7 +258,7 @@ impl PreviousNodes {
     ///
     /// If the frame is currently blocked by other nodes, it will add itself to the list of blockers,
     /// much like a reader in a read-write lock.
-    pub fn get_dependency_for_next_blocker(
+    fn get_dependency_for_next_blocker(
         &mut self,
         node: ScheduledGraphNode,
     ) -> Option<ScheduledGraphNode> {
--- a/src/program/graph.rs
+++ b/src/program/graph.rs
@@ -200,6 +200,75 @@ pub struct InstructionBlock {
     pub terminator: BlockTerminator,
 }
 
+/// PreviousNodes is a structure which helps maintain ordering among instructions which operate on a given frame.
+/// It works similarly to a multiple-reader-single-writer queue, where an instruction which "uses" a frame is like
+/// a writer and an instruction which blocks that frame is like a reader. Multiple instructions may concurrently
+/// block a frame, but an instruction may not use a frame while it is concurrently used or blocked.
+///
+/// ## Examples
+///
+/// Note that "depends on" is equivalent to "must execute after completion of".
+///
+/// ```text
+/// user --> user # a second user takes a dependency on the first
+///
+/// user --> blocker # multiple blockers take a dependency on the most recent user
+///      \-> blocker
+///      \-> blocker
+///
+/// blocker --> user --> blocker # users and blockers take dependencies on one another,
+///                              # but blockers do not depend on other blocking instructions
+/// ```
+struct PreviousNodes {
+    using: Option<ScheduledGraphNode>,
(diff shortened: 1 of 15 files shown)


genos:
Issue-linking-wise, @kalzoo, is this aimed at #75?

src/program/graph.rs
```diff
@@ -200,6 +200,75 @@ pub struct InstructionBlock {
     pub terminator: BlockTerminator,
 }
 
+/// PreviousNodes is a structure which helps maintain ordering among instructions which operate on a given frame.
```
genos, line 203: `PreviousNodes` _are_ a structure?

j/k

src/program/graph.rs
```diff
@@ -233,24 +301,47 @@ impl InstructionBlock {
...
-                            .unwrap_or(&ScheduledGraphNode::BlockStart);
-                        add_dependency!(graph, *previous_node_id => node, ExecutionDependency::ReferenceFrame);
+                        let previous_node_ids = last_instruction_by_frame
+                            .entry(frame.clone())
+                            .or_insert(PreviousNodes {
+                                using: None,
+                                blocking: vec![ScheduledGraphNode::BlockStart]
+                                    .into_iter()
+                                    .collect(),
+                            })
```
genos, line 324:
```rust
.or_default()
```

src/program/graph.rs
```diff
@@ -233,24 +301,47 @@ impl InstructionBlock {
...
                     for frame in blocked_frames {
-                        last_instruction_by_frame.insert(frame.clone(), node);
+                        if let Some(previous_node_id) = last_instruction_by_frame
+                            .entry(frame.clone())
+                            .or_insert(PreviousNodes {
+                                using: None,
+                                blocking: vec![ScheduledGraphNode::BlockStart]
+                                    .into_iter()
+                                    .collect(),
+                            })
```
genos, line 340:
```rust
.or_default()
```

src/program/graph.rs
```diff
@@ -233,24 +301,47 @@ impl InstructionBlock {
                     Ok(())
                 }
                 InstructionRole::RFControl => {
-                    let used_frames = program
+                    let used_frames: HashSet<&FrameIdentifier> = program
```
genos, line 304: `get_frames_for_instruction` returns an `Option<Vec<<&FrameIdentifier>>`, but its only uses always convert that to a `HashSet<&FrameIdentifier>`. Would it make sense to DRY this out and change the function definition?

src/program/graph.rs
```diff
@@ -295,8 +386,10 @@ impl InstructionBlock {
         // does not terminate until these are complete
         add_dependency!(graph, last_classical_instruction => ScheduledGraphNode::BlockEnd, ExecutionDependency::StableOrdering);
 
-        for (_, last_instruction) in last_instruction_by_frame {
-            add_dependency!(graph, last_instruction => ScheduledGraphNode::BlockEnd, ExecutionDependency::ReferenceFrame);
+        for (_, last_instructions) in last_instruction_by_frame {
```
genos, line 389:
```rust
for last_instruction in last_instruction_by_frame.values()
```

src/program/graph.rs
```diff
@@ -200,6 +200,75 @@ pub struct InstructionBlock {
...
+        result
+    }
+
+    /// Register a node as blocking a frame, and return the instructions on which it should depend/wait for scheduling (if any).
+    ///
+    /// A node which blocks a frame will block on any previous user of the frame, but not concurrent blockers.
+    ///
+    /// If the frame is currently blocked by other nodes, it will add itself to the list of blockers,
+    /// much like a reader in a read-write lock.
+    pub fn register_blocker(&mut self, node: ScheduledGraphNode) -> Option<ScheduledGraphNode> {
```
genos, line 258: I can't think of a better name, so maybe ignore this, but I think a more suggestive function name that says something about returning `self.using` might be helpful.

I suppose I have a similar (-ly unhelpful) note for `register_user`'s naming, too.

genos (commented):
Some requests for shorter code 🏌️, a suggestion about changing the return type of a function, but otherwise LGTM

src/program/graph.rs
kalzoo: Ah right! I added the default impl then forgot to use it

src/program/graph.rs
kalzoo: Yes - although that's a breaking change - _although_ that's fine these days.

kalzoo:
Note: this fails the example cited in [this issue](https://github.com/quil-lang/quil/issues/59); if the spec remains unchanged, this will have to be modified.

src/program/graph.rs
kalzoo: Changed - what do you think?

src/program/mod.rs
```diff
@@ -130,15 +130,10 @@ impl Program {
...
+    ) -> Option<HashSet<&'a FrameIdentifier>> {
         instruction
             .get_frame_match_condition(include_blocked)
-            .map(|condition| {
-                self.frames
-                    .get_matching_keys(condition)
-                    .into_iter()
-                    .collect()
-            })
+            .map(|condition| self.frames.get_matching_keys(condition))
```
notmgsk, line 136: How come this doesn't need a `.collect()`? 🤔

src/program/graph.rs
```diff
@@ -200,6 +200,81 @@ pub struct InstructionBlock {
...
+            blocking: vec![ScheduledGraphNode::BlockStart].into_iter().collect(),
+        }
+    }
+}
+
+impl PreviousNodes {
+    /// Register a node as using a frame, and return the instructions on which it should depend/wait for scheduling (if any).
+    ///
+    /// A node which uses a frame will block on any previous user or blocker of the frame, much like a writer in a read-write lock.
+    pub fn get_dependencies_for_next_user(
```
notmgsk, line 243: Do these need to be `pub`?

notmgsk (approved):
Nice. Took me a hot minute to understand the logic, but that might be because I've been awake since 5AM. Only question is: do the `PreviousNodes` methods need to be public?

src/program/mod.rs
genos: Because `get_matching_keys` already returns a `HashSet`?

That said, you might be able to shorten this to `.map(self.frames.get_matching_keys)`.

src/program/graph.rs
genos: 👍

src/program/graph.rs
kalzoo: nope, good call (although the struct itself is not pub)

src/program/mod.rs
kalzoo: Compiler doesn't like passing that function to `map`, although I'm not sure why.

PR #78, opened 2022-06-06
New: Program::get_used_qubits()

notmgsk:


nilslice (approved):
I think this looks good -- I just checked that you match all `Instruction`s with one or more `Qubit`s that could undergo some operation. 

🚢... unless there's something else that you'd like double-checked

PR #79, opened 2022-06-06
chore: convert terminate_working_block macro to fn

nilslice:
While building up some more familiarity with #75, this was easier to follow w/ language server help. I think this function is pretty well covered in other tests, which would have caught issues, but that's an area I could look into more.

PR branch targets and is based on `75-frame-dependencies`. 

UPDATE: forgot to test with `--all-features` locally.


README.md
```diff
@@ -9,3 +9,12 @@ It serves three purposes:
 3. Construct a dependency graph among program instructions
 
 It should be considered unstable until the release of v1.0.
+
+
+## Testing
+
+When testing this crate, you should run with the `--all-features` flag to ensure all tests are executed.
```
kalzoo, line 16: 👍 true

kalzoo (approved):
Good change, thank you 🙏

README.md
genos: 👏

src/program/graph.rs
```diff
@@ -452,33 +452,36 @@ pub struct ScheduledProgram {
...
+fn terminate_working_block(
+    terminator: Option<BlockTerminator>,
+    working_instructions: &mut Vec<Instruction>,
+    blocks: &mut IndexMap<String, InstructionBlock>,
+    working_label: &mut Option<String>,
+    program: &Program,
+    instruction_index: Option<usize>,
+) -> ScheduleResult<()> {
+    // If this "block" has no instructions and no terminator, it's not worth storing - skip it
+    if working_instructions.is_empty() && terminator.is_none() && working_label.is_none() {
```
genos, line 465: Since you assign `*working_label = None` in both cases, you could pull that outside the conditional and have a single arm:
```rust 
if !working_instructions.is_empty() || !terminator.is_none() || !working_label.is_none() {
    …
}
*working_label = None;
Ok(())
```

src/program/graph.rs
```diff
@@ -452,33 +452,36 @@ pub struct ScheduledProgram {
...
-                    variant: ScheduleErrorVariant::DuplicateLabel,
-                }), // Duplicate label
-                None => Ok(()),
-            }?;
-            $working_instructions = vec![];
-            $working_label = None
-        }
-        Ok(())
-    }};
+#[allow(unused_assignments)]
```
genos, line 455: Not too troubled, just curious: why is this necessary? Because of reassigning `working_label` at the end?

src/program/graph.rs
```diff
@@ -452,33 +452,36 @@ pub struct ScheduledProgram {
...
+        let label = working_label
+            .clone()
+            .unwrap_or_else(|| ScheduledProgram::generate_autoincremented_label(blocks));
+
+        match blocks.insert(label.clone(), block) {
+            Some(_) => Err(ScheduleError {
+                instruction_index,
+                instruction: Instruction::Label(Label(label)),
+                variant: ScheduleErrorVariant::DuplicateLabel,
+            }), // Duplicate label
```
genos, line 478: With `variant: ScheduleErrorVariant::DuplicateLabel` here, the comment seems a bit superfluous.

genos (commented):
Nice! A question for my own edification, some few ignorable suggestions, but LGTM.

src/program/graph.rs
nilslice: good eye -- updated to early return!

src/program/graph.rs
nilslice: Yep, no longer needed though after the branching change

src/program/graph.rs
nilslice: removed! good catch ✅

src/program/graph.rs
```diff
@@ -452,33 +452,40 @@ pub struct ScheduledProgram {
...
-                None => Ok(()),
-            }?;
-            $working_instructions = vec![];
-            $working_label = None
-        }
-        Ok(())
-    }};
+/// Builds an [`InstructionBlock`] from provided instructions, terminator, and program, then tracks
+/// the block and its label, and resets the instruction list and label for a future block to use.
+fn terminate_working_block(
```
nilslice, line 457: @kalzoo - feedback on depth/correctness of this comment here would be great, thanks!

src/program/graph.rs
```diff
@@ -452,33 +452,40 @@ pub struct ScheduledProgram {
...
+    terminator: Option<BlockTerminator>,
+    working_instructions: &mut Vec<Instruction>,
+    blocks: &mut IndexMap<String, InstructionBlock>,
+    working_label: &mut Option<String>,
+    program: &Program,
+    instruction_index: Option<usize>,
+) -> ScheduleResult<()> {
+    // If this "block" has no instructions and no terminator, it's not worth storing - skip it
+    if working_instructions.is_empty() && terminator.is_none() && working_label.is_none() {
+        *working_label = None;
```
nilslice, line 467: Ah this can be removed, since we just check that `working_label.is_none()` the line above...

src/program/graph.rs
```diff
@@ -365,33 +452,40 @@ pub struct ScheduledProgram {
...
-                None => Ok(()),
-            }?;
-            $working_instructions = vec![];
-            $working_label = None
-        }
-        Ok(())
-    }};
+/// Builds an [`InstructionBlock`] from provided instructions, terminator, and program, then tracks
+/// the block and its label, and resets the instruction list and label for a future block to use.
+fn terminate_working_block(
```
kalzoo, line 457:
```suggestion
/// Builds an [`InstructionBlock`] from provided instructions, terminator, and program, then tracks
/// the block and its label, and resets the instruction list and label for a future block to use.
///
/// The "working block" is that being traversed within the program, accumulating instructions located
/// between control instructions (such as `LABEL` and 'JUMP`). When such a control instruction is reached,
/// this function performs the work to close out and store the instructions in the current "working block"
/// and then reset the state to prepare for the next block.
fn terminate_working_block(
```

src/program/graph.rs
kalzoo: made a suggestion to add some context on the `working_block` term, but otherwise lgtm 👍

PR #81, opened 2022-06-08
Fix: DEFCAL MEASURE serialization

### LINKED ISSUE #80: Bug in DEFCAL MEASURE serialization
Currently, `DEFCAL MEASURE` instructions are:

1. Not included in `program.to_string` regardless of that method call's argument
2. Serialized incorrectly, as `DEFCAL MEASURE ... %addr` where `DEFCAL MEASURE ... addr` is expected.

kalzoo:
closes #80 

Simple fix with a test case for regression. See issue for context.

### DIFF
--- a/src/parser/lexer.rs
+++ b/src/parser/lexer.rs
@@ -533,7 +533,7 @@ mod tests {
       DEFCAL RX(%theta) 0:
       \tPULSE 0 \"xy\" my_waveform()
       
-      DEFCAL MEASURE 0 %dest:
+      DEFCAL MEASURE 0 dest:
       \tDECLARE iq REAL[2]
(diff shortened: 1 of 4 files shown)


src/parser/lexer.rs
```diff
@@ -533,7 +533,7 @@ mod tests {
       DEFCAL RX(%theta) 0:
       \tPULSE 0 \"xy\" my_waveform()
       
-      DEFCAL MEASURE 0 %dest:
```
nilslice, line 536: [The spec](https://quil-lang.github.io/) seems to only use the "formal identifier" variant `%dest`. Is there a definition for this variant in this position?

PR #87, opened 2022-07-08
Breaking: Use a proper error type for Program::from_str

### LINKED ISSUE #86: FromStr implementation for Program has String for Err type
We currently have 

```rust
impl FromStr for Program {
    type Err = String;
```

But unfortunately, `String` doesn't implement `std::error::Error`, which is causing some issues and need to wrap things in a downstream project. `Program::from_str` should return something that implements `Error`.

genos:
Closes #86.

This adds a few variants to the public `Program::Error` enumeration and uses them for the possible `Err`s coming off of `Program::from_str`.

### DIFF
--- a/src/program/error.rs
+++ b/src/program/error.rs
@@ -16,7 +16,7 @@ use thiserror::Error;
 
 use crate::instruction::Instruction;
 
-#[derive(Debug, Error)]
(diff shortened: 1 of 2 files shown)


dbanty (approved):
Nice!

notmgsk (approved):

PR #88, opened 2022-08-09
Update: Add optional type-checking, including custom errors

genos:
This PR stems from work we had going internally for something that consumes Quil programs, but it made more sense to upstream it to this repository.

Changes:

- add `src/program/checks.rs`, making it public with a single public function `type_check(&Program) -> ProgramResult<()>`
- add more type-checking-related error types to `src/program/error.rs`
- add `rstest` to the dev dependencies to express a lot of test cases in a rather terse format

### DIFF
--- a/src/program/checks.rs
+++ b/src/program/checks.rs
@@ -1,0 +1,919 @@
+//! Checks for consistency and validity of Quil programs.
+use crate::{
+    expression::Expression,
+    instruction::{
+        Arithmetic, ArithmeticOperand, ArithmeticOperator, BinaryLogic, BinaryOperand,
+        BinaryOperator, Comparison, ComparisonOperand, ComparisonOperator, Exchange, Instruction,
+        Load, MemoryReference, Move, ScalarType, SetFrequency, SetPhase, SetScale, ShiftFrequency,
+        ShiftPhase, Store, UnaryLogic, UnaryOperator,
+    },
+    program::{
+        error::{ProgramError, ProgramResult},
+        MemoryRegion,
+    },
+    Program,
+};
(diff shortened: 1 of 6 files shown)


src/program/checks.rs
```diff
@@ -0,0 +1,917 @@
...
+                        data_type_mismatch(instruction, source, "`literal real`", destination, dt)
+                    }
+                }
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
kalzoo, line 566: Excellent tests, thanks!

kalzoo (approved):
lgtm 🏁

kalzoo:
One thought, for any future reviewers to consider - would it make sense to return a separate error type from the type checks, so that (a) this isn't a breaking change and (b) because these errors can only be returned from `type_check` and are the only errors that can be returned from `type_check`? Then if we want to include them in `ProgramError` later they could be a `TypeError` variant and `thiserror` handles that well.

genos:
> One thought, for any future reviewers to consider - would it make sense to return a separate error type from the type checks, so that (a) this isn't a breaking change and (b) because these errors can only be returned from `type_check` and are the only errors that can be returned from `type_check`? Then if we want to include them in `ProgramError` later they could be a `TypeError` variant and `thiserror` handles that well.

That sounds nice; I'll make that change and revamp the `Breaking:` prefix.

genos:
I've changed the title to try to ensure this doesn't get picked up as a breaking change. Will that be enough for `semantic-release`? Should I modify the commit history here? Squash the commits upon merge? (cc @dbanty)

PR #89, opened 2022-09-01
refactor: make errors more human-readable

### LINKED ISSUE #51: Improve Error Messages
Currently, parsing errors look like this:

```
Error("InvalidCommand { command: DefCal, error: \"Parsing Error: ExpectedToken { actual: Colon, expected: \\\"Identifier\\\" } (at Colon)\" } (at Command(DefCal))")'
```

What caused this: `DEFCAL MEASURE 0:` should have been `DEFCAL MEASURE 0 addr:`.

Not great. With in-depth knowledge of the parser and a simple program, this is actionable information. However, with program lengths in the hundreds of lines or more, and without personal experience with this library, this is almost useless. Further, given that this library is under rapid development, there are still plenty of bugs, and that there are inconsistencies within the Quil spec itself, this can cause a jarring experience to someone who submits a valid program but cannot parse it.

Error messages should:

1. Include the line number within the input string at which the error was encountered. Bonus points for index within the line, if that comes as a package deal.
2. Display the readable string version of the offending instruction, not the debug version
3. Include a clear error message explaining what failed in lexing or parsing.

Shadow53:
Resolves #51.

The issue mentions using `miette`, but for now this work leverages the current error types.

- Track line and column information for strings (with `LocatedSpan`) and tokens (with `TokenWithLocation`)
  - Replace all parser inputs with the location-bundled versions
  - Add helper methods to keep existing code working
- `parser::error::{Error, ErrorKind}` now implement `Display` and `std::error::Error`
- Places where errors were converted to strings using their `Debug` impls were changed to use `Display`

Items not targeted by this PR:
- Conversion of errors to strings before propagating (should probably use `Box<dyn Error>` instead)
- Creating or using any new error types

Examples:

Only one for now, using the one from the issue:

Input:
```
DEFCAL MEASURE 0:
```
Old error:
```
Error("InvalidCommand { command: DefCal, error: \"Parsing Error: ExpectedToken { actual: Colon, expected: \\\"Identifier\\\" } (at Colon)\" } (at Command(DefCal))")'
```
New error:
```
Unrecoverable error while parsing: Error while parsing tokens at line 1, column 1 (Command(DefCal)): failed to parse arguments for DEFCAL: Parsing error: Error while parsing tokens at line 1, column 17 (Colon): expected Identifier, found Colon
```


notmgsk:
looking good 👍🏽

src/parser/gate.rs
```diff
@@ -22,18 +22,19 @@ use super::{
...
     let (input, name) = token!(Identifier(v))(input)?;
     let (input, parameters) = opt(delimited(
         token!(LParenthesis),
-        separated_list0(token!(Comma), parse_expression),
+        separated_list0(token!(Comma), ignore_leftover_error(parse_expression)),
         token!(RParenthesis),
     ))(input)?;
     let parameters = parameters.unwrap_or_default();
-    let (input, qubits) = many0(common::parse_qubit)(input)?;
+    let (input, qubits) = many0(ignore_leftover_error(common::parse_qubit))(input)?;
```
kalzoo, line 37: Overall, the need for such wide use of `ignore_leftover_error` isn't clear to me. In this particular case, for example, `parse_qubit` can't return a `LeftoverError` variant anyway, and an error for that wouldn't make sense in the conventions of `nom` anyway (erroring on leftover is more an opt-in thing with `all_consuming` but not really applicable elsewhere).

The fact that there are panics to handle this and that there are ~100 invocations of this parser modifier suggests to me that there's a neater way that doesn't make a leftover error an edge case that needs special handling everywhere. I'd probably have to spend some time hands-on with it, either solo or alongside you, to better understand what that might look like.

Shadow53: If it helps, here's an explanation of why `ignore_leftover_error` exists.

`(Parse)Error` (hereafter `Error`) contains a variant `Leftover` that contains a `LeftoverError { leftover: I, parsed: O }`. This allows consumers to optionally ignore that the input wasn't fully consumed and use the parsed value anyways.

This is stored in `Error` because it can be returned along with the other error variants. Because the input and returned types can change (and because sometimes other errors can occur), `Error` is defined as `Error<I, O, E>`.

`ignore_leftover_error` and `Error::convert_not_leftover` convert `Error<I, O, E>` into an `Error<I, O2, E>`. _This_ is necessary because if `parser1(I) -> IResult<I, O2, Error<I, O2, E>>` uses `parser2(I) -> IResult<I, O, Error<I, O, E>>`, the error type `O2` needs to be changed to `O`, which is essentially free if `Error::Leftover` is impossible.

----------

That said, taking a second look at this, it looks possible that `LeftoverError` can be moved from `Error` to `ProgramError`. I thought it was used in at least one place where `ProgramError` _wasn't_ used, but I couldn't find it through quick search through the code.

I'll take a look tomorrow at trying the refactor. If it works, that will simplify a lot of these changes.

Shadow53:
Turns out the refactor was in fact possible, and is now complete.

src/parser/token.rs
```diff
@@ -0,0 +1,167 @@
...
+}
+
+#[derive(Clone, PartialEq)]
+pub enum Token {
+    As,
+    Colon,
+    Comma,
+    Command(Command),
+    Comment(String),
+    // Constant(Constant),
```
kalzoo, line 75: I know this was pre-existing, but since we're reworking this section please clean it out. I can't recall _exactly_ but my guess is that I changed from lexing `i` and `pi` as tokens to just treating them as identifiers and then special-casing them in parsing.

kalzoo (changes requested):
Good change, @Shadow53 ! Reads much more cleanly. I'm not totally sure about having multiple parsing-related variants on `ProgramError` (since the program's user probably doesn't care if it was a failure to lex or to parse), so there may be a need for one layer of hierarchy in between the two, but that's stylistic and I'm not wedded to it.

What I'd need to see for a 👍:

- Docstrings on all the new non-trivial functions you've added
- No net-new TODOs that don't have github issue links embedded

src/parser/token.rs
```diff
@@ -0,0 +1,167 @@
...
+pub(crate) fn token_with_location<'i, E, P>(
+    mut parser: P,
+) -> impl FnMut(LexInput<'i>) -> LexResult<'i, TokenWithLocation, E>
+where
+    P: nom::Parser<LexInput<'i>, Token, E>,
+    E: nom::error::ParseError<LexInput<'i>>,
+{
+    move |input| {
+        let line = input.location_line();
+        // TODO: naive_get_utf8_column might be faster for shorter lines
```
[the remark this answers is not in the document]
Shadow53: Created #93 to track this

src/program/mod.rs
```diff
@@ -191,23 +198,22 @@ impl Program {
...
-        let mut program = Self::new();
-
-        for instruction in instructions {
-            program.add_instruction(instruction)
-        }
-
-        Ok(program)
+        let lexed = lex(s).map_err(ProgramError::LexError)?;
+        parse_instructions(&lexed)
+            // Will panic if error is Err::Incomplete, which only happens if using streaming variants of
```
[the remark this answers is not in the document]
Shadow53: Removed in recent commits

src/parser/gate.rs
Shadow53: Removed in recent commits.

Shadow53:
Made all requested changes. CI should also pass now because I ran `cargo fmt`.

src/program/error/syntax.rs
```diff
@@ -0,0 +1,49 @@
...
+/// The standard [`Display`](fmt::Display) implementation shows only the initial error.
+/// Use the alternative syntax (e.g., `format!("{:#}", err)` instead of `format!("{}", err)`) to
+/// see a backtrace of errors that caused this one.
+///
+/// See also [`Error`](crate::parser::Error).
+#[derive(Debug, PartialEq)]
+pub enum SyntaxError {
+    LexError(LexError),
+    ParsingError(ParseError),
+}
```
kalzoo, line 18:
```suggestion
pub enum SyntaxError {
    LexError(LexError),
    ParseError(ParseError),
}
```

Matching tense - `Lexing` and `Parsing` or `Lex` and `Parse`

src/program/error/syntax.rs
```diff
@@ -0,0 +1,49 @@
+use std::{error::Error, fmt};
+
+use crate::parser::{LexError, ParseError};
+
+/// Failed to parse a Quil program for some reason.
```
kalzoo, line 5:
```suggestion
/// Failed to deserialize a Quil program for some reason.
```

Just to avoid two diverging meanings of parse in the same file

src/program/mod.rs
```diff
@@ -93,12 +95,12 @@ impl Program {
...
     /// unchanged. Recurses though each instruction while ensuring there is no cycle in the expansion
     /// graph (i.e. no calibration expands directly or indirectly into itself)
-    pub fn expand_calibrations(&self) -> error::ProgramResult<Self> {
+    pub fn expand_calibrations(&self) -> Result<Self> {
         let mut expanded_instructions: Vec<Instruction> = vec![];
 
         // TODO: Do this more efficiently, possibly with Vec::splice
         for instruction in &self.instructions {
-            match self.calibrations.expand(instruction, &[])? {
+            match convert_leftover(self.calibrations.expand(instruction, &[]))? {
```
kalzoo, line 103: parsing-related utilities seem to be leaking out of parsing-related code here

src/program/error/mod.rs
```diff
@@ -0,0 +1,114 @@
...
+/// Errors that may occur while parsing a [`Program`](crate::program::Program).
+#[derive(Debug, PartialEq)]
+pub enum ProgramError<T> {
+    InvalidCalibration {
+        instruction: Instruction,
+        message: String,
+    },
+    RecursiveCalibration(Instruction),
+    Syntax(SyntaxError),
+    Leftover(LeftoverError<T>),
```
kalzoo, line 38: Leftover is also a syntax / parsing-related error. Shouldn't that be down at that level?

kalzoo (changes requested):
Pretty close, but still looks like the concern of parsing a string is finding its way into unrelated parts of the code. 

Also, I noticed that some of the new files are missing the copyright header. Can you add that?

src/program/mod.rs
Shadow53: This may generally be parsing related, but the function itself is defined in `src/program/error/result.rs`, which is not a parsing-specific module. There isn't really a better way to handle this, IMO. `expand()` returns an error of `ProgramError<Option<Vec<Instruction>>>`, while `expand_calibrations()` returns `ProgramError<Program>`.

I've made a commit that is the only other fix to this that I can think of -- changing the error to `ProgramError<Program>`. I initially avoided this because `expand()` is `pub fn` and so, as far as I can tell, part of the public API. As such, making the error `ProgramError<Program>` feels off, even if it works in the type system.

This commit works because the only callers of `expand()` are `expand()` itself and `expand_calibrations()`.

src/program/error/mod.rs
Shadow53: It could be, yeah. I don't have a preference, so I'll make the change.

src/program/mod.rs
Shadow53: This change might actually be preferable, because it's the only use of `convert_leftover`, the removal of which made moving `Leftover` from `ProgramError` to `SyntaxError` easier to do.

kalzoo (approved):
Looks good, thanks for the creativity, @Shadow53 ! I agree that `ProgramError<Program>` or just having it be parametric at all seems unusual, but it's in so few places that we can change it later if we want without it getting in the way (even if that will be a breaking change). I'd like to get this out to consumers - so let's 🚢 it!

PR #92, opened 2022-09-14
feat: Parse DEFGATE instructions

randall-fulton:
Add parser support `DEFGATE`.

### DIFF
--- a/src/instruction.rs
+++ b/src/instruction.rs
@@ -1074,12 +1074,13 @@ impl Instruction {
             Instruction::WaveformDefinition(WaveformDefinition { definition, .. }) => {
                 definition.matrix.iter_mut().for_each(closure);
             }
-            Instruction::GateDefinition(GateDefinition { specification, .. }) => {
-                if let GateSpecification::Matrix(matrix) = specification {
-                    for row in matrix {
-                        for cell in row {
-                            closure(cell);
-                        }
+            Instruction::GateDefinition(GateDefinition {
+                specification: GateSpecification::Matrix(matrix),
+                ..
+            }) => {
+                for row in matrix {
+                    for cell in row {
+                        closure(cell);
                     }
                 }
             }
--- a/src/instruction.rs
+++ b/src/instruction.rs
@@ -322,8 +322,8 @@ pub struct CircuitDefinition {
 pub struct GateDefinition {
     pub name: String,
     pub parameters: Vec<String>,
-    pub matrix: Vec<Vec<Expression>>,
-    pub permutation: Vec<u64>,
+    pub matrix: Option<Vec<Vec<Expression>>>,
+    pub permutation: Option<Vec<u64>>,
     pub r#type: GateType,
 }
 
@@ -783,27 +783,31 @@ impl fmt::Display for Instruction {
                 writeln!(f, "DEFGATE {}{} AS {}:", name, parameter_str, r#type)?;
                 match r#type {
                     GateType::Matrix => {
-                        for row in matrix {
-                            writeln!(
-                                f,
-                                "\t{}",
-                                row.iter()
-                                    .map(|cell| format!("{}", cell))
-                                    .collect::<Vec<String>>()
-                                    .join(",")
-                            )?;
+                        if let Some(matrix) = matrix {
+                            for row in matrix {
+                                writeln!(
+                                    f,
+                                    "\t{}",
+                                    row.iter()
+                                        .map(|cell| format!("{}", cell))
+                                        .collect::<Vec<String>>()
+                                        .join(",")
+                                )?;
+                            }
                         }
                     }
                     GateType::Permutation => {
-                        writeln!(
-                            f,
-                            "\t{}",
-                            permutation
-                                .iter()
-                                .map(|i| i.to_string())
-                                .collect::<Vec<String>>()
-                                .join(", ")
-                        )?;
+                        if let Some(permutation) = permutation {
+                            writeln!(
+                                f,
+                                "\t{}",
+                                permutation
+                                    .iter()
+                                    .map(|i| i.to_string())
+                                    .collect::<Vec<String>>()
+                                    .join(", ")
+                            )?;
+                        }
                     }
                 }
                 Ok(())
@@ -1071,9 +1075,11 @@ impl Instruction {
                 definition.matrix.iter_mut().for_each(closure);
             }
             Instruction::GateDefinition(GateDefinition { matrix, .. }) => {
-                for row in matrix {
-                    for cell in row {
-                        closure(cell);
+                if let Some(matrix) = matrix {
+                    for row in matrix {
+                        for cell in row {
+                            closure(cell);
+                        }
                     }
                 }
             }
--- a/src/instruction.rs
+++ b/src/instruction.rs
@@ -319,11 +319,16 @@ pub struct CircuitDefinition {
 }
 
 #[derive(Clone, Debug, PartialEq, Eq)]
+pub enum GateSpecification {
+    Matrix(Vec<Vec<Expression>>),
+    Permutation(Vec<u64>),
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
 pub struct GateDefinition {
     pub name: String,
     pub parameters: Vec<String>,
-    pub matrix: Option<Vec<Vec<Expression>>>,
-    pub permutation: Option<Vec<u64>>,
+    pub specification: GateSpecification,
     pub r#type: GateType,
 }
 
@@ -775,39 +780,34 @@ impl fmt::Display for Instruction {
             Instruction::GateDefinition(GateDefinition {
                 name,
                 parameters,
-                matrix,
-                permutation,
+                specification,
                 r#type,
             }) => {
                 let parameter_str: String = parameters.iter().map(|p| p.to_string()).collect();
                 writeln!(f, "DEFGATE {}{} AS {}:", name, parameter_str, r#type)?;
-                match r#type {
-                    GateType::Matrix => {
-                        if let Some(matrix) = matrix {
-                            for row in matrix {
-                                writeln!(
-                                    f,
-                                    "\t{}",
-                                    row.iter()
-                                        .map(|cell| format!("{}", cell))
-                                        .collect::<Vec<String>>()
-                                        .join(",")
-                                )?;
-                            }
-                        }
-                    }
-                    GateType::Permutation => {
-                        if let Some(permutation) = permutation {
+                match specification {
+                    GateSpecification::Matrix(matrix) => {
+                        for row in matrix {
                             writeln!(
                                 f,
                                 "\t{}",
-                                permutation
-                                    .iter()
-                                    .map(|i| i.to_string())
+                                row.iter()
+                                    .map(|cell| format!("{}", cell))
                                     .collect::<Vec<String>>()
-                                    .join(", ")
+                                    .join(",")
                             )?;
                         }
+                    }
+                    GateSpecification::Permutation(permutation) => {
+                        writeln!(
+                            f,
+                            "\t{}",
+                            permutation
+                                .iter()
+                                .map(|i| i.to_string())
+                                .collect::<Vec<String>>()
+                                .join(", ")
+                        )?;
                     }
                 }
                 Ok(())
@@ -1074,8 +1074,8 @@ impl Instruction {
             Instruction::WaveformDefinition(WaveformDefinition { definition, .. }) => {
                 definition.matrix.iter_mut().for_each(closure);
             }
-            Instruction::GateDefinition(GateDefinition { matrix, .. }) => {
-                if let Some(matrix) = matrix {
+            Instruction::GateDefinition(GateDefinition { specification, .. }) => {
+                if let GateSpecification::Matrix(matrix) = specification {
                     for row in matrix {
                         for cell in row {
                             closure(cell);
--- a/src/instruction.rs
+++ b/src/instruction.rs
@@ -323,6 +323,7 @@ pub struct GateDefinition {
     pub name: String,
     pub parameters: Vec<String>,
     pub matrix: Vec<Vec<Expression>>,
+    pub permutation: Vec<u64>,
     pub r#type: GateType,
 }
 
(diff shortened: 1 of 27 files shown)


src/parser/command.rs
```diff
@@ -223,6 +228,42 @@ pub fn parse_defframe<'a>(input: ParserInput<'a>) -> ParserResult<'a, Instructio
...
+        token!(RParenthesis),
+    ))(input)?;
+    let (input, gate_type) = opt(preceded(
+        token!(As),
+        alt((
+            map(token!(Matrix), |()| GateType::Matrix),
+            map(token!(Permutation), |()| GateType::Permutation),
+        )),
+    ))(input)?;
+    let gate_type = gate_type.unwrap_or(GateType::Matrix);
```
notmgsk, line 246: Think `unwrap_or_else` is generally preferred

src/parser/command.rs
```diff
@@ -668,4 +711,140 @@ mod tests {
             ]
         })
     );
+
+    make_test!(
+        defgate,
+        parse_defgate,
+        "HADAMARD:\n\t1/sqrt(2), 1/sqrt(2)\n\t1/sqrt(2), -1/sqrt(2)",
```
notmgsk, line 718: Using raw strings might make this more readable 
```
#r"HADAMARD:
    1/sqrt(2), 1/sqrt(2)
    1/sqrt(2), -1/sqrt(2)
#;

File: src/parser/common.rs [hunk 3]
```diff
@@ -373,4 +415,65 @@ mod tests {
...
+        let (remainder, matrix) = parse_permutation(&lexed).unwrap();
+        assert!(
+            remainder.is_empty(),
+            "expected remainder to be empty, got {:?}",
+            remainder
+        );
+        assert_eq!(
+            matrix,
+            vec![vec![
+                Expression::Number(real!(0.0)),
```
notmgsk, line 445: Ideally, these would parse as non-negative integers. Not sure how feasible that is, though.

notmgsk (commented):
Looking good so far

src/parser/command.rs
randall-fulton: I'd have to provide a function to `unwrap_or_else`, even though I need to return a constant value. That would be more verbose for no gain.

src/parser/common.rs
randall-fulton: This is where I struggled the most. I looked into adding integers as a valid expression, but that requires trickles into a lot of other places. Once I got to `evaluate()` I realized it would have to ultimately be converted into a float and that just felt weird.

I'll take any suggestions though.

src/parser/command.rs
Shadow53: Correct. `unwrap_or_else` is preferred when there is a function call being made, so that the function call is evaluated lazily instead of eagerly.

src/parser/command.rs
```diff
@@ -656,4 +698,145 @@ mod tests {
...
+        parse_defgate,
+        r#"H:
+            1/sqrt(2), 1/sqrt(2)
+            1/sqrt(2), -1/sqrt(2)"#,
+        Instruction::GateDefinition(GateDefinition {
+            name: "H".to_string(),
+            parameters: vec![],
+            matrix: vec![
+                vec![
+                    Expression::Infix {
```
kalzoo, line 713: For brevity, you could define this once and clone it

src/parser/common.rs
```diff
@@ -155,6 +155,43 @@ pub fn parse_gate_modifier<'a>(input: ParserInput<'a>) -> ParserResult<'a, GateM
...
+        preceded(
+            many1(token!(Indentation)),
+            separated_list0(token!(Comma), parse_expression),
+        ),
+    )(input)?;
+    Ok((input, matrix))
+}
+
+/// Parse permutation representation of a `DEFGATE` matrix.
+pub fn parse_permutation<'a>(input: ParserInput<'a>) -> ParserResult<'a, Vec<Vec<Expression>>> {
```
kalzoo, line 171: Per the spec, this is a comma-separated list of integers, not expressions: https://quil-lang.github.io/ sect 4.2.3

src/parser/common.rs
```diff
@@ -155,6 +155,43 @@ pub fn parse_gate_modifier<'a>(input: ParserInput<'a>) -> ParserResult<'a, GateM
     ))
 }
 
+/// Parse matrix used to define gate with `DEFGATE`.
+pub fn parse_matrix<'a>(input: ParserInput<'a>) -> ParserResult<'a, Vec<Vec<Expression>>> {
+    let (input, matrix) = separated_list0(
+        token!(NewLine),
+        preceded(
+            many1(token!(Indentation)),
```
kalzoo, line 163: I believe this should be exactly one indentation. Likewise with permutation below. See `parse_block_instruction` and just swap `parse_instruction` for the expected line contents here.

On that note, `parse_indented_block` could be its own useful combinator here which takes the line's parser as an argument... would be used in 4 places as of this PR:

* permutation
* matrix
* waveform
* block_instruction

src/parser/common.rs
```diff
@@ -155,6 +155,43 @@ pub fn parse_gate_modifier<'a>(input: ParserInput<'a>) -> ParserResult<'a, GateM
...
+        ),
+    )(input)?;
+    Ok((input, matrix))
+}
+
+/// Parse permutation representation of a `DEFGATE` matrix.
+pub fn parse_permutation<'a>(input: ParserInput<'a>) -> ParserResult<'a, Vec<Vec<Expression>>> {
+    let (input, matrix) = preceded(
+        many1(token!(Indentation)),
+        separated_list0(token!(Comma), parse_expression),
```
kalzoo, line 174: Per the spec, a minimum of one item (integer) is expected

src/parser/common.rs
kalzoo: An integer is a token already - don't need to consider it an expression. See `parse_expression` to see that it's handling integer tokens internally, which you can do directly

kalzoo (changes requested):

Shadow53:
Note: my PR just got merged, so you should be able to merge/rebase on `main` one more time.

src/parser/common.rs
randall-fulton: Ah, I missed that. Thank you @kalzoo. That was the one thing I really disliked about the implementation I have.

src/parser/command.rs
randall-fulton: Don't believe that is possible. We can't initiailize a `Box` within a constant expression.

src/parser/command.rs
```diff
@@ -213,6 +217,52 @@ pub fn parse_defframe<'a>(input: ParserInput<'a>) -> ParserResult<'a, Instructio
...
+            let (input, m) = parse_matrix(input)?;
+            matrix = Some(m);
+            input
+        }
+        GateType::Permutation => {
+            let (input, p) = parse_permutation(input)?;
+            permutation = Some(p);
+            input
+        }
+    };
```
randall-fulton, line 252: Is there a more idiomatic way to handle this case where I need to initialize two mutually exclusive fields based on a switch variable?

genos:
```rust
let (input, matrix, permutation) = match gate_type {
    GateType::Matrix => {
        let (input, m) = parse_matrix(input)?;
        (input, Some(m), None)
    }
    GateType::Permutation {
        let (input, p) = parse_permutation(input)?;
        (input, None, Some(p))
    }
};

Comment (randall-fulton) [reply src/parser/command.rs:252]: 🔥 
It was staring me right in the face...

File: src/parser/command.rs (hunk 4 above)
Comment (kalzoo) [reply src/parser/command.rs:719]: Indeed it is! Check out #97 . The last macro argument is just an expression, which can be wrapped in a block and remain an expression - if that's what you mean in your comment.

File: src/parser/common.rs (hunk 6 above)
Comment (kalzoo) [reply src/parser/common.rs]: Yeah, this would be a good one! Time might be hard to come by in the near future, but let's earmark this one for a later 1x1. We can punt it for this PR.

**Review by kalzoo — CHANGES_REQUESTED** [review]
Looking good! Let's get the `GateDefinition` type right while we're here, even though that was pre-existing. After that, should be good to 🚢

File: src/parser/command.rs (hunk 4 above)
Comment (randall-fulton) [reply src/parser/command.rs:719]: Oh! That's super cool.

My immediate thought was raising it to a `const` and reusing it. In that case, you can't use the `Box`. I never even considered a block that returns the final value. Very slick.

File: src/parser/command.rs [hunk 9]
```diff
@@ -213,6 +217,48 @@ pub fn parse_defframe<'a>(input: ParserInput<'a>) -> ParserResult<'a, Instructio
...
+        input,
+        Instruction::GateDefinition(GateDefinition {
+            name,
+            parameters: parameters.unwrap_or_default(),
+            matrix,
+            permutation,
+            r#type: gate_type,
+        }),
+    ))
+}
```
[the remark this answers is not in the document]
randall-fulton: With this change in place, what is your opinion on removing `r#type` on the `GateDefinition`?

The only place it is used is the instruction's `impl Display`, but we could get the same information by checking which variant of `GateSpecification` we have. At the moment, it is possible for `r#type` and `specification` to be out of sync, so I'm leaning towards removing `r#type` altogether.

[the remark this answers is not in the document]
kalzoo: I agree, it's no longer needed as the information is now embedded in the enum variant.

src/instruction.rs
```diff
@@ -774,20 +780,35 @@ impl fmt::Display for Instruction {
...
-                            .join(",")
-                    )?;
+                match specification {
+                    GateSpecification::Matrix(matrix) => {
+                        for row in matrix {
+                            writeln!(
+                                f,
+                                "\t{}",
+                                row.iter()
+                                    .map(|cell| format!("{}", cell))
```
kalzoo, line 795: nit: given that `format!("{}", cell)` and `cell.to_string()` are the same, we should stick to one or the other.

src/parser/command.rs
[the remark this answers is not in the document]
randall-fulton: I think we still keep `GateType` here. While it isn't needed for `GateDefinition`, it's used as part of parsing a gate. We could remove it there too, but we'd have to either replace it with a string or have the parsing of type return a function to build the specification (instead of a flag that tells you what kind to build).

To me, both of those feel messier than just retaining `GateType`. Open to any other ideas though, or if someone thinks the approaches I lay out above are actually more idiomatic here.

src/instruction.rs
Shadow53: I'd prefer `to_string()` for converting a single item to string and leave `format!()` for all other cases (building strings from multiple values, etc.)

src/parser/command.rs
```diff
@@ -213,6 +220,42 @@ pub fn parse_defframe<'a>(input: ParserInput<'a>) -> ParserResult<'a, Instructio
...
+pub fn parse_defgate<'a>(input: ParserInput<'a>) -> ParserResult<'a, Instruction> {
+    let (input, name) = token!(Identifier(v))(input)?;
+    let (input, parameters) = opt(delimited(
+        token!(LParenthesis),
+        separated_list1(token!(Comma), token!(Variable(v))),
+        token!(RParenthesis),
+    ))(input)?;
+    let (input, gate_type) = opt(preceded(
+        token!(As),
+        alt((
```
Shadow53, line 233: Should use the `wrapped_parsers` version of `alt`, which gives a much better error message.

src/parser/command.rs
```diff
@@ -213,6 +220,42 @@ pub fn parse_defframe<'a>(input: ParserInput<'a>) -> ParserResult<'a, Instructio
...
+            map(token!(Matrix), |()| GateType::Matrix),
+            map(token!(Permutation), |()| GateType::Permutation),
+        )),
+    ))(input)?;
+    let (input, _) = token!(Colon)(input)?;
+
+    let gate_type = gate_type.unwrap_or(GateType::Matrix);
+    let (input, specification) = match gate_type {
+        GateType::Matrix => {
+            parse_matrix(input).map(|(input, matrix)| (input, GateSpecification::Matrix(matrix)))?
```
Shadow53, line 243: The `map` combinator from `nom` allows you to map just the output without worrying about the leftover input. I think you could rewrite this line as:

```
map(parse_matrix, GateSpecification::Matrix)(input)
```

src/parser/command.rs
```diff
@@ -213,6 +220,42 @@ pub fn parse_defframe<'a>(input: ParserInput<'a>) -> ParserResult<'a, Instructio
...
+        }
+        GateType::Permutation => parse_permutation(input)
+            .map(|(input, permutation)| (input, GateSpecification::Permutation(permutation)))?,
```
Shadow53, line 246: Similar for here

src/parser/error/input.rs
```diff
@@ -74,6 +74,24 @@ impl ErrorInput for ParserInput<'_> {
...
+
+    fn snippet(&self) -> String {
+        self.as_slice().snippet()
+    }
+
+    fn is_empty(&self) -> bool {
+        self.as_slice().is_empty()
+    }
+}
+
```
Shadow53, line 94: Minor nit, but this feels like avoidable duplication. Should double check that this works, but I think you can do this to delegate to the non-reference version:

```suggestion
impl ErrorInput for &Vec<TokenWithLocation> {
    fn line(&self) -> u32 {
        (*self).line()
    }

    fn column(&self) -> usize {
        (*self).column()
    }

    fn snippet(&self) -> String {
        (*self).snippet()
    }

    fn is_empty(&self) -> bool {
        (*self).is_empty()
    }
}

```

Shadow53 (changes requested):
Aside from the above, there are a few `cargo check` warnings that are showing up in the diff.

src/parser/error/input.rs
randall-fulton: I think this is just an artifact of trying to reconcile my changes with rebasing against your branch. There were quite a lot of conflicts I had to manually resolve. Looks like this one was done incorrectly.

I'll try to remove it and make sure that doesn't cause issues.

src/parser/command.rs
randall-fulton: That version expects a `LexInput`, not a `ParserInput`. It's specific to lexing.

notmgsk (approved):
:ship:

Issue #96, opened 2022-09-16
Vend public parser for MemoryReference and Expression


kalzoo:
In the course of program execution with the Rust SDK, there's a need to be able to parse and interpret a memory reference, such as `ro[0]`. While this can be done with some simple string manipulation, it would be less brittle to use our purpose-built Quil parser.

However, this library does not offer an entrypoint to parse a memory reference on its own. `pyQuil` solves this problem for expressions by inserting the string-to-be-parsed into a gate instruction, parsing _that_ as a program, and then extracting the expression value ... but this is inelegant.

AC: this library exports functions allowing for the parsing of MemoryReference and Expression from strings. These should be all-consuming and not allow for leftovers. IMO, this would be best done using `FromStr`.

jselig-rigetti:
`Expression::from_str` already exists: https://github.com/rigetti/quil-rs/blob/main/src/expression.rs#L414-L421

PR #98, opened 2022-09-22
feat: impl FromStr for MemoryReference

jselig-rigetti:
Closes https://github.com/rigetti/quil-rs/issues/96

Note that `Expression::from_str` already exists at https://github.com/rigetti/quil-rs/blob/main/src/expression.rs#L414-L421

### For consideration:
Despite https://github.com/quil-lang/quil/blob/master/spec/Quil.md#names having restrictions on leading and trailing `-`, while testing I saw that `parse_memory_reference_with_brackets` will actually accept trailing `-`, and `\` is not mentioned but is allowed. For example `_\\-[0]` will parse into a `MemoryReference`.

This is due to how identifiers are lexed, see https://github.com/rigetti/quil-rs/blob/main/src/parser/lexer/mod.rs#L246-L265:

```rs
//!! The `\` is not mentioned in the spec, and `-` is not valid in the last position
fn is_valid_identifier_character(chr: char) -> bool {
    is_valid_identifier_leading_character(chr) || chr.is_ascii_digit() || chr == '\\' || chr == '-'
}

fn is_valid_identifier_leading_character(chr: char) -> bool {
    chr.is_ascii_alphabetic() || chr == '_'
}

fn lex_identifier_raw(input: LexInput) -> LexResult<String> {
    expecting(
        "a valid identifier",
        map(
            tuple::<_, _, InternalLexError, _>((
                take_while1(is_valid_identifier_leading_character),
                take_while(is_valid_identifier_character),
                //!! The last position should be a different branch
            )),
            |(left, right)| format!("{}{}", left, right),
        ),
    )(input)
}
```

My guess is that there's no real issue here, and the spec should probably just be updated. The behavior is present in `Program::from_str` so e.g. the following works:
```sh
        let input = "
DECLARE _\\- BIT
MEASURE q _\\-
";
        dbg!(Program::from_str(input).expect("").memory_regions);
```

Output:
```
Program::from_str(input).expect("").memory_regions = {
    "_\\-": MemoryRegion {
        size: Vector {
            data_type: Bit,
            length: 1,
        },
        sharing: None,
    },
}
```

### DIFF
--- a/src/instruction.rs
+++ b/src/instruction.rs
@@ -1165,6 +1165,7 @@ impl Instruction {
 
 #[cfg(test)]
 mod tests {
+    use rstest::*;
     use std::str::FromStr;
 
     use crate::{expression::Expression, Program};
@@ -1194,50 +1195,35 @@ RX(%a) 0",
         assert_eq!(expected_program, program);
     }
 
-    #[test]
-    fn it_parses_memory_reference_from_str() {
-        [
-            "_",
-            "a",
-            "a---b",
-            "_a_b_",
-            "a-2_b-2",
-            "_[0]",
-            "a[1]",
-            "a---b[2]",
-            "_a_b_[3]",
-            "a-2_b-2[4]",
-        ]
-        .iter()
-        .for_each(|&s| {
-            assert!(
-                MemoryReference::from_str(s).is_ok(),
-                "MemoryReference should have parsed: {}",
-                s
-            )
-        });
+    #[rstest(input, expected,
+        case("_", MemoryReference { name: "_".to_string(), index: 0 }),
+        case("a", MemoryReference { name: "a".to_string(), index: 0 }),
+        case("a---b", MemoryReference { name: "a---b".to_string(), index: 0 }),
+        case("_a_b_", MemoryReference { name: "_a_b_".to_string(), index: 0 }),
+        case("a-2_b-2", MemoryReference { name: "a-2_b-2".to_string(), index: 0 }),
+        case("_[0]", MemoryReference { name: "_".to_string(), index: 0 }),
+        case("a[1]", MemoryReference { name: "a".to_string(), index: 1 }),
+        case("a---b[2]", MemoryReference { name: "a---b".to_string(), index: 2 }),
+        case("_a_b_[3]", MemoryReference { name: "_a_b_".to_string(), index: 3 }),
+        case("a-2_b-2[4]", MemoryReference { name: "a-2_b-2".to_string(), index: 4 }),
+    )]
+    fn it_parses_memory_reference_from_str(input: &str, expected: MemoryReference) {
+        assert_eq!(MemoryReference::from_str(input), Ok(expected));
     }
 
-    #[test]
-    fn it_fails_to_parse_memory_reference_from_str() {
-        [
-            "",
-            "[0]",
-            "a[-1]",
-            "2a[2]",
-            "-a",
-            "NOT[3]",
-            "a a",
-            "a[5] a[5]",
-            "DECLARE a[6]",
-        ]
-        .iter()
-        .for_each(|&s| {
-            assert!(
-                MemoryReference::from_str(s).is_err(),
-                "MemoryReference should not have parsed: {}",
-                s
-            )
-        });
+    #[rstest(
+        input,
+        case(""),
+        case("[0]"),
+        case("a[-1]"),
+        case("2a[2]"),
+        case("-a"),
+        case("NOT[3]"),
+        case("a a"),
+        case("a[5] a[5]"),
+        case("DECLARE a[6]")
+    )]
+    fn it_fails_to_parse_memory_reference_from_str(input: &str) {
+        assert!(MemoryReference::from_str(input).is_err());
     }
 }


src/instruction.rs
```diff
@@ -301,6 +304,17 @@ impl fmt::Display for MemoryReference {
...
 }
 
+impl FromStr for MemoryReference {
+    type Err = SyntaxError<Self>;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let tokens = lex(s)?;
+        let (_, memory_reference) = all_consuming(parse_memory_reference_with_brackets)(&tokens)
+            .map_err(extract_nom_err)?;
+        Ok(memory_reference)
```
Shadow53, line 314: Need to double check, but I think you also want a `disallow_leftover()` here, maybe _instead_ of `all_consuming`.

It's the same effect, but `disallow_leftover` gives a better error and removes the need for `map_err`.

```suggestion
        disallow_leftover(parse_memory_reference_with_brackets(&tokens))
```

Shadow53: To make it work in this context, you'll need to change `disalllow_leftover` to be this. (Alternatively, change the error type to `ProgramError<T>`, but this is the more correct option, IMO)

```rust
pub fn disallow_leftover<I, O, E>(result: nom::IResult<I, O, ParseError>) -> Result<O, E>
where
    I: ErrorInput,
    E: From<SyntaxError<O>>,
{
    match result.finish() {
        Ok((leftover, parsed)) => {
            if leftover.is_empty() {
                Ok(parsed)
            } else {
                Err(E::from(SyntaxError::from(LeftoverError::new(leftover, parsed))))
            }
        }
        Err(err) => Err(E::from(SyntaxError::from(err))),
    }
}
```

src/instruction.rs
```diff
@@ -301,6 +304,17 @@ impl fmt::Display for MemoryReference {
     }
 }
 
+impl FromStr for MemoryReference {
+    type Err = SyntaxError<Self>;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let tokens = lex(s)?;
+        let (_, memory_reference) = all_consuming(parse_memory_reference_with_brackets)(&tokens)
```
Shadow53, line 312: The [quil spec](https://quil-lang.github.io) says in section 6.4 that a memory reference can be a lone identifier. Why are we (if I understand right) requiring the bracket syntax here instead of allowing the lone identifier?

Shadow53 (changes requested):

notmgsk:
@jselig-rigetti Refer to [this doc](https://quil-lang.github.io) as the most up-to-date version of the spec. In particular, an [identifier is defined](https://quil-lang.github.io/#3-2Syntactic-Rudiments) as
```
⟨Identifier⟩ ::= [A-Za-z_]
               | [A-Za-z_][A-Za-z0-9\-_]*[A-Za-z0-9_]
```

src/instruction.rs
jselig-rigetti: You're right - I had misread the source I was trying to de-duplicate in [qcs-sdk-rust/crates/lib/src/qpu/mod.rs: organize_ro_sources](https://github.com/rigetti/qcs-sdk-rust/blob/0a14eea26bbb2f4acdd323d5f56a8162ade41d8d/crates/lib/src/qpu/mod.rs#L133-L149) in that I had missed the part where a missing index was an implied `[0]`, and thought that this particular usage required the index.

jselig-rigetti:
> @jselig-rigetti Refer to [this doc](https://quil-lang.github.io) as the most up-to-date version of the spec. In particular, an [identifier is defined](https://quil-lang.github.io/#3-2Syntactic-Rudiments) as
> 
> ```
> ⟨Identifier⟩ ::= [A-Za-z_]
>                | [A-Za-z_][A-Za-z0-9\-_]*[A-Za-z0-9_]
> ```

@notmgsk right - the above regex should fail against `_\` and `_-` but [quil-rs/parser/lexer/mod.rs: lex_identifier_raw](https://github.com/rigetti/quil-rs/blob/main/src/parser/lexer/mod.rs#L246-L265) will accept it due to not having a distinct third group, looking more like: 
```
⟨Identifier⟩ ::= [A-Za-z_][A-Za-z0-9\-_]*
```
[pyquil/_parser/grammar.lark: IDENTIFIER](https://github.com/rigetti/pyquil/blob/master/pyquil/_parser/grammar.lark#L312) differs slightly from both in that it looks to not accept `\` in any position:
```
IDENTIFIER          : ("_"|LETTER) [ ("_"|"-"|LETTER|DIGIT)* ("_"|LETTER|DIGIT) ]
```

jselig-rigetti:
> if the tests pass, feel free to align it to the spec.

I'll avoid doing that here but capture it in https://github.com/rigetti/quil-rs/issues/99  since it is a "breaking" change.

I've removed the test cases from this MR that violate the spec though.

src/instruction.rs
```diff
@@ -1180,4 +1193,51 @@ RX(%a) 0",
...
+            "_[0]",
+            "a[1]",
+            "a---b[2]",
+            "_a_b_[3]",
+            "a-2_b-2[4]",
+        ]
+        .iter()
+        .for_each(|&s| {
+            assert!(
+                MemoryReference::from_str(s).is_ok(),
```
Shadow53, line 1214: Can you change this to assert that the parsed value is as expected, not just that it didn't error?

Shadow53 (changes requested):
Along with the one request, you should consider using either [`test_case`](https://crates.io/crates/test-case) or [`rstest` `case`](https://crates.io/crates/rstest) to parametrize the tests instead of using a loop. That way, we can see multiple failures at once instead of only the first.

notmgsk (approved):

PR #100, opened 2022-09-27
Fix: identifier parser

### LINKED ISSUE #99: Parser for IDENTIFIER is too permissive
From #98 

The [Quil spec for an Identifier](https://quil-lang.github.io/#3-2Syntactic-Rudiments) is as follows:
> Identifiers in Quil are alphanumeric Latin characters, along with hyphens and underscores.  
> Identifiers cannot start or end with a hyphen '-'.
> ```
> ⟨Identifier⟩ ::= [A-Za-z_]
>                | [A-Za-z_][A-Za-z0-9\-_]*[A-Za-z0-9_]
> ```

However, [quil-rs/parser/lexer/mod.rs: lex_identifier_raw](https://github.com/rigetti/quil-rs/blob/main/src/parser/lexer/mod.rs#L246-L265) seems to implement this: 
```
⟨Identifier⟩ ::= [A-Za-z_][A-Za-z0-9\\\-_]*
```

To align with the spec, `\` should not be an acceptable character and `-` should not be acceptable in the last position.

jselig-rigetti:
Closes #99

### DIFF
--- a/src/parser/lexer/mod.rs
+++ b/src/parser/lexer/mod.rs
@@ -243,12 +243,16 @@ fn recognize_command_or_identifier(identifier: String) -> Token {
     }
 }
 
-fn is_valid_identifier_character(chr: char) -> bool {
-    is_valid_identifier_leading_character(chr) || chr.is_ascii_digit() || chr == '\\' || chr == '-'
-}
-
 fn is_valid_identifier_leading_character(chr: char) -> bool {
     chr.is_ascii_alphabetic() || chr == '_'
+}
+
+fn is_valid_identifier_end_character(chr: char) -> bool {
+    is_valid_identifier_leading_character(chr) || chr.is_ascii_digit()
+}
+
+fn is_valid_identifier_middle_character(chr: char) -> bool {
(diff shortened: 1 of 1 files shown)


src/parser/lexer/mod.rs
```diff
@@ -243,25 +243,39 @@ fn recognize_command_or_identifier(identifier: String) -> Token {
...
+                take_while(is_valid_identifier_middle_character),
             )),
             |(left, right)| format!("{}{}", left, right),
         ),
     )(input)
+    .and_then(|(remaining, result)| {
+        if !result.ends_with(is_valid_identifier_end_character) {
+            Err(nom::Err::Failure(LexError::from_kind(
+                input,
+                LexErrorKind::ExpectedContext("valid identifier ending character"),
```
Shadow53, line 273: Passing `input` here means the error will start at the beginning of the identifier.
```suggestion
                LexErrorKind::ExpectedContext("valid identifier"),
```

Shadow53 (changes requested):
Looks good, just one requested change to make an error message less potentially confusing.

notmgsk (approved):

PR #110, opened 2022-10-12
Breaking: support integer arguments for PRAGMA

### LINKED ISSUE #109: Allow PRAGMA's second parameter to be an int
Per the [Quil spec](https://quil-lang.github.io/#9-2Pragmas), pragmas have the form
```
⟨Pragma⟩ ::= PRAGMA ⟨Identifier⟩(⟨Identifier⟩ | ⟨Integer⟩)*⟨String⟩?
```

Currently we do not support the `<Integer>` part.

notmgsk:
I marked this as `Breaking` because it changes the public type `Pragma`. I suppose _technically_ this could actually be considered a `Fix` since it was a bug...

Closes #109.

### DIFF
--- a/src/instruction.rs
+++ b/src/instruction.rs
@@ -396,8 +396,14 @@ pub struct MeasureCalibrationDefinition {
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct Pragma {
     pub name: String,
-    pub arguments: Vec<String>,
+    pub arguments: Vec<PragmaArgument>,
     pub data: Option<String>,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum PragmaArgument {
+    Identifier(String),
+    Integer(u64),
 }
 
 #[derive(Clone, Debug, PartialEq, Eq)]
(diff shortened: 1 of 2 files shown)


src/instruction.rs
```diff
@@ -892,7 +898,15 @@ impl fmt::Display for Instruction {
...
-                    write!(f, " {}", arguments.join(" "))?;
+                    let arguments = arguments
+                        .iter()
+                        .map(|arg| match arg {
+                            PragmaArgument::Identifier(i) => i.clone(),
+                            PragmaArgument::Integer(i) => format!("{}", i),
+                        })
+                        .collect::<Vec<_>>()
+                        .join(" ");
+                    write!(f, " {}", arguments)?;
```
Shadow53, line 909: GitHub won't let me use the suggestion feature, so here's a code snippet that avoids extra string allocations:

```rust
for arg in &arguments {
    write!(f, " {}", arg)?;
}
```

Then elsewhere, add

```rust
impl fmt::Display for PragmaArgument {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        match self {
            Self::Identifier(id) => write!(f, "{}", id),
            Self::Integer(i) => write!(f, "{}", i),
        }
    }
}
```

Shadow53: (Could also just do the `match` inside of the `for` loop, but I prefer this since `PragmaArgument` is a `pub` type)

Shadow53 (changes requested):
Looks good. Only nit is that the `impl Display for Instruction` is inefficient and could be improved and simplified.

PR #113, opened 2022-10-24
fix: fix performance regression

Shadow53:
Rough stats for these changes. Timing is based on parsing the sample calibrations file.

| Before regression  | Before Fix                  | After fix   | After, with `lto=fat,codegen-units=1`|
|  ----------------- | --------------------- | --------- | ------------------------------------- |
| ~40ms                    | ~30,000-60,000ms   |  ~230ms | ~185 ms                                                 |

Parsing is generally more expensive now because we are tracking location information while parsing, which adds some overhead (though not as much after these changes).

All of the `pub fn parse_*` functions were already crate-private by being marked `pub` in a private module. I just made them `pub(crate)` for clarity, since I was modifying return types already.

Edit: thousands of milliseconds, not thousands of seconds

### DIFF
--- a/src/parser/error/internal.rs
+++ b/src/parser/error/internal.rs
@@ -19,7 +19,7 @@ use nom::error::{ErrorKind as NomErrorKind, ParseError};
 /// Internal-only error that should be converted to an [`Error`](super::Error)
 /// before being returned in a public API.
 ///
-/// `InternalError` is an itermediate error that contains a reference to the failed input
+/// `InternalError` is an intermediate error that contains a reference to the failed input
 /// along with the lower-level error `E` that describes the parsing failure. This allows us
 /// to generate the more user friendly [`Error`](super::Error) once parsing completes, while
 /// avoiding the performance losses of generating the user-friendly information


src/parser/macros.rs
```diff
@@ -12,11 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// TODO: Internal variants?
```
notmgsk, line 15: here or a future issue?

notmgsk, line 15:
```suggestion
```

notmgsk:
Before merging we should:
- check it doesn't have any incompatibilities with something that uses this crate

genos:
🔥

src/parser/error/internal.rs
```diff
@@ -12,15 +12,81 @@
...
+    I: ErrorInput,
+
+    E: std::error::Error + Send,
+{
+    fn from_error_kind(input: I, kind: nom::error::ErrorKind) -> Self {
+        Self::new(input, ErrorKind::Internal(GenericParseError(kind)))
+    }
+
+    fn append(input: I, kind: nom::error::ErrorKind, _other: Self) -> Self {
+        // TODO: track previous error?
```
kalzoo, line 66: TODO without linked issue => PR should be draft

src/parser/error/internal.rs
```diff
@@ -12,15 +12,81 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use nom::error::ErrorKind;
+use super::ErrorKind;
+use crate::parser::ErrorInput;
+use nom::error::{ErrorKind as NomErrorKind, ParseError};
+
+/// Internal-only error that should be converted to an [`Error`](super::Error)
+/// before being returned in a public API.
```
kalzoo, line 20: Maybe add a little bit about _why_?

- why there are two error types
- why the conversion should happen when it does

Cargo.toml
```diff
@@ -30,6 +30,10 @@ rstest = "0.15.0"
 [features]
 graphviz-dot = ["dot-writer"]
 
+[[bench]]
+name = "calibration"
```
kalzoo, line 34: This is still benching the `parser`, it's just an additional fixture, can we include it in the same command?

kalzoo (commented):
There and back again, @Shadow53 ! good find and fix. Approved pending fixes for my comments.

kalzoo:
(also, please create an additional issue to drive the time back down with any ideas you have of where to start looking)

Shadow53:
> (also, please create an additional issue to drive the time back down with any ideas you have of where to start looking)

Created #115 to address this.

notmgsk (approved):
My test program still takes ~5s, but that is much less frustrating than ~30s.

Shadow53:
> My test program still takes ~5s, but that is much less frustrating than ~30s.

Leaving a note here for documentation:

The numbers in the OP are using default benchmarking settings, which are equivalent to using release mode. If you tell it to use the `dev` profile instead...

```
$ cargo bench --profile=dev
# ... snip ...
     Running benches/parser.rs (target/debug/deps/parser-d2d0ad29d261c397)
Gnuplot not found, using plotters backend
Benchmarking calibration file/Sample calibration file: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 1173.2s, or reduce sample count to 10.
Benchmarking calibration file/Sample calibration file: Collecting 100 samples in estimated 1173.2 s (100 iterations)
```

src/parser/error/internal.rs
```diff
@@ -12,15 +12,89 @@
...
+///
+/// `InternalError` is an itermediate error that contains a reference to the failed input
```
kalzoo, line 22:
```suggestion
/// `InternalError` is an intermediate error that contains a reference to the failed input
```

kalzoo (approved):
Looks good! Small typo fix, otherwise 🚢

kalzoo:
> > My test program still takes ~5s, but that is much less frustrating than ~30s.
> 
> Leaving a note here for documentation:
> 
> The numbers in the OP are using default benchmarking settings, which are equivalent to using release mode. If you tell it to use the `dev` profile instead...
> 
> ```
> $ cargo bench --profile=dev
> # ... snip ...
>      Running benches/parser.rs (target/debug/deps/parser-d2d0ad29d261c397)
> Gnuplot not found, using plotters backend
> Benchmarking calibration file/Sample calibration file: Warming up for 3.0000 s
> Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 1173.2s, or reduce sample count to 10.
> Benchmarking calibration file/Sample calibration file: Collecting 100 samples in estimated 1173.2 s (100 iterations)
> ```
> 
> 1173.2s / 100 iterations = ~12s to parse the sample calibration file.

Good to note - I had not realized that. This bumps up the priority of the extended fix here (#115).

PR #116, opened 2022-10-26
Fix: do not get line/column info for tokens except on error

Shadow53:
This is another performance improvement. Ish.

When running benchmarks in `dev`/`debug` mode, this showed a 35% decrease in runtime (edit: 26% when I generated HTML reports) on the sample-calibrations file (parse in ~2.33s rather than ~3.6s). It also showed a ~15% improvement on most corpus files, but ~8% regression on three of them.

When running in normal `bench` mode (basically release), the story is a bit different. Before these changes, parsing the sample calibrations file happened in ~223ms on avg, and ~239ms after these changes (~7% regression). Corpus files also showed a single-digit-percent performance regression.

OTOH, the sample calibrations had 15% outliers in the initial run and 5.6% outliers after applying changes. I am not sure whether criterion includes the outliers in average calculations or not.

### DIFF
--- a/src/expression.rs
+++ b/src/expression.rs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 use lexical::{format, to_string_with_options, WriteFloatOptions};
+use nom_locate::LocatedSpan;
 use num_complex::Complex64;
 use std::collections::{hash_map::DefaultHasher, HashMap};
 use std::f64::consts::PI;
@@ -415,7 +416,8 @@ impl FromStr for Expression {
     type Err = ProgramError<Self>;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        let tokens = lex(s)?;
+        let input = LocatedSpan::new(s);
+        let tokens = lex(input)?;
         disallow_leftover(parse_expression(&tokens).map_err(ParseError::from_nom_internal_err))
     }
 }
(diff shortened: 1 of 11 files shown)


Shadow53:
Screen Shot 2022-10-26 at 12 05 08 PM

Screen Shot 2022-10-26 at 12 06 30 PM

Attaching a screenshots of the plotted `criterion` report for reference.

genos:
Re:

> OTOH, the sample calibrations had 15% outliers in the initial run and 5.6% outliers after applying changes. I am not sure whether criterion includes the outliers in average calculations or not.

Per [the book](https://bheisler.github.io/criterion.rs/book/analysis.html), 

> Note, however, that outlier samples are not dropped from the data, and are used in the following analysis steps along with all other samples.

I suppose the reasoning for this is given in the comments [`outliers/mod.rs`](https://github.com/bheisler/criterion.rs/blob/master/src/stats/univariate/outliers/mod.rs), which say 

```rust
//! WARNING: There's no formal/mathematical definition of what an outlier actually is. Therefore,
//! all outlier classifiers are *subjective*, however some classifiers that have become *de facto*
//! standard are provided here.
```

@Shadow53 what was your running environment like at the time of data collection? Was your laptop doing other things, too, or just running benchmarks?

notmgsk:
@Shadow53 can you fix the commits to use the `Fix:` prefix. We haven't switched to angular yet 😅

PR #119, opened 2022-10-28
ci: use Knope and angular commits

rigetti-githubbot:


.github/workflows/release.yml
```diff
@@ -1,41 +1,23 @@
...
 on:
-  push:
-    branches:
-      - main
-      - rc
-
+  release:
+    types: [published]
+  workflow_dispatch:
+    description: "Manually publish release"
```
Shadow53, line 7: Seems like this conflicts with `prepare-release.yml::workflow_dispatch`. I think we're good here with just `on: release: ...`

Shadow53 (changes requested):
I don't think we need to trigger the release workflow manually if we have triggering `knope` manually, which itself triggers a release.

PR #125, opened 2022-11-07
Parse remaining Quil instructions

randall-fulton:
Add support for
- `CONVERT`
- `DEFGATE`
- `INCLUDE`
- `NOP`


Shadow53 (changes requested):
Left one suggestion, looks good to me otherwise, at least as far as Rust is concerned. I'll leave reviewing quil-specific logic to someone else.

src/instruction.rs
```diff
@@ -549,41 +560,44 @@ pub struct JumpUnless {
 
 #[derive(Clone, Debug, PartialEq)]
 pub enum Instruction {
```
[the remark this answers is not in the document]
randall-fulton: Yes. It was difficult to track what existed, so I sorted as well.

src/parser/instruction.rs
```diff
@@ -42,75 +42,70 @@ pub(crate) fn parse_instruction(input: ParserInput) -> InternalParserResult<Inst
...
+            Command::SetScale => command::parse_set_scale(remainder),
+            Command::ShiftFrequency => command::parse_shift_frequency(remainder),
+            Command::ShiftPhase => command::parse_shift_phase(remainder),
+            Command::Store => command::parse_store(remainder),
+            Command::Sub => command::parse_arithmetic(ArithmeticOperator::Subtract, remainder),
+            Command::Xor => command::parse_logical_binary(BinaryOperator::Xor, remainder),
+            other => Err(nom::Err::Failure(InternalParseError::from_kind(
+                &input[..1],
+                ParserErrorKind::UnsupportedInstruction(*other),
+            ))),
```
kalzoo, line 98: This should be unreachable - if there's anything that would be caught, it should be removed from `Command` if we don't plan to support it.

kalzoo (approved):
I agree with @Shadow53 's suggestion and have one change of my own; pending those, lgtm 🚢

src/parser/command.rs
```diff
@@ -154,6 +155,13 @@ pub(crate) fn parse_capture(
     ))
 }
 
+/// Parse the contents of a `CONVERT` instruction.
+pub(crate) fn parse_convert<'a>(input: ParserInput<'a>) -> InternalParserResult<'a, Instruction> {
+    let (input, to) = token!(Identifier(v))(input)?;
```
notmgsk, line 160: These should permit memory references as arguments:

```
❯ echo "DECLARE unadjusted-theta INTEGER[2]
DECLARE theta REAL[2]
MOVE unadjusted-theta[1] 2
CONVERT theta[1] unadjusted-theta[1]
" | qvm
...
Classical memory (low -> high indexes):
    unadjusted-theta: 0 2
    theta: 0.0d0 2.0d0
Amplitudes:
    |0>: 1.0,                                                   P=100.0%
```

src/parser/instruction.rs
```diff
@@ -768,6 +763,25 @@ mod tests {
...
+
+    make_test!(
+        include,
+        parse_instructions,
+        r#"INCLUDE "another/quil/file.quil""#,
+        vec![Instruction::Include(Include {
+            filename: "another/quil/file.quil".to_string()
+        })]
+    );
+
```
notmgsk, line 784: `NOP` test?

src/parser/instruction.rs
randall-fulton: `WAIT` and `OR` have been removed. The former was discussed previously and the latter doesn't appear to even be reserved, according to the Quil spec.

notmgsk:
Noice. 

We should probably have a conversation about how to handle `INCLUDE`d files -- i.e. should we parse them as part of the regular parsing flow?

notmgsk (approved):

PR #126, opened 2022-11-08
feat: Support Expression arithmetic operations

### LINKED ISSUE #124: Support `Expression` arithmetic operations
Would be useful if we could do e.g. `x + y` where `x` and `y` are `Expression`.

genos:
Closes #124

### DIFF
--- a/src/expression.rs
+++ b/src/expression.rs
@@ -20,6 +20,7 @@ use std::f64::consts::PI;
 use std::fmt;
 use std::hash::{Hash, Hasher};
 use std::num::NonZeroI32;
+use std::ops::{Add, Div, Mul, Sub};
 use std::str::FromStr;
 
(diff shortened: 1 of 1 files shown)


src/expression.rs
```diff
@@ -151,6 +154,33 @@ impl PartialEq for Expression {
...
+                Expression::Infix {
+                    left: Box::new(self),
+                    operator: InfixOperator::$operator,
+                    right: Box::new(other),
+                }
+            }
+        }
+        impl $name_assign for Expression {
+            fn $function_assign(&mut self, other: Self) {
+                let result = self.clone().$function(other);
```
genos, line 171: Is there some clever `std::mem` maneuver to sidestep the need for `clone()` here?

notmgsk (approved):
👌🏽

PR #128, opened 2022-11-15
Program.into_simplified

kalzoo:
Simple approach:

- Expand calibrations
- Remove calibrations
- Scan through instructions and keep only the frames and waveforms used by at least one instruction

Did not address memory references given that a full simplification of those would reduce shrinking them to the minimum size (removing unnecessary capacity), and there's some complexity there. On top of that, too many memory regions is not as common of a problem as is thousands of lines of unused calibrations.

### DIFF
--- a/src/program/frame.rs
+++ b/src/program/frame.rs
@@ -87,6 +87,19 @@ impl FrameSet {
         self.frames.insert(identifier, attributes);
     }
 
+    /// Return a new [FrameSet] which describes only the given [FrameIdentifier]s.
+    pub fn intersection(&self, identifiers: &HashSet<&FrameIdentifier>) -> Self {
+        let mut new_frameset = Self::new();
+
+        for (identifier, definition) in &self.frames {
+            if identifiers.contains(&identifier) {
+                new_frameset.insert(identifier.clone(), definition.clone())
+            }
+        }
+
+        new_frameset
+    }
+
     /// Iterate through the contained frames.
     pub fn iter(&self) -> std::collections::hash_map::Iter<'_, FrameIdentifier, FrameAttributes> {
         self.frames.iter()
--- a/src/program/mod.rs
+++ b/src/program/mod.rs
@@ -519,6 +519,12 @@ DEFCAL MEASURE 0 addr:
 DEFCAL MEASURE 1 addr:
     CAPTURE 1 \"ro_rx\" custom addr
 
+DEFFRAME 0 \"ro_rx\":
+    ATTRIBUTE: \"value\"
+
+DEFFRAME 1 \"ro_rx\":
+    ATTRIBUTE: \"other\"
+
 DEFWAVEFORM custom:
     0.0, 1.0
 
@@ -531,6 +537,9 @@ MEASURE 0 ro
 
         let expected = "
 DECLARE ro BIT
+
+DEFFRAME 0 \"ro_rx\":
+    ATTRIBUTE: \"value\"
 
 DEFWAVEFORM custom:
     0.0, 1.0
--- a/src/program/mod.rs
+++ b/src/program/mod.rs
@@ -146,10 +146,8 @@ impl Program {
         instruction: &'a Instruction,
         include_blocked: bool,
     ) -> Option<HashSet<&'a FrameIdentifier>> {
-        let qubits_used_in_program = self.get_used_qubits();
-
         instruction
-            .get_frame_match_condition(include_blocked, qubits_used_in_program)
+            .get_frame_match_condition(include_blocked)
             .map(|condition| self.frames.get_matching_keys(condition))
     }
 
--- a/src/program/mod.rs
+++ b/src/program/mod.rs
@@ -172,7 +172,7 @@ impl Program {
             .collect::<HashSet<_>>()
     }
 
-    /// Simplify this program into a new [Program] which contains only instructions
+    /// Simplify this program into a new [`Program`] which contains only instructions
     /// and definitions which are executed; effectively, perform dead code removal.
     ///
     /// Removes:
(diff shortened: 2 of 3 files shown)


src/program/frame.rs
```diff
@@ -87,6 +87,19 @@ impl FrameSet {
...
+    pub fn intersection(&self, identifiers: &HashSet<&FrameIdentifier>) -> Self {
+        let mut new_frameset = Self::new();
+
+        for (identifier, definition) in &self.frames {
+            if identifiers.contains(&identifier) {
+                new_frameset.insert(identifier.clone(), definition.clone())
+            }
+        }
+
+        new_frameset
```
genos, line 100: Perfectly straightforward, but I think it might be more idiomatic as
```rust
FrameSet {
    frames: self
        .frames
        .iter()
        .filter_map(|(id, attr)| {
            identifiers
                .contains(id)
                .then_some((id.clone(), attr.clone()))
        })
        .collect(),
}
```

genos: Could arguably do something shorter with `self.frames.retain(…)`, but that might need cloning the entire thing first.

src/program/mod.rs
```diff
@@ -155,11 +164,48 @@ impl Program {
...
+            if let Some(frames) = expanded_program.get_frames_for_instruction(instruction, false) {
+                frames_used.extend(frames.into_iter())
+            }
+
+            if let Some(waveform) = instruction.get_waveform_invocation() {
+                waveforms_used.insert(&waveform.name);
+            }
+        }
+
+        expanded_program.frames = self.frames.intersection(&frames_used);
```
genos, line 201: Why not use `expanded_program.frames.frameset.retain` in the same way you have below for `waveforms`?

```rust
expanded_program.frames.frameset.retain(|n, _| frames_used.contains(n));
```

genos (approved):
A couple of questions, but mostly for my own benefit. 👍 either way!

src/program/mod.rs
kalzoo: Because the waveforms are just directly a `BTreeMap` but the `FrameSet` keeps its inner frames private

src/program/frame.rs
Shadow53: To improve on @genos's suggestion, there's also:

```rust
FrameSet {
    frames: self
        .frames
        .iter()
        .filter(|(id, _)| identifiers.contains(id))
        .cloned()
        .collect(),
}
```

src/program/mod.rs
```diff
@@ -155,11 +164,48 @@ impl Program {
...
                 Instruction::Fence(fence) => fence.qubits.clone(),
+                Instruction::Capture(capture) => capture.frame.qubits.clone(),
+                Instruction::Pulse(pulse) => pulse.frame.qubits.clone(),
+                Instruction::RawCapture(raw_capture) => raw_capture.frame.qubits.clone(),
                 _ => vec![],
             })
             .collect::<HashSet<_>>()
     }
 
+    /// Simplify this program into a new [Program] which contains only instructions
```
Shadow53, line 175:
```suggestion
    /// Simplify this program into a new [`Program`] which contains only instructions
```

src/program/mod.rs
```diff
@@ -155,11 +164,48 @@ impl Program {
...
+    ///
+    /// Removes:
+    /// - All calibrations, following calibration expansion
+    /// - Frame definitions which are not used by any instruction such as `PULSE` or `CAPTURE`
+    /// - Waveform definitions which are not used by any instruction
+    ///
+    /// When a valid program is simplified, it remains valid.
+    pub fn into_simplified(&self) -> Result<Self> {
+        let mut expanded_program = self.expand_calibrations()?;
+        expanded_program.calibrations = CalibrationSet::default();
```
Shadow53, line 186: A comment on why you replace the calibrations would be nice. As someone new to this part of the code, it looks like you are setting the calibrations, only to reset them.

src/program/mod.rs
```diff
@@ -155,11 +164,48 @@ impl Program {
...
+
+        let mut frames_used: HashSet<&FrameIdentifier> = HashSet::new();
+        let mut waveforms_used: HashSet<&String> = HashSet::new();
+
+        for instruction in &expanded_program.instructions {
+            if let Some(frames) = expanded_program.get_frames_for_instruction(instruction, false) {
+                frames_used.extend(frames.into_iter())
```
Shadow53, line 193: This should be sufficient, as `extend` takes an `impl IntoIterator`.

```suggestion
                frames_used.extend(frames)
```

Shadow53 (commented):
Left a few small style suggestions, but code looks good overall

src/program/mod.rs
```diff
@@ -155,11 +164,48 @@ impl Program {
...
+    ///
+    /// Removes:
+    /// - All calibrations, following calibration expansion
+    /// - Frame definitions which are not used by any instruction such as `PULSE` or `CAPTURE`
+    /// - Waveform definitions which are not used by any instruction
+    ///
+    /// When a valid program is simplified, it remains valid.
+    pub fn into_simplified(&self) -> Result<Self> {
+        let mut expanded_program = self.expand_calibrations()?;
+        expanded_program.calibrations = CalibrationSet::default();
```
notmgsk, line 186:
```suggestion
        // Remove calibrations such that the resulting program contains
        // only instructions. Calibrations have already been expanded, so
        // technically there is no need to keep them around anyway.
        expanded_program.calibrations = CalibrationSet::default();
```

src/program/mod.rs
notmgsk: @Shadow53 It doesn't _set_ calibrations -- it _expands_ them, and then removes the calibration definition. Think of it as inline a function body: the "calibration" is the function, and instructions that reference the calibration are replaced with the calibration body.

notmgsk (approved):

PR #129, opened 2022-11-15
Frame calculations for RESET

### LINKED ISSUE #33: Support RESET Instructions
`RESET` instructions are properly parsed, but their "used frames" are computed as `None`. This is incorrect - a `RESET` acts on the following frames:

1. If a qubit is specified, it acts on frames which use exactly that qubit, and blocks all frames which intersect that qubit
2. If no qubit is specified, apply rule # 1 to all qubits used in the program.

This is an over-simplification in that the implementation of `RESET` is backend-dependent and thus may act on a smaller number of frames than described here. However, this model will allow the timing of a user's program to be properly modeled on the client side.

kalzoo:
Simple approach to RESET frame calculation support as described in the issue.

Reviewers:

- Does this approach make sense?
- Are there cases in which "overly greedy" blocking of frames as computed here in `quil-rs` will cause issues in compilation or elsewhere downstream?

Closes #33 

Linked to #128 for use in tandem downstream.

### DIFF
--- a/src/instruction.rs
+++ b/src/instruction.rs
@@ -14,6 +14,8 @@
 
 use nom_locate::LocatedSpan;
 use serde::{Deserialize, Serialize};
+use std::borrow::Cow;
+use std::collections::HashSet;
 use std::str::FromStr;
 use std::{collections::HashMap, fmt};
 
@@ -1158,6 +1160,7 @@ impl Instruction {
     pub(crate) fn get_frame_match_condition(
         &self,
         include_blocked: bool,
+        qubits_available: HashSet<Qubit>,
     ) -> Option<FrameMatchCondition> {
         match self {
             Instruction::Pulse(Pulse {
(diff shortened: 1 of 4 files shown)


src/instruction.rs
```diff
@@ -1178,18 +1181,35 @@ impl Instruction {
...
             } else {
-                FrameMatchCondition::AnyOfQubits(qubits)
+                FrameMatchCondition::AnyOfQubits(Cow::Borrowed(qubits))
             }),
+            Instruction::Reset(Reset { qubit }) => {
+                let qubits = match qubit {
+                    Some(qubit) => {
+                        let mut set = HashSet::new();
+                        set.insert(qubit.clone());
+                        set
```
genos, line 1201: `Hashset::from([qubit.clone()])` or `[qubit.clone()].into()`.

src/program/frame.rs
```diff
@@ -12,7 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::collections::{HashMap, HashSet};
+use std::{
+    borrow::Cow,
```
genos, line 16: Why is the 🐮 stuff necessary?

genos (commented):
One comment, one question

notmgsk (approved):

PR #131, opened 2022-11-15
fix: make dynamic error `Sync` as well

### LINKED ISSUE #122: Add `Sync` to previous error field in parser error
`quil_rs::parser::error::error::Error` has the field:

```rs
 previous: Option<Box<dyn std::error::Error + 'static + Send>>
```

We should also add `Sync` so that downstream code can reliably used in asynchronous contexts

```rs
 previous: Option<Box<dyn std::error::Error + 'static + Send + Sync>>
```

Shadow53:
Resolves #122.

### DIFF
--- a/src/parser/error/error.rs
+++ b/src/parser/error/error.rs
@@ -43,13 +43,13 @@ where
     column: usize,
     snippet: String,
     kind: ErrorKind<E>,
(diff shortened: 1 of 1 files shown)


MarquessV (approved):
Send (or Sync?) it!

PR #135, opened 2022-11-17
ci: fix release workflow

### LINKED ISSUE #123: New version release is broken :(
The latest `quil-rs` version on [crates.io is v0.14.1](https://crates.io/crates/quil-rs). The version specified in `Cargo.toml` is `0.10.0`, which happened in a recent merge which _should have_ bumped the version to 0.15.0. 

Not really sure how it got into this pickle, but it's something we should fix before it gets us into trouble.

Changing `Cargo.toml`'s version to `0.14.1` and running `knope --dry-run prerelease` shows the version would be bumped to `0.11.0`, so that's not the issue...

(Paging the wise and wonderful @dbanty just in case they can spare a few minutes for their old friends.)

Shadow53:
Resolves #123.

- Updates the `knope` version
- Per @dbanty, `--prerelease-label` calls a workflow but as a prerelease. So to deduplicate the `knope.toml` (and avoid issues like with missing the `Release` step), the file now contains only a `release` workflow, and the CI calls it with `--prerelease-label` for prereleases.
  - One potential argument against this is that the prerelease commit message now says "release" instead of "prerelease". Since it includes the release tag, including a `rc.x` part, I think that conveys "prerelease" appropriately.
  - That said, I am open to arguments for restoring a "prerelease" workflow.


notmgsk:
We're probably going to have a problem when we get to 0.15.0, since there is already a 0.15.0-rc.0 [on crates.io](https://crates.io/crates/quil-rs/versions). Do you think we should just tag 0.15.1 and go from there?

Shadow53:
> We're probably going to have a problem when we get to 0.15.0, since there is already a 0.15.0-rc.0 [on crates.io](https://crates.io/crates/quil-rs/versions). Do you think we should just tag 0.15.1 and go from there?

I am hoping that, once we reach that point, `knope` will detect the prerelease and automatically skip it.

If I read [this function](https://github.com/knope-dev/knope/blob/a46fe2b5b8c6e6a8f378afcd7449148d30f81e48/src/releases/semver.rs#L480) correctly, it will ignore the latest prerelease if the major/minor/patch versions do not match the current stable.

... Which made me wonder if we'll run into an issue where we keep hitting `0.14.3-rc.0`. It looks like yes, so we might need a new PR to knope :/

kalzoo (approved):

Shadow53:
Updated the `knope` version to 0.7.0 and fixed some `clippy` lints so that CI passes.

BatmanAoD:
@Shadow53 Is this PR still relevant?

Shadow53:
This PR is so old, better to start over

PR #138, opened 2022-12-06
fix: Waveforms w/o params need no parens

### LINKED ISSUE #137: Waveforms without parameters are serialized with unnecessary parentheses
Waveforms that aren't parameterized should not include be written with parentheses.

Example of what we see right now:
```
	NONBLOCKING PULSE 0 1 "cz" q0_q1_cz/CZ()
```

What we should see:
```
	NONBLOCKING PULSE 0 1 "cz" q0_q1_cz/CZ
```

randall-fulton:
Closes #137

### DIFF
--- a/src/instruction.rs
+++ b/src/instruction.rs
@@ -289,16 +289,36 @@ impl fmt::Display for WaveformInvocation {
 
         key_value_pairs.sort_by(|(k1, _), (k2, _)| k1.cmp(k2));
 
-        write!(
-            f,
-            "{}({})",
-            self.name,
-            key_value_pairs
-                .iter()
-                .map(|(k, v)| format!("{}: {}", k, v))
-                .collect::<Vec<String>>()
-                .join(", ")
-        )
+        if key_value_pairs.is_empty() {
+            write!(f, "{}", self.name,)
+        } else {
+            write!(
+                f,
+                "{}({})",
+                self.name,
+                key_value_pairs
+                    .iter()
+                    .map(|(k, v)| format!("{}: {}", k, v))
+                    .collect::<Vec<String>>()
+                    .join(", ")
+            )
+        }
+    }
+}
+
+#[cfg(test)]
+mod waveform_invocation_tests {
+    use std::collections::HashMap;
+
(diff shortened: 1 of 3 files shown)


src/instruction.rs
```diff
@@ -289,16 +289,36 @@ impl fmt::Display for WaveformInvocation {
...
+                self.name,
+                key_value_pairs
+                    .iter()
+                    .map(|(k, v)| format!("{}: {}", k, v))
+                    .collect::<Vec<String>>()
+                    .join(", ")
+            )
+        }
+    }
+}
```
Shadow53, line 307: Suggesting this because, while it doesn't necessarily look as nice, it should avoid (some of) the unnecessary allocations from building a `Vec<String>` for printing purposes.

```suggestion
        write!(f, "{}", self.name)?;
        if !key_value_pairs.is_empty() {
            write!(f, "(")?;
            for (i, (k, v)) in key_value_pairs.iter().enumerate() {
                if i > 0 { write!(f, ", ")? }
                write!(f, "{}: {}", k, v)?;
            }
            write!(f, ")")?;
        }
        Ok(())
    }
}
```

Shadow53: You may also want to run `cargo fmt` and `cargo check` on this locally, just to make sure I didn't typo anything -- already caught one while writing this comment.

Shadow53 (commented):
Left a suggestion that avoids allocating new strings when there are key-value pairs.

src/instruction.rs
kalzoo: I agree that this is a good change - note though that we Display by means of `collect::<Vec<_>>().join(", ")` in many places in this library. Rather than a suggestion on this MR, it should be a fix in all the places to avoid confusing future developers ("why is this optimized in only one of a dozen places? is this more important than the others?")

kalzoo (approved):

src/instruction.rs
randall-fulton: I created #139 to capture @Shadow53's suggestion. Feel free to add more detail there.

PR #140, opened 2023-01-06
Introducing: quil-py

MarquessV:
This PR turns this repo into a workspace and adds `quil-py` as a new crate and Python package.

In order to support the Python bindings, some breaking changes were necessary. Primarily, any enum struct variant had to be broken out into its own type.

In other cases, code in `quil-rs` was refactored in a non-breaking way to support the Python bindings, such as by lifting the implementations for `fmt::Display` up to the types used for `Instruction` variants. Some bugs were revealed and squashed in the process as well.

Some features like `Program` addition have been added as well. While motivated by the goal to use this as a backend for `PyQuil`s `Program`, functionality was only ported back from `PyQuil` if it makes sense for `quil-rs` to have. For example, `Program` addition stands on its own as a feature this library have. `num_shots` however is not a `quil` concept, and wasn't ported back. This is important to keep in mind during review. I've tried my best to keep the  responsibilities of `quil-rs` and `PyQuil` separate, but the lines are sometimes blurry.

This PR is targeting a feature branch because it is not a complete implementation of `quil-py`. This PR, along with the [pyQuil counterpart](https://github.com/rigetti/pyquil/pull/1518) should form the foundation for all the future work that needs to be done. Unlike the pyQuil PR though, most of the crate is still functional with these changes.

All feedback is welcome, I've also included discussion prompts and clarifications as comments on the PR.


quil-py/src/instruction/gate.rs
```diff
@@ -0,0 +1,110 @@
...
+                Vec::<Qubit>::py_try_from(py, &qubits)?,
+                Vec::<GateModifier>::py_try_from(py, &modifiers)?,
+            )
+            .map_err(GateError::from)
+            .map_err(GateError::to_py_err)?,
+        ))
+    }
+
+    fn dagger(&self, py: Python<'_>) -> PyResult<Self> {
+        self.as_inner().clone().dagger().to_python(py)
```
MarquessV, line 56: Unlike `pyQuil`, we return a new Gate when using these methods.

quil-py/src/instruction/mod.rs
```diff
@@ -0,0 +1,77 @@
...
+    expression::{
+        PyExpression, PyExpressionFunction, PyFunctionCallExpression, PyInfixExpression,
+        PyInfixOperator,
+    },
+    frame::{PyAttributeValue, PyFrameAttributes, PyFrameDefinition, PyFrameIdentifier},
+    gate::{PyGate, PyGateDefinition, PyGateError, PyGateModifier, PyGateSpecification},
+    memory_region::{PyMemoryReference, PyMemoryRegion},
+    qubit::PyQubit,
+    waveform::{PyWaveform, PyWaveformDefinition},
+};
```
MarquessV, line 20: I found it difficult to work in one large `instructions.rs` file, so I started breaking them down into groups. Categorizing can sometimes be arbitrary so these groupings are transparent to users of the crate.

quil-py/src/lib.rs
```diff
@@ -0,0 +1,20 @@
...
+#[pymodule]
+fn quil(py: Python<'_>, m: &PyModule) -> PyResult<()> {
+    init_submodule("quil", py, m)?;
+    Ok(())
+}
+
+pub fn init_quil_submodule(name: &str, py: Python<'_>, m: &PyModule) -> PyResult<()> {
+    init_submodule(name, py, m)?;
+    Ok(())
+}
```
MarquessV, line 20: This is what allows us to re-export these binding through `qcs-sdk-python`. One unanswered question, how will type hints work with this approach?

quil-rs/src/instruction/gate.rs
```diff
@@ -0,0 +1,222 @@
+use std::fmt;
```
MarquessV, line 1: Like in `quil-py`, I started breaking out instructions into multiple files. I started with Gate and it's counterparts, since I chose it as the pilot instruction for backing a `pyQuil` `AbstractInstruction`. As we implement more instruction constructors, I plan to continue breaking these out, such that the file structure of `quil-rs` and `quil-py` mirror each other.

quil-rs/src/instruction/gate.rs
```diff
@@ -0,0 +1,222 @@
...
+
+use super::{format_qubits, get_expression_parameter_string, Qubit};
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct Gate {
+    pub name: String,
+    pub parameters: Vec<Expression>,
+    pub qubits: Vec<Qubit>,
+    pub modifiers: Vec<GateModifier>,
+}
```
MarquessV, line 16: All of the instruction structs have public members. However, as we'll see in the `impl Gate` block below, it requires some knowledge to update them in a correct way. Should we strip away visibility into these members?

quil-rs/src/instruction/gate.rs
```diff
@@ -0,0 +1,222 @@
...
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum GateModifier {
+    Controlled,
+    Dagger,
+    Forked,
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum GateError {
```
MarquessV, line 26: Is this an appropriate pattern for Rust Errors?

quil-rs/src/program/mod.rs
```diff
@@ -217,36 +238,64 @@ impl Program {
         Ok(expanded_program)
     }
 
+    pub fn is_protoquil(&self) -> Result<()> {
```
MarquessV, line 241: Is a `Result<()>` an appropriate return type for this kind of function? The `Result` allows us to attach valuable context to the error but makes a `bool` success variant redundant since it would never be `false`.

quil-rs/src/program/mod.rs
```diff
@@ -217,36 +238,64 @@ impl Program {
...
-            result.extend(self.calibrations.to_instructions());
+            result.extend(self.to_headers())
         }
 
         result.extend(self.instructions.clone());
 
         result
     }
 
+    pub fn to_headers(&self) -> Vec<Instruction> {
```
MarquessV, line 275: are `headers` and `instructions` good buckets to separate things by?

quil-rs/src/validation/identifier.rs
```diff
@@ -0,0 +1,34 @@
+//! Types and functions related to validating Quil identifiers
```
MarquessV, line 1: In order to have a "source of truth" for validation rules across the parser and constructors for different instruction types, I opted to create a validation module.

quil-rs/src/validation/identifier.rs
```diff
@@ -0,0 +1,34 @@
...
+use std::str::FromStr;
+
+use fancy_regex::Regex;
```
MarquessV, line 4: `fancy_regex` supports negative look behind, which we use as part of `IDENTIFIER_REGEX` to ensure that an identifier doesn't end with a `-`, even though the middle characters of the identifier can contain it.

quil-rs/src/program/calibration.rs
```diff
@@ -316,6 +316,12 @@ impl CalibrationSet {
         self.measure_calibrations.push(calibration)
     }
 
+    /// Append another [CalibrationSet] onto this one
```
kalzoo, line 319:
```suggestion
    /// Append another [`CalibrationSet`] onto this one
```

quil-rs/src/program/mod.rs
```diff
@@ -100,6 +101,26 @@ impl Program {
...
     }
 
+    /// Creates a new conjugate transpose of the [`Program`] by reversing the order of gate
+    /// instructions and applying the DAGGER modifier to each.
+    ///
+    ///
+    /// # Errors
+    ///
+    /// Errors if any of the instructions in the program are not [`Instruction::Gate`]
+    pub fn dagger(&self) -> Result<Self> {
```
kalzoo, line 111: 👍

quil-rs/src/program/mod.rs
kalzoo: Yes, this is how I'd do it.

(that being said, see my other note about whether we should have this function at all)

quil-rs/src/program/mod.rs
kalzoo: nope, and I regret that decision from long ago! Even the "headers" (which is a term I invented here) are themselves instructions. I think the ability to simplify a program removes the need to separate "headers" from the other instructions, which I likely added to just trim the tens of thousands of lines of calibrations off of a program.

tl;dr I think `to_instructions()` should take no arguments and return all of the "headers" too.

quil-rs/src/validation/identifier.rs
kalzoo: Is lookbehind needed to enforce this?

quil-rs/src/validation/identifier.rs
```diff
@@ -0,0 +1,34 @@
...
+pub enum IdentifierValidationError {
+    #[error("{0} is a reserved token")]
+    Reserved(ReservedToken),
+
+    #[error("{0} is not a valid identifier")]
+    Invalid(String),
+}
+
+/// A regex that matches only valid Quil identifiers
+const IDENTIFIER_REGEX: &str = r"^[A-Za-z_]+[A-Za-z0-9\-_]*.*(?<!\-)$";
```
kalzoo, line 19: The `.*` seems incorrect to me. Why not just use the pattern directly from the spec? `[A-Za-z_][A-Za-z0-9\-_]*[A-Za-z0-9_]`

MarquessV: I started with that initially, but it doesn't work for 1 character strings like `H`. Attempting to patch that causes other problems with the regex.

The `quil` spec actually defines _two_ regexes for an identifier which didn't strike me at the time as something that could be composed into a single regex, but of course it can, so I stitched them together and switched back to the `regex` crate.

I also pushed up some tests for validation.

quil-rs/src/instruction/gate.rs
```diff
@@ -0,0 +1,222 @@
...
+                "({})",
+                self.parameters
+                    .iter()
+                    .map(|p| p.to_string())
+                    .collect::<String>()
+            ),
+        };
+        writeln!(
+            f,
+            "DEFGATE {}{} AS {}:",
```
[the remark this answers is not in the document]
MarquessV: Sounds good to me. My only counter-argument would be that `quil-rs` is hard coding what the default behavior should be instead of faithfully mirroring the specification. Ie. should `quil-rs` really be defining the default? Or should it omit and let a compiler or whatever else handle that? It's so unlikely to be relevant that it's not worth spending much time on and I think it's fine for `quil-rs` to have some opinions on the code it parses anyways.

quil-rs/src/instruction/mod.rs
```diff
@@ -1284,6 +1191,45 @@ impl Instruction {
...
+            self,
+            Instruction::Declaration(_)
+                | Instruction::Gate(_)
+                | Instruction::Pragma(_)
+                | Instruction::Reset(_)
+                | Instruction::Measurement(_)
+        )
+    }
+
+    pub fn is_valid_quilt(&self) -> bool {
```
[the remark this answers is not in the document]
MarquessV: Yeah, sounds good to me. This is a backport from `pyQuil` that I thought might be useful. I see what you mean about them overcommitting and underdelivering. I'll deprecate or otherwise find a way to deal with it on the pyQuil side.

quil-py/src/program/calibration_set.rs
```diff
@@ -0,0 +1,32 @@
...
+    PyCalibrationSet(CalibrationSet) as "CalibrationSet" {
+        calibrations: Vec<Calibration> => Vec<PyCalibration>,
+        measure_calibrations: Vec<MeasureCalibrationDefinition> => Vec<PyMeasureCalibrationDefinition>
+    }
+}
+
+#[pymethods]
+impl PyCalibrationSet {
+    #[new]
+    pub fn new(
```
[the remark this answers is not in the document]
MarquessV: Definitely, one of the follow-up issues to this is to flesh out this API

quil-rs/src/instruction/gate.rs
MarquessV: Added a CONTRIBUTING.md

quil-rs/src/instruction/mod.rs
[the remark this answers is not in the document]
MarquessV: These have been removed.

quil-rs/src/instruction/gate.rs
MarquessV: Per offline discussion, we're letting this slide so we can press forward instead of adding more boiler plate.

quil-rs/src/validation/identifier.rs
```diff
@@ -0,0 +1,64 @@
...
+    let re = Regex::new(IDENTIFIER_REGEX).expect("regex should be valid");
+
+    match re.is_match(ident) {
+        true => Ok(true),
+        false => Err(IdentifierValidationError::Invalid(ident.to_string())),
+    }
+}
+
+/// Returns an error if the given identifier is reserved, or if it is not a valid Quil identifier
+pub fn validate_user_identifier(ident: &str) -> Result<bool, IdentifierValidationError> {
```
Shadow53, line 32: Same here with `Result<(), _>`

Shadow53 (changes requested):
Left some suggestions. A bunch are rewrites to `fmt::Display` implementations to avoid allocating unnecessary `String`s or `Vec`s. I realized later that they were likely moved from previous code as part of creating structs for enum variants, so while I'd still like to see the changes, I'm also okay with creating a tech debt issue for later.

CONTRIBUTING.md
```diff
@@ -0,0 +1,34 @@
...
+
+### Instruction Module
+
+The `Instruction` module is large, so we follow a few guidelines when adding or changing instructions to help keep
+them organized and consistent in both the `quil-rs` and `quil-py` crates.
+
+All instruction definitions live inside of the `Instruction` module. For ease of development,
+instruction types are grouped into submodules so that similar instructions can be modified and viewed together
+rather than having all instructions in one large file. However, to avoid imposing potentially arbritrary
+instruction categories onto the user, the `Instruction` module publically re-exports all of its types from
```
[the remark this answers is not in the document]
MarquessV: Yeah, good point, fixed

quil-py/src/instruction/arithmetic.rs
```diff
@@ -0,0 +1,76 @@
...
+use super::PyMemoryReference;
+
+py_wrap_union_enum! {
+    PyArithmeticOperator(ArithmeticOperator) as "ArithmeticOperator" {
+        add: Add,
+        subtract: Subtract,
+        divide: Divide,
+        multiply: Multiply
+    }
+}
```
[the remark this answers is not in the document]
MarquessV: My main issue with using a mix of `py_wrap_simple_enum` and `py_wrap_union_enum` is the inconsistent API it exposes. For example, in order to initialize and check equality of an enum variant with `py_wrap_simple_enum` you do something like:

```py
a = MyEnum.Field
assert a == MyEnum.Field
```

Whereas with a `py_wrap_union_enum`, you would do something like:

```py
a = MyEnum.new_a()
assert a.is_a()
```

As a user of the library, I would expect that enums are consistent in their API _or_ there is a good reason as to why they are exposed differently. Because one is "all unit variants" doesn't feel like a justified reason to me, so I opted to make them all with `py_wrap_union_enum!` so they have the same API.

quil-rs/src/validation/identifier.rs
```diff
@@ -0,0 +1,64 @@
...
+    #[error("{0} is not a valid identifier")]
+    Invalid(String),
+}
+
+/// A regex that matches only valid Quil identifiers
+const IDENTIFIER_REGEX: &str = r"^([A-Za-z_]|[A-Za-z_][A-Za-z0-9\-_]*[A-Za-z0-9_])$";
+
+/// Returns an error if the given identifier is not a valid Quil Identifier
+pub fn validate_identifier(ident: &str) -> Result<bool, IdentifierValidationError> {
+    let re = Regex::new(IDENTIFIER_REGEX).expect("regex should be valid");
```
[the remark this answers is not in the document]
MarquessV: Makes sense, refactored with `once_cell`

quil-rs/src/reserved.rs
```diff
@@ -0,0 +1,130 @@
...
+/// An enum that can represent any reserved token in quil.
+#[derive(Debug, PartialEq)]
+pub enum ReservedToken {
+    Keyword(ReservedKeyword),
+    Gate(ReservedGate),
+    Constant(ReservedConstant),
+}
+
+impl FromStr for ReservedToken {
+    type Err = String;
```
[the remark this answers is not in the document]
MarquessV: Done

quil-py/src/instruction/expression.rs
```diff
@@ -0,0 +1,94 @@
...
+
+#[pymethods]
+impl PyExpression {
+    #[staticmethod]
+    fn parse_from_str(py: Python<'_>, expression: String) -> PyResult<Self> {
+        Expression::from_str(&expression)
+            .map_err(|e| PyValueError::new_err(e.to_string()))?
+            .to_python(py)
+    }
+}
```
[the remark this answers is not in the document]
MarquessV: `impl_parse` needs `impl_from_str` and `impl_from_str` needs an error type has `From<E>` for an error type `E`. The error type is `ProgramError<Expression>` which isn't something that is easily wrapped via the macros. That said, I can rewrite this to better match the API of those macros as a middle ground. I may refactor `ProgramError`, but I'm biasing a little bit for speed atm, so I may decide to create an issue and press forward.

quil-py/src/instruction/qubit.rs
```diff
@@ -0,0 +1,33 @@
...
+        variable: Variable => Py<PyString>
+    }
+}
+impl_repr!(PyQubit);
+impl_str!(PyQubit);
+impl_hash!(PyQubit);
+
+#[pymethods]
+impl PyQubit {
+    fn __richcmp__(&self, py: Python<'_>, other: &Self, op: CompareOp) -> PyObject {
```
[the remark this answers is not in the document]
MarquessV: No, there isn't a clear definition of what it means for a qubit to be more or less than another. They can be an integer, but they can also be strings or a placeholder (soon). Returning `NotImplemented` for undefined cases is [what pyo3 recommends](https://pyo3.rs/v0.18.0/class/object.html?highlight=__richcmp__#comparisons).

quil-py/src/lib.rs
MarquessV: Yeah I think a combination of workspace dependencies and specifying a specific version of `quil` as a dependency of the `qcs-sdk-python` package will allow us to enforce consistent `quil` versions.

quil-py/src/program/mod.rs
```diff
@@ -0,0 +1,197 @@
...
+    pub fn new() -> Self {
+        Self(Program::default())
+    }
+
+    #[staticmethod]
+    pub fn from_string(program: &str) -> PyResult<Self> {
+        Ok(Self(
+            Program::from_str(program).map_err(|e| ParseError::new_err(e.to_string()))?,
+        ))
+    }
```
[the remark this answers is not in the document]
MarquessV: Again, blocked by the error type Program uses.

quil-rs/src/instruction/gate.rs
```diff
@@ -0,0 +1,222 @@
...
+            qubits,
+            modifiers,
+        })
+    }
+
+    /// Apply a DAGGER modifier to the gate
+    pub fn dagger(mut self) -> Self {
+        self.modifiers.insert(0, GateModifier::Dagger);
+        self
+    }
```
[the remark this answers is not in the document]
MarquessV: To return a new gate instead of modifying the existing one.

quil-rs/src/parser/gate.rs
```diff
@@ -35,15 +36,16 @@ pub(crate) fn parse_gate<'a>(input: ParserInput<'a>) -> InternalParserResult<'a,
...
-            qubits,
-            modifiers,
-        }),
-    ))
+    let gate = Gate::new(name.as_str(), parameters, qubits, modifiers).map_err(|e| {
+        nom::Err::Error(InternalError::from_kind(
+            input,
+            ParserErrorKind::InvalidInstruction {
+                instruction_type: "Gate".to_string(),
+                reason: e.to_string(),
```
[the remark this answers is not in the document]
MarquessV: This ultimately gets bubbled up to the user, and I thought boxed error types were generally frowned upon as part of a public API. What do we gain by defining it as a boxed error?

quil-rs/src/parser/gate.rs
```diff
@@ -35,15 +36,16 @@ pub(crate) fn parse_gate<'a>(input: ParserInput<'a>) -> InternalParserResult<'a,
...
-            parameters,
-            qubits,
-            modifiers,
-        }),
-    ))
+    let gate = Gate::new(name.as_str(), parameters, qubits, modifiers).map_err(|e| {
+        nom::Err::Error(InternalError::from_kind(
+            input,
+            ParserErrorKind::InvalidInstruction {
+                instruction_type: "Gate".to_string(),
```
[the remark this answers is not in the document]
MarquessV: Not saying it couldn't happen, but I don't see another place this would be needed. These will be isolated to the point of creation of an instruction type from the parser.

quil-rs/src/program/error/mod.rs
```diff
@@ -26,6 +26,8 @@ pub use leftover::LeftoverError;
 pub use result::{disallow_leftover, map_parsed, recover};
 pub use syntax::SyntaxError;
 
+// TODO: Should this be refactored as a enum for any error that can occur while working with a
+// `Program`?
```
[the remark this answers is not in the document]
MarquessV: Yeah, that is what the existing doc comment suggested.

Unfortunately this error turned out to be more tightly wound in a lot of existing code than I expected. Refactoring it would be a heavier lift than I want to shoehorn into this already hefty PR. Unfortunately, The generic parameter makes it incompatible with the pyo3 macros as they exist today so it does get in the way a bit here.

quil-py/src/instruction/arithmetic.rs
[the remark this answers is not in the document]
jselig-rigetti: Hmm... my expectation in a python environment would be that all-unit-variant enums are the only "real" enums and should have class member constants. I would think that anything which requires creating an instance to only be "enum-like", despite the rust representation

quil-rs/src/instruction/gate.rs
```diff
@@ -0,0 +1,222 @@
...
+            f,
+            "DEFGATE {}{} AS {}:",
+            self.name,
+            parameter_str,
+            match self.specification {
+                GateSpecification::Matrix(_) => "MATRIX",
+                GateSpecification::Permutation(_) => "PERMUTATION",
+            }
+        )?;
+        write!(f, "{}", self.specification)
```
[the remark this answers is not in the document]
MarquessV: How much do these allocations affect performance. A few of the other suggestion you gave made things simpler but I find this alternative significantly less readable/maintainable, so I'm looking for a justification.

quil-rs/src/instruction/mod.rs
```diff
@@ -173,6 +177,23 @@ pub struct Calibration {
...
 
+impl fmt::Display for Calibration {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let parameter_str = get_expression_parameter_string(&self.parameters);
+        write!(
+            f,
+            "DEFCAL {}{} {}:",
+            self.name,
+            parameter_str,
+            format_qubits(&self.qubits)
```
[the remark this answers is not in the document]
MarquessV: None of these really were, just moved. I applied most of your suggestions, but I'll create an issue to followup on the rest. See [my comment](https://github.com/rigetti/quil-rs/pull/140#discussion_r1122299270) asking for clarity of the performance improvement when compared to readability/maintainability.

quil-rs/src/instruction/gate.rs
```diff
@@ -0,0 +1,223 @@
...
+    }
+}
+
+/// The
```
jselig-rigetti, line 203: image

quil-rs/src/instruction/gate.rs
```diff
@@ -0,0 +1,222 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        use GateType::*;
+        write!(
+            f,
+            "{}",
+            match self {
+                Matrix => "MATRIX",
+                Permutation => "PERMUTATION",
+            }
+        )
```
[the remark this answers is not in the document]
jselig-rigetti: don't need the `use GateType::*;` import anymore with this change tho right?

quil-rs/src/instruction/gate.rs
MarquessV: 😂

quil-py/src/instruction/arithmetic.rs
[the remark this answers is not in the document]
MarquessV: I went ahead and used `py_wrap_simple_enum` where it was applicable.

quil-rs/src/instruction/mod.rs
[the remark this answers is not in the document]
MarquessV: Issue #149

quil-rs/src/instruction/gate.rs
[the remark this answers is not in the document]
MarquessV: I'm going to resolve this. We can use this to resolve #149 and discuss further there.

quil-rs/src/program/error/mod.rs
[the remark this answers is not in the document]
MarquessV: Created #150 to tracj

quil-py/src/program/mod.rs
```diff
@@ -0,0 +1,197 @@
...
+            .to_instructions(true)
+            .iter()
+            .filter_map(|inst| match inst {
+                Instruction::GateDefinition(gate_def) => Some(gate_def.to_python(py)),
+                _ => None,
+            })
+            .collect()
+    }
+
+    fn dagger(&self) -> PyResult<Self> {
```
jselig-rigetti, line 120: Is there a reason this isn't a public getter?

```suggestion
    #[getter]
    pub fn dagger(&self) -> PyResult<Self> {
```

MarquessV: Good catch, it should be `pub` for consistency. I won't add getter because it's not really a property of the `Program`, it returns a transposed copy of the `Program`

kalzoo (approved):

PR #142, opened 2023-01-24
Fix: correctly expand delays

### LINKED ISSUE #141: Delays aren't correctly expanded
We've found that the following program:

```
DEFCAL I q:
    DELAY q 4e-8
    I 0
```

which _should_ expand into `DELAY 0 4e-8`, instead erroneously leaves the variable `q` unresolved:

```
failures:

---- program::calibration::tests::expansion stdout ----
thread 'program::calibration::tests::expansion' panicked at 'assertion failed: `(left == right)`
  left: `"DELAY q 4e-8\n"`,
 right: `"DELAY 0 4e-8\n"`', src/program/calibration.rs:407:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

genos:
Closes #141 

This adds one test case, which currently fails:

```
failures:

---- program::calibration::tests::expansion stdout ----
thread 'program::calibration::tests::expansion' panicked at 'assertion failed: `(left == right)`
  left: `"DELAY q 4e-8\n"`,
 right: `"DELAY 0 4e-8\n"`', src/program/calibration.rs:407:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

and subsequently modifies `CalibrationSet.expand` to handle `DELAY` instructions similarly to `GATE` instructions.

### DIFF
--- a/src/program/calibration.rs
+++ b/src/program/calibration.rs
@@ -17,8 +17,8 @@ use std::collections::HashMap;
 use crate::{
     expression::Expression,
     instruction::{
-        Calibration, Gate, GateModifier, Instruction, MeasureCalibrationDefinition, Measurement,
-        Qubit,
+        Calibration, Delay, Gate, GateModifier, Instruction, MeasureCalibrationDefinition,
+        Measurement, Qubit,
     },
 };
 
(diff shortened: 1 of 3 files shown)


src/program/calibration.rs
```diff
@@ -115,6 +115,19 @@ impl CalibrationSet {
                                     }
                                 }
                             }
+                            if let Instruction::Delay(Delay { qubits, .. }) = instruction {
```
kalzoo, line 118: Can you switch this to a `match` and incorporate the `Gate`, now that it's the second case to handle?

In fact, they can be the same match arm with `|`, since `qubits` is of the same type and the code is the same.

genos: Great call! See fb04b8f

kalzoo (approved):
Thanks @genos !

PR #153, opened 2023-03-07
Refactor `ProgramError`

MarquessV:
The existing `ProgramError<T>` was created to handle parsing errors but as the Program API expanded, it grew beyond that purpose. 

The existing error type has been renamed to `ProgramParsingError<T>` and any variants that weren't related to parsing have been removed. There is a new `ProgramError` enum that contains a variant for each possible error when working with a program. It also implements `From<ProgramParsingError<T>` so the public API doesn't have to deal with the generics. This makes it possible to use `ProgramError` fairly easily in the python bindings.

### DIFF
--- a/quil-rs/src/program/error/mod.rs
+++ b/quil-rs/src/program/error/mod.rs
@@ -27,21 +27,16 @@ pub use result::{disallow_leftover, map_parsed, recover};
 pub use syntax::SyntaxError;
 
 /// Errors that may occur while parsing a [`Program`](crate::program::Program).
-// TODO: [`Program`] should have its own error type: https://github.com/rigetti/quil-rs/issues/150
 #[derive(Debug, PartialEq)]
-pub enum ProgramError<T> {
+pub enum ProgramParsingError<T> {
     InvalidCalibration {
         instruction: Instruction,
         message: String,
     },
-    RecursiveCalibration(Instruction),
     Syntax(SyntaxError<T>),
-    InvalidQuiltInstruction(Instruction),
-    InvalidProtoQuilInstruction(Instruction),
-    UnsupportedOperation(Instruction),
 }
 
-impl<T> From<LexError> for ProgramError<T>
+impl<T> From<LexError> for ProgramParsingError<T>
 where
     T: fmt::Debug,
 {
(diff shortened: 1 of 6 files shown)


quil-rs/src/program/error/mod.rs
```diff
@@ -27,21 +27,16 @@ pub use result::{disallow_leftover, map_parsed, recover};
 pub use syntax::SyntaxError;
 
 /// Errors that may occur while parsing a [`Program`](crate::program::Program).
-// TODO: [`Program`] should have its own error type: https://github.com/rigetti/quil-rs/issues/150
 #[derive(Debug, PartialEq)]
-pub enum ProgramError<T> {
+pub enum ProgramParsingError<T> {
```
Shadow53, line 31: The smallest nit -- `ParseProgramError` seems like a more conventional name to me. I think it's because it's `(AttemptedAction)Error` but that's not something I've thought about before.

Shadow53: I looked it up out of curiosity, and [this Rust API Guidelines page](https://rust-lang.github.io/api-guidelines/naming.html#names-use-a-consistent-word-order-c-word-order) seems to be defining the convention I mentioned.

quil-rs/src/program/error/mod.rs
```diff
@@ -50,50 +45,44 @@ where
...
     fn from(err: SyntaxError<T>) -> Self {
         Self::Syntax(err)
     }
 }
 
-impl<T> ProgramError<T> {
+impl<T> ProgramParsingError<T> {
     /// Convert the parsed output into another type.
     ///
     /// This delegates to [`LeftoverError::map_parsed`] when a [`ProgramError::Leftover`] and does
```
Shadow53, line 69: This should probably be changed to the new name.

Shadow53 (commented):
Looks good! Made a note about the new parsing error name, which goes against prescribed convention. A `Refactor -> Rename` should take care of most of it, except for references in comments.

quil-rs/src/program/error/mod.rs
MarquessV: The docstring doesn't quite make sense to me. I think it drifted from the actual implementation as there is no `Leftover` variant. I decided to remove that line entirely.

quil-rs/src/program/error/mod.rs
MarquessV: Yeah, I like that better. I renamed it.

quil-rs/src/program/mod.rs
```diff
@@ -35,7 +35,25 @@ pub mod graph;
 mod memory;
 pub mod type_check;
 
-pub type Result<O> = std::result::Result<O, ProgramError<O>>;
+#[derive(Debug, thiserror::Error, PartialEq)]
+pub enum ProgramError {
+    #[error("{0}")]
+    ParsingError(String),
```
Shadow53, line 41: Would prefer not to have a string here. Can we use `ParseProgramError<Program>` instead?

MarquessV: Done

Shadow53 (approved):
LGTM

kalzoo:
🚀

PR #156, opened 2023-03-09
CalibrationSet Python bindings

MarquessV:
Ports all of the `CalibrationSet` methods to Python bindings.

I uncovered #158 while working on this. The corresponding fix (#159) needs to be merged before this.


quil-rs/src/program/calibration.rs
```diff
@@ -398,6 +416,8 @@ mod tests {
                     "    PRAGMA INCORRECT_ORDERING\n",
                     "DEFCAL MEASURE 0 addr:\n",
                     "    PRAGMA CORRECT\n",
+                    "DEFCAL MEASURE q addr:\n",
+                    "    PRAGMA CORRECT\n",
```
MarquessV, line 420: @kalzoo @notmgsk This is syntax `pyQuil` supports. A variable qubit, followed by an address. `quil-rs` bombs on this, because it only expects an optional fixed Qubit. So in this case it thinks the qubit was left out, reads `q` as the target, and then chokes on `addr`. Is this valid? If so, I have an issue and (potential) fix queued up.

quil-rs/src/program/calibration.rs
```diff
@@ -203,6 +199,28 @@ impl CalibrationSet {
...
+    /// qubit (if any), or otherwise the last-specified one that specified no qubit.
+    pub fn get_match_for_measurement(
+        &self,
+        measurement: &Measurement,
+    ) -> Option<&MeasureCalibrationDefinition> {
+        let mut matching_calibration = None;
+        let mut found_matching_calibration_without_qubit = false;
+        for cal in self.measure_calibrations.iter().rev() {
+            if let Some(cal_qubit) = &cal.qubit {
+                if cal_qubit == &measurement.qubit || matches!(cal_qubit, Qubit::Variable(_)) {
```
MarquessV, line 212: @kalzoo @notmgsk 

This was refactored out of expand, the only addition is the addition of ` || matches!(cal_qubit, Qubit::Variable(_))`. I added this so these instructions like these would match:

```
"DEFCAL MEASURE q dest", "MEASURE 0"
```

Is this correct?

kalzoo: Not quite. The fact that that matches is right (except for the lack of mref in the MEASURE, I think that should be a non-match or not allowed at all) - but the bug is in the `break`. There could be a closer match later on, much like the qubit matching for a gate. That's what `found_matching_calibration_without_qubit = true` is doing - it has a matching calibration but then keeps on looking for a later match. The order of precedence:

1. Match fixed qubit
2. Match variable qubit
3. Match no qubit (I don't recall this case but it's supported here)

The tiebreaker among calibrations of equal precedence is "last wins."

That being said, it looks like there are other pre-existing bugs in this implementation which don't follow the "last wins" rule. So your change just extends an existing bug - this could be fixed now or punted.

quil-py/src/program/calibration.rs
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    pub fn get_match_for_gate(
+        &self,
+        py: Python<'_>,
+        gate_modifiers: Vec<PyGateModifier>,
+        gate_name: &str,
+        gate_parameters: Vec<PyExpression>,
+        gate_qubits: Vec<PyQubit>,
+    ) -> PyResult<Option<PyCalibration>> {
```
kalzoo, line 88: (Corresponding to [this comment](https://github.com/rigetti/pyquil/pull/1536#discussion_r1132781614))

Two things here:

* `get_match_for_measurement` takes a measurement as a single param, but `get_match_for_gate` requires the caller to decompose the gate. Why's that?
* ~~Could this just be `get_match_for_instruction` and do the match on this side? Keeps more of the logic together in this repo rather than farming some of it out to pyQuil~~ I see the benefit of having a narrower type in the return signature.

quil-py/src/program/calibration.rs
```diff
@@ -0,0 +1,134 @@
...
+
+    #[getter]
+    pub fn measure_calibrations(
+        &self,
+        py: Python<'_>,
+    ) -> PyResult<Vec<PyMeasureCalibrationDefinition>> {
+        self.as_inner().measure_calibrations().to_python(py)
+    }
+
+    pub fn expand(
```
kalzoo, line 54: I'll make an issue for this:

The ability to expand an entire program, given that `quil-rs` can already do this and it meshes with the content of this MR. This functionality will be needed by some consumers of pyQuil, who currently do that iteration & expansion within their own programs; the v4 changes _may_ make that harder but a method on `Program` could do it for them.

quil-py/src/program/calibration.rs
MarquessV: Yeah, this is a question I was going to pose once this was out of draft, but worth considering now:

To explain why they are different, `get_match_for_gate` was a pre-existing method, so I just mirrored it in the bindings. `get_match_for_measurement` was tied up in `expand`, so I had to lift it out and chose to just use a `Measurement` directly, rather than decompose it.

My preference is to take the latter approach, so I would like to update `get_match_for_gate` to take a `Gate` instead of the individual components. You happened to write `get_match_for_gate`, so I'll reverse the question back on to you: Is there a benefit to having the caller have to decompose the `Gate` rather than just pass a `Gate` object in? If not, then I'll make the change.

quil-rs/src/program/calibration.rs
MarquessV: I have this "fixed" on the calibration matching side, but the parser needs to be updated as well. Here's the fix, targeting `main` since it would be beneficial there as well: https://github.com/rigetti/quil-rs/pull/159

quil-rs/src/program/calibration.rs
MarquessV: Cool, I've started a fix, but it requires a bit more thought and would detract from the goal of this PR, so I created #160 to track.

MarquessV: I've also made it so that a `MemoryReference` is required to match a `Measurement`.

quil-py/src/program/calibration.rs
MarquessV: Makes sense! Thanks for the issue.

quil-py/src/program/calibration.rs
MarquessV: I refactored `get_match_for_gate` to use a `&Gate` and labeled the commit as a breaking change.

quil-rs/src/program/calibration.rs
MarquessV: #159 was merged and propagated back up to this PR.

PR #159, opened 2023-03-10
allow for variable qubit in DEFCAL MEASURE

### LINKED ISSUE #158: The parser should allow for variable qubits in Measure Calibrations
Currently, the parser only expects a `DEFCAL MEASURE` to be followed by an optional fixed qubit and a required destination. As such, both of these parse as expected:

```
DEFCAL MEASURE 0 addr:
    # ...
```

```
DEFCAL MEASURE addr:
    # ...
```

However, per the quil spec, the qubit can be variable. Meaning the following is valid Quil:

```
DEFCAL MEASURE q addr:
    # ...
```

However, the parse bombs on this. Because it only expects the optional qubit to be an integer, it skips the qubit, thinking it's been omitted, and reads `q` as the destination. It then returns an error because it expects the token following the destination to be `:`, but it reads `addr` instead.

MarquessV:
Closes #158

### DIFF
--- a/src/parser/command.rs
+++ b/src/parser/command.rs
@@ -16,7 +16,7 @@ use nom::{
     branch::alt,
     combinator::{map, opt},
     multi::{many0, many1, separated_list0, separated_list1},
-    sequence::{delimited, preceded, tuple},
+    sequence::{delimited, pair, preceded, tuple},
 };
 
 use crate::{
@@ -205,13 +205,11 @@ pub(crate) fn parse_defcal_gate<'a>(
 pub(crate) fn parse_defcal_measure<'a>(
     input: ParserInput<'a>,
 ) -> InternalParserResult<'a, Instruction> {
-    let (input, (qubit, destination)) = map(
-        tuple((parse_qubit, opt(token!(Identifier(v))))),
-        |(a, b)| match (a, b) {
-            (qubit, Some(destination)) => (Some(qubit), destination),
-            (destination, None) => (None, destination.to_string()),
-        },
-    )(input)?;
+    let (input, params) = pair(parse_qubit, opt(token!(Identifier(v))))(input)?;
+    let (qubit, destination) = match params {
+        (qubit, Some(destination)) => (Some(qubit), destination),
+        (destination, None) => (None, destination.to_string()),
+    };
     let (input, _) = token!(Colon)(input)?;
     let (input, instructions) = parse_block(input)?;
     Ok((
--- a/src/parser/command.rs
+++ b/src/parser/command.rs
@@ -206,11 +206,10 @@ pub(crate) fn parse_defcal_measure<'a>(
     input: ParserInput<'a>,
 ) -> InternalParserResult<'a, Instruction> {
     let (input, (qubit, destination)) = map(
-        tuple((opt(parse_qubit), opt(token!(Identifier(v))))),
+        tuple((parse_qubit, opt(token!(Identifier(v))))),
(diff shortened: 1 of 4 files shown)


src/parser/command.rs
```diff
@@ -205,17 +205,22 @@ pub(crate) fn parse_defcal_gate<'a>(
...
-    let qubit = qubit_index.map(Qubit::Fixed);
-    let (input, destination) = token!(Identifier(v))(input)?;
+    let (input, (qubit, destination)) = map(
+        tuple((opt(parse_qubit), opt(token!(Identifier(v))))),
+        |(a, b)| match (a, b) {
+            (Some(qubit), Some(destination)) => (Some(qubit), destination),
+            (Some(destination), None) => (None, destination.to_string()),
+            _ => panic!("Implement this error"),
+        },
+    )(input)?;
```
MarquessV, line 215: This is my first time really trying to parse something non-trivial in quil-rs. Is this acceptable robust? All existing tests pass, plus a new one for the issue this code solves. I don't love the `panic!` but I went with it because thats how all the errors inside of a `map` are implemented elsewhere.

jselig-rigetti: Looks like you could use [`map_res`](https://docs.rs/nom/latest/nom/combinator/fn.map_res.html) ?

```rs
    let (input, (qubit, destination)) = map_res(
        tuple((opt(parse_qubit), opt(token!(Identifier(v))))),
        |(a, b)| match (a, b) {
            (Some(qubit), Some(destination)) => Ok((Some(qubit), destination)),
            (Some(destination), None) => Ok((None, destination.to_string())),
            _ => Err(InternalError::from_kind(
                input,
                ParserErrorKind::UnexpectedEOF("No DEFCAL target"),
            )),
        },
    )(input)?;
```

jselig-rigetti: Although, since the first token is not optional:

```rs
    let (input, (qubit, destination)) = map(
        tuple((parse_qubit, opt(token!(Identifier(v))))),
        |(a, b)| match (a, b) {
            (qubit, Some(destination)) => (Some(qubit), destination),
            (destination, None) => (None, destination.to_string()),
        },
    )(input)?;
```

MarquessV: Oh, that last one is clever, nice thinking.

src/parser/command.rs
```diff
@@ -205,17 +205,21 @@ pub(crate) fn parse_defcal_gate<'a>(
 pub(crate) fn parse_defcal_measure<'a>(
     input: ParserInput<'a>,
 ) -> InternalParserResult<'a, Instruction> {
-    let (input, qubit_index) = opt(token!(Integer(v)))(input)?;
-    let qubit = qubit_index.map(Qubit::Fixed);
-    let (input, destination) = token!(Identifier(v))(input)?;
+    let (input, (qubit, destination)) = map(
+        tuple((parse_qubit, opt(token!(Identifier(v))))),
```
Shadow53, line 209: I believe this also exists, might be slightly faster:

```suggestion
        pair(parse_qubit, opt(token!(Identifier(v)))),
```

Shadow53 (commented):
Left one suggestion about `nom` code. I don't know Quil enough to speak on whether the logic is semantically correct.

src/parser/command.rs
MarquessV: Nice, done!

kalzoo (approved):

PR #161, opened 2023-03-10
use link time optimization for release builds

MarquessV:
Does this offer a performance improvement?


kalzoo:
Nice find and fix!

PR #164, opened 2023-03-13
Support Qubit & Label Placeholders

### LINKED ISSUE #147: Support Qubit & Label Placeholders
`pyQuil` has the capability to insert placeholders for qubits and labels into a program, and `quil-rs` should support the same concept.

These have no analogue in Quil itself and no serialized form, but are an aid in the construction of valid Quil programs. Both serve as a reference which defers the assignment of a concrete value until later.

## Labels

A label placeholder is useful for writing portions of a program in isolation without the concern that a label will conflict with labels in other parts of the program. This is especially important for support of dynamic control flow programs. Imagine that a user may want a program with multiple loops, each of which starts with a label `@loop-start`. 

A label must be unique within a program, so if the user must assign a concrete label value at the time of label construction they must consider all labels already present within the program. With a placeholder, though, they can use Placeholder("loop-start") in several places, and allow the program to distinguish these prior to serialization - perhaps by adding a suffix.

## Qubits 

Similarly, a routine within a program may want to reference some set of qubits without _yet_ resolving which ones those are. Like labels, this allows for qubits to be isolated to a particular scope within a program while remaining unique at the program scope.

# API

Placeholders can be allocated as such:

```rust
QubitPlaceholder::new();
LabelPlaceholder::new("label");
```

They can then be used within the relevant enums. This will be a breaking change:

```rust
enum Qubit {
  Fixed(...),
  Variable(...),
  Placeholder(QubitPlaceholder)
}

enum Label {
  Fixed(String), // This is the existing use case, now as an enum variant
  Placeholder(LabelPlaceholder), // This is new
}
```

These placeholders cannot be resolved in isolation; they must be resolved within / by a Program. So add a relevant method:

```rust
impl Program {
  pub fn resolve_label_placeholders(&mut self) {
    // collect all the existing labels in the program, and then iterate through placeholders and resolve each one to a string label, appending a numeric suffix to make each unique as needed.
  }

  pub fn resolve_qubit_placeholders(&mut self) {
    // collect all the existing fixed qubits in the program and then iterate through, resolving each to the lowest-available fixed qubit index
  }

  pub fn resolve_placeholders(&mut self) {
    self.resolve_label_placeholders();
    self.resolve_qubit_placeholders();
  }
}

impl fmt::Display for Program {
   // ... resolve_placeholders is called prior to display, because there is no Quil-spec-defined way to serialize a label or qubit placeholder
}
```

kalzoo:
What is a placeholder for?

It's when a program author wants to say something like "I want _this_ instruction and _that_ instruction to use the same qubit, but I don't want to decide _which precise_ qubit that is yet." This is valuable for composition of a complex program from reusable pieces. Placeholders get created at a sub-program scope, and resolved to concrete values at a whole-program scope.

This PR supports placeholders in the following ways:

* Alongside `Fixed` and `Variable` (both of which are defined in Quil syntax), a Qubit may also now be a `Placeholder`. To the user, this is an opaque reference, subject only to the constraint that two copies of the same placeholder (two cloned from the same original) are considered _equal_, and all other pairs are considered _inequal_.
* Labels are now not just wrappers for fixed strings - they also now have placeholders as well. Here, a `Placeholder` is initialized with a string as a "base label" - this is the label that will be modified in some way, if needed, to be resolved to a fixed string label unique within the program.

## Implementation

Both of these are implemented as an `Arc<Pin<Box<T>`:

* `Arc` to allow sharing of a placeholder among many locations within the program. Without that feature, placeholders would be mostly pointless.
* `Pin` to ensure that the address of the `Box` stays the same. We use pointer (address) equality to compare equality between two placeholders, so we need to make sure that that address stays the same. _Note: I think this might be overly cautious, given the wrapping `Arc`. The user cannot perform pointer equality testing themselves because the field is private, so there's no case in which the pointer would be stored and then later checked for equality._
* `Box` just because that's what the `Pin` needs - a heap allocation for an address. Stack pinning doesn't make sense here.
* Then, as the inner type `T`:
  * `T = usize` for `QubitPlaceholder` with a thought to make serialization work pseudo-correctly; I started with `()` and may return to that. Serialization is not implemented here.
  * `T = String` for `LabelPlaceholder`, because that's the `base_label` to be suffixed until a unique label is found.

## The Problem

`Program` is now being pulled in two different ways for two different use cases, and as of this PR (and feature) those are apparently incompatible:

* Up the stack: `pyQuil` already has the concept of placeholders, and as its functionality is brought down into `quil-rs` that needs to be ported as well. Placeholders are an application development convenience and are not a part of Quil itself.
* Down the stack: `quil-rs` was written _first_ to parse and write Quil programs as text. `Program` represents a parsed program as a Rust struct.

## Questions

* How should `Display` be implemented for `Label`, `Qubit`, and (transitively) `Instruction` and `Program`?
  * Currently, this outputs syntactically correct Quil in a way that can be round-tripped. However, there is no way to do so with placeholders - these are not valid Quil.
    * Option 1: Break the expectation that `Display` outputs Quil
    * Option 2: Resolve placeholders at display time
    * Option 3: Convert fallibly from these structures to those which do map directly to Quil. See the next point: 
* Should we split `Program` into `Program` for application users and `ParserProgram` to faithfully represent a Quil program?
  * This would let the `Program` include higher-level conveniences like placeholders, while allowing `ParserProgram` to infallibly output Quil. `Program` -> `ParserProgram` would either coerce (resolve) "extra" functions to basic Quil, or would fail.

TODO:

- [x] Display implementation
- [x] Label placeholders
- [ ] Program::get_used_qubits refactor

Closes #147

### DIFF
--- a/src/instruction.rs
+++ b/src/instruction.rs
@@ -1597,6 +1597,7 @@ RX(%a) 0",
 
         use crate::instruction::{Label, LabelPlaceholder, Qubit, QubitPlaceholder};
 
+        #[allow(clippy::redundant_clone)]
         #[test]
         fn label() {
             let placeholder_1 = LabelPlaceholder::new(String::from("label"));
@@ -1627,6 +1628,7 @@ RX(%a) 0",
             assert_eq!(label_2, Label::Placeholder(placeholder_2));
         }
 
+        #[allow(clippy::redundant_clone)]
         #[test]
         fn qubit() {
             let placeholder_1 = QubitPlaceholder::default();
--- a/src/instruction.rs
+++ b/src/instruction.rs
@@ -15,14 +15,15 @@
 use nom_locate::LocatedSpan;
 use serde::{Deserialize, Serialize};
 use std::borrow::Cow;
+use std::collections::HashMap;
 use std::collections::HashSet;
 use std::str::FromStr;
 use std::sync::Arc;
-use std::{collections::HashMap, fmt};
 
 use crate::expression::Expression;
 use crate::parser::{common::parse_memory_reference, lex, ParseError};
 use crate::program::{disallow_leftover, frame::FrameMatchCondition, SyntaxError};
+use crate::quil::{Quil, ToQuilError, ToQuilResult};
 
 #[cfg(test)]
 use proptest_derive::Arbitrary;
@@ -34,13 +35,15 @@ pub enum ArithmeticOperand {
     MemoryReference(MemoryReference),
 }
 
-impl fmt::Display for ArithmeticOperand {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        match &self {
-            ArithmeticOperand::LiteralInteger(value) => write!(f, "{}", value),
-            ArithmeticOperand::LiteralReal(value) => write!(f, "{}", value),
-            ArithmeticOperand::MemoryReference(value) => write!(f, "{}", value),
-        }
+impl Quil for ArithmeticOperand {
+    type Display = String;
+
+    fn to_quil(&self) -> ToQuilResult<Self::Display> {
+        Ok(match &self {
+            ArithmeticOperand::LiteralInteger(value) => format!("{}", value),
+            ArithmeticOperand::LiteralReal(value) => format!("{}", value),
+            ArithmeticOperand::MemoryReference(value) => format!("{}", value.to_quil()?),
+        })
     }
 }
 
@@ -52,14 +55,16 @@ pub enum ArithmeticOperator {
     Multiply,
 }
 
-impl fmt::Display for ArithmeticOperator {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        match &self {
-            ArithmeticOperator::Add => write!(f, "ADD"),
-            ArithmeticOperator::Divide => write!(f, "DIV"),
-            ArithmeticOperator::Multiply => write!(f, "MUL"),
-            ArithmeticOperator::Subtract => write!(f, "SUB"),
-        }
+impl Quil for ArithmeticOperator {
+    type Display = &'static str;
+
+    fn to_quil(&self) -> ToQuilResult<Self::Display> {
+        Ok(match &self {
+            ArithmeticOperator::Add => "ADD",
+            ArithmeticOperator::Divide => "DIV",
+            ArithmeticOperator::Multiply => "MUL",
+            ArithmeticOperator::Subtract => "SUB",
+        })
     }
 }
 
@@ -69,12 +74,14 @@ pub enum BinaryOperand {
     MemoryReference(MemoryReference),
 }
 
-impl fmt::Display for BinaryOperand {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        match &self {
-            BinaryOperand::LiteralInteger(value) => write!(f, "{}", value),
-            BinaryOperand::MemoryReference(value) => write!(f, "{}", value),
-        }
+impl Quil for BinaryOperand {
+    type Display = String;
+
+    fn to_quil(&self) -> ToQuilResult<Self::Display> {
+        Ok(match &self {
+            BinaryOperand::LiteralInteger(value) => format!("{}", value),
+            BinaryOperand::MemoryReference(value) => value.to_quil()?,
+        })
     }
 }
 
@@ -84,13 +91,15 @@ pub enum BinaryOperator {
     Ior,
     Xor,
 }
-impl fmt::Display for BinaryOperator {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        match &self {
-            BinaryOperator::And => write!(f, "AND"),
-            BinaryOperator::Ior => write!(f, "IOR"),
-            BinaryOperator::Xor => write!(f, "XOR"),
-        }
+impl Quil for BinaryOperator {
+    type Display = &'static str;
+
+    fn to_quil(&self) -> ToQuilResult<Self::Display> {
+        Ok(match &self {
+            BinaryOperator::And => "AND",
+            BinaryOperator::Ior => "IOR",
+            BinaryOperator::Xor => "XOR",
+        })
     }
 }
 
@@ -100,12 +109,14 @@ pub enum UnaryOperator {
     Not,
 }
 
-impl fmt::Display for UnaryOperator {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        match &self {
-            UnaryOperator::Neg => write!(f, "NEG"),
-            UnaryOperator::Not => write!(f, "NOT"),
-        }
+impl Quil for UnaryOperator {
+    type Display = &'static str;
+
+    fn to_quil(&self) -> ToQuilResult<Self::Display> {
+        Ok(match &self {
+            UnaryOperator::Neg => "NEG",
+            UnaryOperator::Not => "NOT",
+        })
     }
 }
 
@@ -116,13 +127,15 @@ pub enum ComparisonOperand {
     MemoryReference(MemoryReference),
 }
 
-impl fmt::Display for ComparisonOperand {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        match &self {
-            ComparisonOperand::LiteralInteger(value) => write!(f, "{}", value),
-            ComparisonOperand::LiteralReal(value) => write!(f, "{}", value),
-            ComparisonOperand::MemoryReference(value) => write!(f, "{}", value),
-        }
+impl Quil for ComparisonOperand {
+    type Display = String;
+
+    fn to_quil(&self) -> ToQuilResult<Self::Display> {
+        Ok(match &self {
+            ComparisonOperand::LiteralInteger(value) => format!("{}", value),
+            ComparisonOperand::LiteralReal(value) => format!("{}", value),
+            ComparisonOperand::MemoryReference(value) => value.to_quil()?,
+        })
     }
 }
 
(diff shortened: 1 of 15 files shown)


src/instruction.rs
```diff
@@ -1080,19 +1131,91 @@ mod test_instruction_display {
...
+    //     }
+    // }
+}
+
 impl fmt::Display for Qubit {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         use Qubit::*;
         match self {
             Fixed(value) => write!(f, "{}", value),
+            Placeholder(_) => todo!(),
```
MarquessV, line 1185: ### Option 1 - Break the expectation that Display outputs Quil

This mirrors what `pyQuil` does today, and I'm assuming would give rise to something like a `Quil` trait with a `to_quil() -> Result<String, Err>`. I like how the trait would let us implement a common interface for a `Program` and each of the instruction types.

### Option 2 - Resolve at Display time

Probably the quickest approach in terms of development time. I'm not sure if I love having to bake in "magic" like that in a Display function, but since it would have to clone the `Program` at least it wouldn't unexpectedly mutate the object. The cloning itself isn't fantastic though.

### Option 3 - Introduce a `ParserProgram`

I would prefer Option 1 to this. Mainly because a `ParserProgram` only solves this at the program level and leaves us with the same problem for each of the `Instruction` types. The parser can/should only parse valid Quil so outside of a bug, I don't see how it would build a `Program` that couldn't be turned back into valid Quil. If I remember correctly, the parser parses `Instruction`s, not a `Program`, so a `ParserProgram` wouldn't help guarantee that the parser couldn't mistakenly use something like a program either.

kalzoo: ## Option 1

Agreed, that could work out nicely - the question is then what the standard is for `Display` and thus `to_string()`, if we even still implement those.

## Option 2

We may be able to avoid the clone by resolving each instruction as it's displayed, so instructions that don't require any resolution aren't mutated or copied.

## Option 3

Ah, to clarify, the `ParserProgram` would contain`ParserInstruction`s, which would be infallibly and unambiguously represented in Quil. @Shadow53 suggested the name `QuilProgram` and thus `QuilInstruction` instead.

It might look like

```rust
struct QuilProgram(Vec<QuilInstruction>); // Since we don't need to keep track of DECLARE separately from PULSE, etc
```

It's a simpler wrapper that may not actually need to be public, if I'm thinking about it right. I don't think this change would require any changes in pyQuil; it would be a tool internal to the library and the bindings.

notmgsk: I am leaning towards option 1, and removing the `Display` implementation altogether (replacing it with a `to_quil()` method which fails if there are unresolved placeholders).

Agreed option 2 feels like potentially unexpected behaviour.

Option 3 sounds nice because types are nice, but I don't think it is as simple/clean as option 1 without providing more.

kalzoo: Okay, I'm on board, if we do it as @notmgsk says here - remove `Display` altogether, which is breaking but prevents confusion. There should also likely be a `from_quil` trait which does the parsing, rather than `FromStr`.

src/instruction.rs
```diff
@@ -1080,19 +1131,91 @@ mod test_instruction_display {
...
+
+/// An opaque placeholder for a qubit whose index may be assigned
+/// at a later time.
+#[derive(Clone, Debug, Eq)]
+pub struct QubitPlaceholder(QubitPlaceholderInner);
+
+impl Default for QubitPlaceholder {
+    fn default() -> Self {
+        use rand::Rng;
+        let value = rand::thread_rng().gen();
```
genos, line 1201: It's opaque so the details shouldn't matter too much, and I doubt we'd ever create enough of these to worry about collisions, but: do we need these to be pseudorandom? Why not a UUID, or even just increment a counter?

```rust
use std::sync::atomic::{AtomicUsize, Ordering};

impl Default for QubitPlaceholder {
    fn default() -> Self {
        static COUNTER: AtomicUsize = AtomicUsize::default();
        let value = COUNTER.fetch_add(1, Ordering::Relaxed);
        Self(Arc::new(Box::pin(value)))
    }
}
```

src/instruction.rs
```diff
@@ -535,7 +537,53 @@ pub struct UnaryLogic {
...
+impl std::hash::Hash for LabelPlaceholder {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        (&*self.0 as *const _ as usize).hash(state);
+    }
+}
+
+impl PartialEq for LabelPlaceholder {
+    #[allow(clippy::ptr_eq)]
+    fn eq(&self, other: &Self) -> bool {
+        &*self.0 as *const _ == &*other.0 as *const _
```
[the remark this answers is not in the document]
kalzoo: It sure could, my thinking was that this would keep the implementation visually equivalent to `Hash`. Happy to change it though

src/instruction.rs
```diff
@@ -1080,19 +1131,91 @@ mod test_instruction_display {
...
         use Qubit::*;
         match self {
             Fixed(value) => write!(f, "{}", value),
+            Placeholder(_) => todo!(),
             Variable(value) => write!(f, "{}", value),
         }
     }
 }
 
+type QubitPlaceholderInner = Arc<Pin<Box<usize>>>;
```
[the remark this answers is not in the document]
kalzoo: thankfully, got rid of it. A fixed memory address isn't important if the user never attempts to compare addresses themselves; since they can't access the inner types of the placeholders, that shouldn't be a problem.

src/instruction.rs
kalzoo: I went with ⬆️ that approach. I'm a little unsettled with it, namely in that now a `Placeholder` can be everywhere a `Qubit` is, even when that is not possibly the case (such as when it's being parsed from a string). That means that there are places that have to account for that not-possible possibility; `unreachable!` or `expect` are options of course but I'd like to avoid those as well.

I put two methods on the `Quil` trait:

* `to_quil` is fallible and if it succeeds, it outputs something `Display`.
* `to_quil_or_debug` is available if the struct implements `Debug` and falls back to that representation if Quil stringification fails. This feels better to me than a panic and works well for the tests.

src/instruction.rs
kalzoo: Yep, they don't. The thought here was to support serde serialization but now that that's been ruled out, it's not needed. The inner type is now `()`, so we pretty much just use the Arc for its own address directly.

kalzoo: As far as counter, something would need to store that counter's state, and there's no eligible container as far as we could determine - _maybe_ a static var but that feels messy. Couldn't be a program since a placeholder should be valid and unique across programs or out of the scope of a program entirely

src/quil.rs
```diff
@@ -0,0 +1,29 @@
...
+    /// Return a string in valid Quil syntax if possible or otherwise the debug representation of the item.
+    fn to_quil_or_debug(&self) -> String
+    where
+        Self: std::fmt::Debug,
+    {
+        self.to_quil()
+            .map(|s| s.to_string())
+            .unwrap_or_else(|_| format!("{:#?}", self))
+    }
+}
```
MarquessV, line 17: I'm curious if this is as efficient as it could be? The implementation of `to_quil` looks like it often requires the allocation of a new strings. Maybe trivial for simple implementations, but a lot of instructions are repeatedly calling `format!` in an iterator.

Would something like this would work?

```rs
pub trait Quil {
    fn write(&self, w: &mut String) -> Result<(), ToQuilError>;

    fn try_to_quil(&self) -> Result<String, DisplayError> {
        let mut w = String::new();
        self.write(&mut w)?;
        Ok(w)
    }
}
```

Then, implementations can take advantage of `write!`, writing new data to a buffer instead of allocating and concatenating new strings each time `format!` needs to be called.

```rs
use std::fmt::write;

impl Quil for Instruction {
    fn write(&self, w: &mut String) -> Result<(), ToQuilError> {
        write!(w, "{}", self.value)
    }
}
```

Overall, this seems clearer, and easier to implement, as it doesn't require a custom type definition, and it pretty much looks and works exactly like `Display::fmt`.

MarquessV: Thinking on this a bit more, the argument to `write` could probably be something like `impl Write` to avoid needing to import the trait and to decouple the buffer type from implementations.

src/instruction.rs
```diff
@@ -1296,6 +1457,23 @@ impl Instruction {
             nom::combinator::all_consuming(parse_instruction)(&lexed).map_err(|e| e.to_string())?;
         Ok(instruction)
     }
+
+    pub(crate) fn resolve_placeholders<LR, QR>(&mut self, label_resolver: LR, qubit_resolver: QR)
```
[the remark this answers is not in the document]
kalzoo: Good idea, added, but using closures instead of a map, which should be more flexible and still accommodate the map use case.

src/quil.rs
kalzoo: Yep, that could work! This is the kind of thing I was looking for in our group chat earlier. I'll keep `to_quil`, I think, but give it a default implementation that defers to `write`. What do you think about `write_quil` in place of `write`?

src/instruction.rs
```diff
@@ -1278,6 +1469,44 @@ impl Instruction {
...
+            Instruction::Reset(reset) => match &reset.qubit {
+                Some(qubit) => vec![qubit],
+                None => vec![],
+            },
+            Instruction::Delay(delay) => delay.qubits.iter().collect(),
+            Instruction::Fence(fence) => fence.qubits.iter().collect(),
+            Instruction::Capture(capture) => capture.frame.qubits.iter().collect(),
+            Instruction::Pulse(pulse) => pulse.frame.qubits.iter().collect(),
+            Instruction::RawCapture(raw_capture) => raw_capture.frame.qubits.iter().collect(),
+            _ => vec![],
```
MarquessV, line 1488: `pyQuil` supports using `QubitPlaceholder`s in `Pragma`s. That looks to be missing here, presumably because the `Pragma` instruction in `quil-rs` doesn't distinguish a qubit from any other kind of argument. I'm not sure what `Pragma`s are associated with qubits, and how useful they are, but is that something we should try to port back here?

kalzoo:
Superseded by #266

src/instruction.rs
kalzoo: Yeah, that's a mistake, per para 9.2 in the [Quil spec](https://quil-lang.github.io/), so I'd rather not try to coerce it in here. It should be a breaking change in pyQuil v4. Do you think that'll impact many use cases?

src/instruction.rs
```diff
@@ -1303,6 +1532,25 @@ impl Instruction {
         Ok(instruction)
     }
 
+    pub(crate) fn resolve_placeholders<QR>(
+        &mut self,
+        label_resolver: &dyn Fn(&LabelPlaceholder) -> Option<String>,
+        qubit_resolver: QR,
+    ) where
+        QR: Fn(&QubitPlaceholder) -> Option<u64>,
```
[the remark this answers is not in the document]
kalzoo: probably just a leftover development change; I made both generic params in #266

src/instruction.rs
```diff
@@ -1278,6 +1469,44 @@ impl Instruction {
...
+            Instruction::Delay(delay) => delay.qubits.iter().collect(),
+            Instruction::Fence(fence) => fence.qubits.iter().collect(),
+            Instruction::Capture(capture) => capture.frame.qubits.iter().collect(),
+            Instruction::Pulse(pulse) => pulse.frame.qubits.iter().collect(),
+            Instruction::RawCapture(raw_capture) => raw_capture.frame.qubits.iter().collect(),
+            _ => vec![],
+        }
+    }
+
+    /// Return immutable references to the [`Qubit`]s contained within an instruction
```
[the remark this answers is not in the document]
kalzoo: Fixed in #266

src/instruction.rs
kalzoo: Fortunately this has been obviated altogether

src/instruction.rs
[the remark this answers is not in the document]
kalzoo: Changed in #266

PR #168, opened 2023-03-14
Add API, type hints for existing bindings.

### LINKED ISSUE #162: Instruction API: Frames
`AttributeValue`, `FrameAttributes`, `FrameDefinition`, `FrameIdentifier` all need new constructors and Python bindings.

### LINKED ISSUE #167: Add Instruction API and type hints for existing bindings
#140 introduced `quil-py`, and with it a bunch of bindings. As that PR evolved, some of the initial bindings weren't given the same treatment as the newer ones. They should all have an API, be usable as subclasses, and have type hints.

MarquessV:
This is a pretty large PR, but the approach was very mechanical.

For any existing module in `quil-py`: 

1) I checked that it mirrored `quil-rs`, and if not, reconciled them to match. In most cases, this was a matter of refactoring `quil-rs` instructions out of `mod.rs` and into a submodule. There were a few cases where `quil-py` had a module in the wrong place. For example, the `expression` module was in `instructions` instead of it's own top-level module. In these cases, I refactored `quil-py` to match.

2) Made sure each binding had a `new` constructor, derived the same attributes as its corresponding `quil-rs` type, and wasn't missing any derived implementations (e.g. `impl_hash!` for `Hash` or `fn __richcmp__` for `PartialEq`/`Eq`).

3) Added missing variants to the `PyInstruction` enum, if the module contained any of the needed types.

4) Made sure they all had type hints and didn't fail any `stubtest` checks. As part of this I had to flatten the type stubs. We can't break them into private modules for organization because those private modules aren't actually exported by the package. Technically, there are still two `stubtest` failures. The bindings export `quil.quil` and `quil.validation.quil.validation.identifier` as empty modules. I've seen this in the `qcs_sdk` as well, and I think it's a bug with either the pyo3 macros or pyo3 itself. We could silence the errors by adding dummy stubs, but I'd rather have the errors and not expose them to users. I created [this issue](https://github.com/rigetti/rigetti-pyo3/issues/21) to follow-up.

I also handled a few extra things, like building out the `Expression` bindings, because they were in the path and it made sense to just get it out of the way here.

Closes #167, and because the frame instruction bindings were pre-existing, this also closes #162. This also adds a `Makefile.toml` with a workflow for running `stubtest` to validate the type stubs.


quil-py/Makefile.toml
```diff
@@ -0,0 +1,21 @@
...
+
+[tasks.poetry-install]
+command = "poetry"
+args = ["install"]
+
+[tasks.install-quil]
+command = "poetry"
+args = ["run", "maturin", "develop"]
+
+[tasks.stubtest]
```
kalzoo, line 12: Should these be documented? ie in Contributing.md

kalzoo (approved):

quil-py/Makefile.toml
MarquessV: Yep! Good call. I've documented it. I've also made it so that `stubtest-flow` is part of the default flow, so that one can just run `cargo make` and get the appropriate validation on both crates.

PR #172, opened 2023-03-17
Support Declare OFFSETs

### LINKED ISSUE #171: Declare should support offsets
Per the [Quil spec](https://github.com/quil-lang/quil/blob/master/spec/Quil.md#aliased-declaration-with-offset), one can optionally add an `OFFSET {int} {ScalarType}` to declare a memory region that coincides with an intermediate segment of a shared memory region.

MarquessV:
Closes #171


jselig-rigetti (approved):

quil-rs/benches/parser.rs
```diff
@@ -43,8 +43,6 @@ fn from_corpus() -> Vec<QuilBenchConfig> {
         init_submodules()
     }
 
-    dbg!(&corpus_dir);
```
kalzoo, line 46: 👍

kalzoo (approved):
Smooth addition, all the way from python bindings to the parser 🕺 . Looks great!

PR #173, opened 2023-03-21
Support pickling a `PyProgram` by implementing `__getstate__` and `__setstate__` dunder methods.

MarquessV:
This allows a `PyProgram` to be pickled by implementing the `__getstate__` and `__setstate__` dunder methods.


quil-py/src/program/mod.rs
```diff
@@ -181,6 +191,18 @@ impl PyProgram {
             _ => py.NotImplemented(),
         }
     }
+
+    pub fn __getstate__<'a>(&self, py: Python<'a>) -> &'a PyBytes {
+        PyBytes::new(py, self.as_inner().to_string(true).as_bytes())
```
MarquessV, line 196: This approach leverages that fact that we can currently infallibly convert a `Program` to a string. Looking ahead to #164, this won't always be the case. The only alternative I can think of is to use `serde` and derive `Serialize` and `Deserialize` the `Program`, which would mean deriving them on every instruction and its constituent types.

kalzoo: `serde` seems like a good fit for the use case - but it does (softly) lock us into the fields and types we have now. That is, we'll have one more API to preserve across changes.

Another option - fail to pickle in those cases?

kalzoo (approved):
I note that it's still  marked as draft, but if that's just because of the placeholder question that work could be deferred to that MR. The `to_string(bool)` API should go away, forcing us to update this here, so it won't get lost.

quil-py/src/program/mod.rs
MarquessV: Luckily. I was able to [derive the traits](https://github.com/rigetti/quil-rs/pull/173/commits/b605d024bed23108b01234c2acbcb217da66f5c3) without too much trouble thanks to vim macros.

I do see your point. Failing to pickle on placeholders is an option. I don't think pyQuil has this limitation, so it would be a soft break. It does feel heavy handed to apply `Serialize` and `Deserialize` everywhere in `quil-rs` just to support pickling on the Python side. Do you have a preference either way? I'm a bit torn.

MarquessV: Yeah, I'll revert the serialization stuff and merge the version of the PR you approved, with an additional note about these findings in the form of a comment.

PR #175, opened 2023-03-21
174 fix expression defgate parsing

MarquessV:
There were two issues to solve here:

1. Ignore casing for function calls

This may be controversial. The Quil spec only ever refers to these functions in their lowercase form, but `pyQuil` and its parser accepts any casing.

2. Fix parameter output

Parameters for `GateDefintion` and `WaveDefinition` are stored in a `Vec<String>`. These parameters are Quil identifiers prefixed with a `%` (aka "formal parameters). The parser recognizes these as formal parameters, but stores them on the instruction without the `%`. This meant they were getting printed back out as `Quil` without the `%`, which is incorrect. This fix restores the `%` on output, though it may make more sense to keep it in the parser.

### DIFF
--- a/quil-rs/src/instruction/mod.rs
+++ b/quil-rs/src/instruction/mod.rs
@@ -413,7 +413,7 @@ pub fn get_expression_parameter_string(parameters: &[Expression]) -> String {
         return String::from("");
     }
 
-    let parameter_str: String = parameters.iter().map(|e| format!("{e}")).collect();
+    let parameter_str: String = parameters.iter().map(|e| format!(r"{e}")).collect();
     format!("({parameter_str})")
 }
 
@@ -422,7 +422,11 @@ pub fn get_string_parameter_string(parameters: &[String]) -> String {
         return String::from("");
     }
 
-    let parameter_str: String = parameters.join(",");
+    let parameter_str: String = parameters
+        .iter()
+        .map(|param| format!(r"%{param}"))
+        .collect::<Vec<_>>()
+        .join(",");
     format!("({parameter_str})")
 }
 
--- a/quil-rs/src/instruction/gate.rs
+++ b/quil-rs/src/instruction/gate.rs
@@ -2,6 +2,7 @@ use std::fmt;
 
 use crate::{
     expression::Expression,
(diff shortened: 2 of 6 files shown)


quil-rs/src/instruction/mod.rs
```diff
@@ -413,7 +413,7 @@ pub fn get_expression_parameter_string(parameters: &[Expression]) -> String {
         return String::from("");
     }
 
-    let parameter_str: String = parameters.iter().map(|e| format!("{e}")).collect();
+    let parameter_str: String = parameters.iter().map(|e| format!(r"{e}")).collect();
```
Shadow53, line 416: What's the purpose of the raw string here?

quil-rs/src/instruction/mod.rs
```diff
@@ -422,7 +422,11 @@ pub fn get_string_parameter_string(parameters: &[String]) -> String {
         return String::from("");
     }
 
-    let parameter_str: String = parameters.join(",");
+    let parameter_str: String = parameters
+        .iter()
+        .map(|param| format!(r"%{param}"))
```
Shadow53, line 427: Same question about the `r""` here.

quil-rs/src/instruction/mod.rs
```diff
@@ -422,7 +422,11 @@ pub fn get_string_parameter_string(parameters: &[String]) -> String {
         return String::from("");
```
Shadow53, line 422: Not introduced by these changes, but this should be `String::new()`

Shadow53 (commented):
I'm curious about the use of `r""` over `""` in a couple places, otherwise this looks fine to me.

quil-rs/src/instruction/mod.rs
MarquessV: Oops, good catch, I was testing something else that needed it and forgot about it. Cleaned it up.

quil-rs/src/instruction/mod.rs
kalzoo: got the other one, but this one's still here

kalzoo (approved):
lgtm 👍

PR #177, opened 2023-03-23
Support PAULI-SUM Gate Specifications

MarquessV:


quil-rs/src/parser/common.rs
```diff
@@ -195,6 +195,60 @@ pub(crate) fn parse_permutation<'a>(input: ParserInput<'a>) -> InternalParserRes
     )(input)
 }
 
+/// Parse a collection of [`crate::instruction::PauliWord`]s. These are a special kind of a
+/// [`Token::Identifier`] used in Pauli sum `DEFGATE` specifications where each identifier
+/// is a single string made up of one or more `I`, `X`, `Y`, `Z` characters (e.g. `X`, `YY`),
+/// each representing a Pauli word.
```
MarquessV, line 201: This is the simplest approach I could come up with. A new `Token` variant for Pauli words won't work, because it would have to be parsed before a `Command`/`Identifier`, but the individual characters overlap with Reserved `Identifier`s (Each of the Pauli words correspond to the standard gate of the same name).

quil-rs/src/parser/common.rs
```diff
@@ -195,6 +195,60 @@ pub(crate) fn parse_permutation<'a>(input: ParserInput<'a>) -> InternalParserRes
...
+                    tuple((
+                        parse_pauli_words,
+                        delimited(token!(LParenthesis), parse_expression, token!(RParenthesis)),
+                        many1(token!(Identifier(i))),
+                    )),
+                    |(words, expression, arguments)| PauliTerm {
+                        words,
+                        expression,
+                        arguments,
+                    },
```
MarquessV, line 245: Per the Quil spec, the length of `words` and `arguments` should be the same. Is there a good pattern to enforce this kind of rule in the parser?

If not, maybe this a case where we could use a `new` constructor to perform additional validation in the parser?

MarquessV: Maybe [verify](https://docs.rs/nom/latest/nom/combinator/fn.verify.html)?

kalzoo: The catch will be if the error types are awkward to mesh together, but you had that pattern in use already on a previous MR so I assume it works out alright.

jselig-rigetti: I think using the `PauliTerm::new` here makes sense 👍

quil-rs/src/instruction/gate.rs
```diff
@@ -128,13 +131,53 @@ impl fmt::Display for GateModifier {
...
+    X,
+    Y,
+    Z,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct PauliTerm {
+    pub words: Vec<PauliWord>,
+    pub expression: Expression,
+    pub arguments: Vec<String>,
```
jselig-rigetti, line 147: Similar to https://github.com/rigetti/quil-rs/pull/177/files#r1146660689, should/could these be made private so that using `PauliTerm::new` is required?

quil-rs/src/instruction/snapshots/quil_rs__instruction__gate__test_gate_definition__display@Pauli Sum GateDefinition.snap
```diff
@@ -0,0 +1,9 @@
+---
+source: quil-rs/src/instruction/gate.rs
+expression: gate_def.to_string()
+---
+DEFGATE PauliSumGate(%theta) AS PAULI-SUM:
```
notmgsk, line 5: This is missing `p` and `q` in the signature

quil-rs/src/instruction/gate.rs
```diff
@@ -128,13 +131,53 @@ impl fmt::Display for GateModifier {
...
+pub enum PauliWord {
+    I,
+    X,
+    Y,
+    Z,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct PauliTerm {
+    pub words: Vec<PauliWord>,
```
notmgsk, line 145: I think some of the terminology is wrong here. A `PauliTerm` has a _single_ `PauliWord`, and a `PauliWord` is a string of characters drawn from the set `{I, X, Y, Z}`.

quil-rs/src/instruction/snapshots/quil_rs__instruction__gate__test_gate_definition__display@Pauli Sum GateDefinition.snap
MarquessV: Nice catch! Took a bit of refactoring, but their included now.

quil-rs/src/instruction/gate.rs
MarquessV: Ah yeah, that makes sense. I've renamed the types, let me know what you think.

quil-rs/src/parser/common.rs
MarquessV: Done!

quil-rs/src/parser/common.rs
```diff
@@ -195,6 +195,60 @@ pub(crate) fn parse_permutation<'a>(input: ParserInput<'a>) -> InternalParserRes
...
+/// Parse Pauli sum representation of a `DEFGATE` specification.
+pub(crate) fn parse_pauli_sum<'a>(
+    input: ParserInput<'a>,
+) -> InternalParserResult<'a, Vec<PauliTerm>> {
+    preceded(
+        token!(NewLine),
+        separated_list1(
+            token!(NewLine),
+            preceded(
+                token!(Indentation),
```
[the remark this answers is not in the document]
MarquessV: There is a `parse_block` that returns `Vec<Instruction>` specifically, but nothing more general purpose for that. I grepped for other methods using the `Indentation` token, and most matched this pattern, including matrix specification parser.

quil-rs/src/parser/common.rs
```diff
@@ -195,6 +195,60 @@ pub(crate) fn parse_permutation<'a>(input: ParserInput<'a>) -> InternalParserRes
...
+                map(
+                    tuple((
+                        parse_pauli_words,
+                        delimited(token!(LParenthesis), parse_expression, token!(RParenthesis)),
+                        many1(token!(Identifier(i))),
```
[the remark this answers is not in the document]
MarquessV: Good idea, done.

quil-py/quil/instructions/__init__.pyi
```diff
@@ -471,6 +471,46 @@ class Gate:
...
+@final
+class PauliGate(Enum):
+    I = "I"
+    X = "X"
+    Y = "Y"
+    Z = "Z"
+    @staticmethod
+    def parse(word: str) -> "PauliGate":
+        """
+        Parses a ``PauliWord`` from a string. Raises a ``ParseEnumError`` if the
```
Shadow53, line 483: Should this be `PauliGate`?

quil-rs/src/instruction/gate.rs
```diff
@@ -128,13 +131,67 @@ impl fmt::Display for GateModifier {
...
+    X,
+    Y,
+    Z,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct PauliTerm {
+    pub word: Vec<PauliGate>,
+    pub expression: Expression,
+    pub arguments: Vec<String>,
```
Shadow53, line 147: Does each word correspond to an argument? Would `Vec<(PauliGate, String)>` make sense here?

quil-rs/src/instruction/gate.rs
```diff
@@ -187,24 +257,31 @@ impl GateDefinition {
...
+            GateSpecification::Matrix(_) => writeln!(f, " AS MATRIX:")?,
+            GateSpecification::Permutation(_) => writeln!(f, " AS PERMUTATION:")?,
+            GateSpecification::PauliSum(sum) => writeln!(
+                f,
+                " {} AS PAULI-SUM:",
+                sum.get_arguments()
+                    .into_iter()
+                    .collect::<Vec<String>>()
+                    .join(" ")
+            )?,
```
Shadow53, line 276: Would prefer to avoid string allocation inside of `Display`.

```suggestion
            GateSpecification::PauliSum(sum) => {
                for arg in sum.get_arguments() {
                    write!(f, " {arg}")?;
                }
                writeln!(f, " AS PAULI-SUM:")?;
            )?,
```

quil-py/quil/instructions/__init__.pyi
MarquessV: Yep, good catch.

quil-rs/src/instruction/gate.rs
MarquessV: I tried that initially, and I do like that it fits the data semantically, but the ergonomics around it are a little worse because it doesn't match the syntax as well. You have to zip it on parse, and split it on Display. On top of that tuples don't really play nice with the `rigetti_pyo3` macros, so it felt like a lot of trouble to go through for little tangible benefit. That said, I could be convinced otherwise.

quil-rs/src/instruction/gate.rs
```diff
@@ -128,13 +137,77 @@ impl fmt::Display for GateModifier {
...
+    pub arguments: Vec<String>,
+    pub terms: Vec<PauliTerm>,
+}
+
+impl PauliSum {
+    pub fn new(arguments: Vec<String>, terms: Vec<PauliTerm>) -> Result<Self, GateError> {
+        let term_arguments =
+            HashSet::<String>::from_iter(terms.iter().flat_map(|t| t.arguments.clone()));
+        let argument_set = HashSet::from_iter(arguments.clone().into_iter());
+        let diff: Vec<&String> = term_arguments.difference(&argument_set).collect();
```
MarquessV, line 188: Anyone have the Rust chops to do this without all the cloning? I tried fighting with the borrow checker for a while with no luck. It feels like there should be a better way to do this, but it's not coming to me.

jselig-rigetti: This seems to compile but not sure if it is better 

```rs
        let term_arguments = terms
            .iter()
            .flat_map(|t| t.arguments.iter())
            .collect::<HashSet<_>>();

        let argument_set = arguments.iter().collect::<HashSet<_>>();

        let diff = term_arguments.difference(&argument_set).collect::<Vec<_>>();

        if !diff.is_empty() {
            return Err(GateError::PauliSumArgumentMismatch {
                expected_arguments: arguments.clone(),
                mismatches: diff.into_iter().map(|&v| v.clone()).collect(),
            });
        }
```

jselig-rigetti: Or this? /shrug

```rs
        let diff = terms
            .iter()
            .flat_map(|t| t.arguments.iter())
            .collect::<HashSet<_>>()
            .difference(&arguments.iter().collect::<HashSet<_>>())
            .map(|v| v.to_owned().to_owned()) // probably a better way
            .collect::<Vec<_>>();

        if !diff.is_empty() {
            return Err(GateError::PauliSumArgumentMismatch {
                expected_arguments: arguments,
                mismatches: diff,
            });
        }

        Ok(Self { arguments, terms })
```

Shadow53: In @jselig-rigetti's second example, I think you can replace `.map(|v| v.to_owned().to_owned())` with `.copied().cloned()` -- references are `Copy`, so `copied` should dereference the second reference and `cloned` should convert the leftover references into owned values.

A second option is `.map(|&v| v.to_owned())`, which implicitly dereferences the outer reference via pattern matching.

Shadow53: Cloned this locally and tweaked @jselig-rigetti's suggestion a little to get this, which at least passes `cargo check`:

```rust
        let diff = terms
            .iter()
            .flat_map(|t| t.arguments.iter().map(|(_, a)| a))
            .collect::<HashSet<_>>()
            .difference(&arguments.iter().collect::<HashSet<_>>())
            .copied()
            .cloned()
            .collect::<Vec<_>>();

        if !diff.is_empty() {
            return Err(GateError::PauliSumArgumentMismatch {
                expected_arguments: arguments,
                mismatches: diff,
            });
        }

        Ok(Self { arguments, terms })

```

quil-rs/src/instruction/gate.rs
```diff
@@ -220,13 +226,13 @@ impl fmt::Display for GateSpecification {
...
-                for term in &pauli_sum.0 {
+                for term in &pauli_sum.terms {
                     write!(f, "\t")?;
-                    for word in term.word.iter() {
+                    for (word, _) in term.arguments.iter() {
                         write!(f, "{word}")?;
                     }
                     write!(f, "{}", term.expression)?;
-                    for argument in term.arguments.iter() {
+                    for (_, argument) in term.arguments.iter() {
```
Shadow53: Would it be useful to add methods `words` and `arguments` that return iterators for each of these? I know my suggest for the above conversation would be (slightly) improved with an `arguments()` method.

quil-py/src/instruction/gate.rs
```diff
@@ -149,46 +160,88 @@ impl PyPauliTerm {
...
+            Vec::<(PauliGate, String)>::py_try_from(
+                py,
+                &PyPauliTerm::py_pairs_from_tuples(py, arguments)?,
+            )?,
+            Expression::py_try_from(py, &expression)?,
+        )))
+    }
+
+    // Override the getters/setters generated by [`py_wrap_data_struct!`] so that they
+    // return/take tuples instead of the wrapping [`PyPauliPair`] type.
```
Shadow53, line 176: Does this actually work? Huh.

quil-rs/src/instruction/gate.rs
MarquessV: Nice! Thanks for playing golf! It seems like collecting into a `HashSet` was the way to go after all. I hit a dead end with that approach the first try, but this last suggestion works perfectly. I made a small tweak so I could shove the `clone()` into the error case.

quil-py/src/instruction/gate.rs
MarquessV: Yeah, a convenient, but not ideal, workaround. Added a test to make sure it works as we need it to. I also created [this issue](https://github.com/rigetti/rigetti-pyo3/issues/22) for `rigetti-pyo3`.

quil-rs/src/instruction/gate.rs
MarquessV: Yeah, I like that idea, done.

quil-rs/src/parser/common.rs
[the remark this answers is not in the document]
kalzoo: Gotcha. Might be good to have a parser wrapper to do the newline and tab delimiting to dedupe the code, but that can be a future thing.

kalzoo (approved):
image

PR #186, opened 2023-04-10
feat!: introduce ExecutionDependency::Scheduled

### LINKED ISSUE #185: Better Express Timing Dependencies
Currently, any two instructions with `InstructionRole::RFControl` are presumed to operate on a frame and thus incur an `ExecutionDependency::Frame` between them during `ScheduledProgram` construction. However, that misses the point of this graph.

Each edge should express some condition to be satisfied in the scheduling of the program on a backend. In this case, what `Frame` means is ambiguous. A `PULSE` which follows a `DELAY` should follow it immediately; a `SET-FREQUENCY` which follows a `PULSE` can be executed at any time, subject to other constraints in the program. However, today, those two pairs are expressed the same way, with an `ExecutionDependency::Frame`.

`Frame` should be replaced with a combination of the existing `StableOrdering` and a new `Timing`:

* `Timing` between two instructions whose timing within the pulse program is governed by the Quil-T specification. That is, an instruction must start immediately following the end of the latest preceding instruction with which it shares a `Timing` edge in the graph.
* `StableOrdering` for all other cases, such as between a `SET-FREQUENCY` and `PULSE`.

kalzoo:
Closes #185 

Introduces `ExecutionDependency::Timing` as described in the issue, with updated snapshot tests to reflect the change.

### DIFF
--- a/src/instruction.rs
+++ b/src/instruction.rs
@@ -1303,10 +1303,10 @@ impl Instruction {
         Ok(instruction)
     }
 
-    /// Per the language spec, whether this instruction's timing within the pulse
-    /// program must be precisely controlled, so as to begin exactly on the end of
+    /// Per the Quil-T spec, whether this instruction's timing within the pulse
+    /// program must be precisely controlled so as to begin exactly on the end of
     /// the latest preceding timed instruction
-    pub(crate) fn requires_precise_timing(&self) -> bool {
+    pub(crate) fn is_scheduled(&self) -> bool {
         match self {
             Instruction::Capture(_)
             | Instruction::Delay(_)
--- a/src/instruction.rs
+++ b/src/instruction.rs
@@ -1208,11 +1208,17 @@ impl Instruction {
                     FrameMatchCondition::AnyOfNames(frame_names),
                 ])
             }),
-            Instruction::Fence(Fence { qubits }) => Some(if qubits.is_empty() {
-                FrameMatchCondition::All
-            } else {
-                FrameMatchCondition::AnyOfQubits(Cow::Borrowed(qubits))
-            }),
+            Instruction::Fence(Fence { qubits }) => {
+                if include_blocked {
+                    Some(if qubits.is_empty() {
+                        FrameMatchCondition::All
+                    } else {
+                        FrameMatchCondition::AnyOfQubits(Cow::Borrowed(qubits))
+                    })
+                } else {
+                    None
+                }
+            }
             Instruction::Reset(Reset { qubit }) => {
                 let qubits = match qubit {
                     Some(qubit) => {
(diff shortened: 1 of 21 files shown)


src/program/graphviz_dot.rs
```diff
@@ -413,6 +422,14 @@ CAPTURE 0 \"ro_rx\" test ro
 PULSE 0 \"rf\" test"
         );
 
+        // assert that a block "waits" for a capture to complete even with a pulse after it
+        build_dot_format_snapshot_test_case!(
+            pulse_after_set_frequency,
+            "DECLARE ro BIT
+SET-FREQUENCY 0 \"rf\" 3e9
+PULSE 0 \"rf\" test"
```
notmgsk, line 430:
```suggestion
            r#"DECLARE ro BIT
SET-FREQUENCY 0 "rf" 3e9
PULSE 0 "rf" test"#
```

src/instruction.rs
```diff
@@ -1208,11 +1208,17 @@ impl Instruction {
                     FrameMatchCondition::AnyOfNames(frame_names),
                 ])
             }),
-            Instruction::Fence(Fence { qubits }) => Some(if qubits.is_empty() {
-                FrameMatchCondition::All
-            } else {
-                FrameMatchCondition::AnyOfQubits(Cow::Borrowed(qubits))
-            }),
+            Instruction::Fence(Fence { qubits }) => {
```
notmgsk, line 1211: Is this change for a separate bug?

kalzoo: It's related, actually - this change is necessary to get the right behavior from the test cases. Prior to this, fences "used" frames which meant that they were serialized against one another, which can throw the timing of a program off because it can serialize two otherwise concurrent instructions. 

This change fixes it so FENCE only blocks frames, which makes more intuitive sense as well.

src/program/graph.rs
```diff
@@ -240,8 +240,8 @@ impl Default for PreviousNodes {
     /// (in other words, this instruction cannot be scheduled prior to the start of the instruction block).
```
mhodson-rigetti, line 240: This comment refers to blocking but semantically it looks like the block-start is now of an exclusive "using" type; does the comment need revising?

src/instruction.rs
```diff
@@ -1296,6 +1302,52 @@ impl Instruction {
...
+            | Instruction::CalibrationDefinition(_)
+            | Instruction::CircuitDefinition(_)
+            | Instruction::Convert(_)
+            | Instruction::Comparison(_)
+            | Instruction::Declaration(_)
+            | Instruction::Exchange(_)
+            | Instruction::FrameDefinition(_)
+            | Instruction::Gate(_)
+            | Instruction::GateDefinition(_)
+            | Instruction::Halt
```
mhodson-rigetti, line 1327: A query on the spec; wouldn't a HALT have a precise timing? ie. you would know precisely when the program ends?

mhodson-rigetti:
Code looks neat and tidy; good test coverage.

src/program/graph.rs
kalzoo: Indeed!

src/instruction.rs
kalzoo: You make a good point, but it may have more to do with the docs or choice of names.

Many instructions will have timing fixed at compile time, but whether a halt is fixed to time X or time Y does not materially affect the program, if we see its pulse sequence as essential to its purpose.

Put another way, depending on the semantics of a hardware provider, the HALT may be 10us or 10ms after the previous instruction, but that does not affect the readout results that one would expect to collect.

That’s in contrast to a PULSE, where durations and times may be statically (if parametrically) determined, and which do not depend on hardware provider implementation details.

How could I better communicate that, maybe in docs but ideally in the code itself?

mhodson-rigetti: I think the documentation. As a user, I never write HALT; it's always implicit, and if the hardware back end chooses to add it for some tracking of the type you describe, so be it. The place I would care is Quil-T based pulse timing analysis. One of the things we can do with Quil-T entirely on the user side is mimic pulse timing rules and determine the timing of all of the instructions ourselves. Here, we report total program duration as the end time of the last pulse. But, where HALT is in play (implicitly or explicitly) and has some real time duration as you describe, that real time should be accounted for (start time of HALT was the end time of the last pulse, but then there is some real end time to HALT that is greater?). Which brings me back to the point-- wouldn't you be able to tell me precisely what that time was? If the back end adds 10 microseconds, shouldn't I be able to see that somehow? Is this was "precise" means here, and what is `quil-rs`'s part to play in this workflow (it may be it is simply out of scope at this level).

This was a query of the semantics only, and not a practical issue, so I'm happy whatever you decide.

kalzoo: Thanks for that feedback, @mhodson-rigetti !

Following up from our discussion offline, I've renamed this method to `is_scheduled`, and `ExecutionDependency::Timing` to `ExecutionDependency::Scheduled`.

For anyone following along - "precise timing" vs scheduling is a worthwhile semantic distinction here; only some instructions are considered for "scheduling" as part of the program, and their position within that schedule is essential to the program doing what the author intended. That is now what this function identifies.

All instructions, though, may have "precise times" fixed as part of compilation for a backend. The timing of non-scheduled instructions (like `ADD` or `MOVE`) is flexible within some bounds without affecting the output of the program.

kalzoo:
We're stuck with the CI failure until `criterion 0.4.1` is released; see https://github.com/bheisler/criterion.rs/pull/628

mhodson-rigetti (approved):
Additional changes look good to me!

PR #194, opened 2023-04-17
Type hint improvements

MarquessV:
Some fixes and improvements influenced by resolving some mypy issues in [pyQuil#1566](https://github.com/rigetti/pyquil/pull/1566).


quil-py/quil/instructions/__init__.pyi
```diff
@@ -269,13 +273,12 @@ class ArithmeticOperator(Enum):
...
 
 class Arithmetic:
-    @staticmethod
     def __new__(
         cls,
         operator: ArithmeticOperator,
         destination: ArithmeticOperand,
         source: ArithmeticOperand,
-    ) -> "Arithmetic": ...
+    ) -> Self: ...
```
MarquessV, line 281: This fixes an issue where `mypy` would expect a sub class to return an instance of itself even when leaning on it's parent class for initialization. For exampe:

```py
class Base:
    def __new__(cls) -> "Base": ...

class Sub(Base):
    def __new__(cls) -> "Sub":
        return super().__new__(cls)  # mypy complains that the function should return `Sub` but it returns `Base`
```

The above is valid code, an instance of `Sub()` would work as you would expect `Sub` to work, and calling `super().__new__` to initialize a subclass is what the [Python docs recommend](https://docs.python.org/3/reference/datamodel.html?highlight=__new__#object.__new__).

The [`Self` type](https://peps.python.org/pep-0673/) is essentially a shortcut for a generic `TypeVar` bounded by the parent class, meaning that a return type of `Self` can be used in a child class without violating any type rules.

quil-py/quil/instructions/__init__.pyi
```diff
@@ -360,73 +361,86 @@ class BinaryLogic:
...
         name: str,
-        parameters: List[Expression],
-        qubits: List[Qubit],
-        instructions: List[Instruction],
-        modifiers: List[GateModifier],
-    ) -> "Calibration": ...
+        parameters: Sequence[Expression],
+        qubits: Sequence[Qubit],
+        instructions: Sequence[Instruction],
+        modifiers: Sequence[GateModifier],
```
MarquessV, line 370: Accepting `Sequence` instead of `List` makes these type hints more flexible. A `List` is a `Sequence`, and `pyo3` converts `Sequence`s to Rust `Vec`s, so this doesn't  betray the reality of the bindings.

kalzoo (approved):

PR #196, opened 2023-04-20
Instruction API: Everything but control flow

MarquessV:


quil-py/src/instruction/arithmetic.rs
MarquessV, line 1: renamed this module to `classical` to better represent the instructions it groups together.

quil-py/src/instruction/classical.rs
```diff
@@ -0,0 +1,473 @@
...
+            _ => py.NotImplemented(),
+        }
+    }
+}
+
+py_wrap_data_struct! {
+    #[derive(Debug, PartialEq, Eq)]
+    #[pyo3(subclass)]
+    PyConvert(Convert) as "Convert" {
+        original: MemoryReference => PyMemoryReference,
```
MarquessV, line 233: This used to be called `from`, but that didn't workout very well because `from` is a Python keyword. Open to other ideas for the naming here.

kalzoo: In general, we probably don't want to to tailor the Rust library to suit Python, but now that I look at it, it seems like `CONVERT` could fix the same convention as other memory-related instructions like `MOVE`:

- `from` -> `source`
- `to` -> `destination`


This is of course a breaking change

kalzoo (changes requested):

quil-rs/src/instruction/classical.rs
```diff
@@ -0,0 +1,268 @@
...
+}
+
+impl fmt::Display for UnaryOperator {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match &self {
+            UnaryOperator::Neg => write!(f, "NEG"),
+            UnaryOperator::Not => write!(f, "NOT"),
+        }
+    }
+}
```
Shadow53, line 268:
```suggestion
#[derive(Copy, Clone, Debug, PartialEq, Eq, strum::Display)]
#[strum(serialize_all = "UPPERCASE")]
pub enum UnaryOperator {
    Neg,
    Not,
}
```

kalzoo (approved):
lgtm -  I agree with @Shadow53 's recommendations to slim down the code

quil-py/quil/instructions/__init__.pyi
```diff
@@ -360,6 +425,124 @@ class BinaryLogic:
...
+    def right(self, right: MemoryReference): ...
+
+@final
+class ComparisonOperand:
+    """
+    A Quil binary operand.
+
+    Variants:
+        ``literal_integer``: An integer literal.
+        ``literal_real``: An floating point literal.
```
jselig-rigetti, line 470:
```suggestion
        ``literal_real``: A floating point literal.
```

quil-py/quil/instructions/__init__.pyi
```diff
@@ -360,6 +425,124 @@ class BinaryLogic:
...
+        memory_reference: MemoryReference,
+    ) -> "ComparisonOperand": ...
+
+@final
+class ComparisonOperator(Enum):
+    Equal = "EQUAL"
+    GreaterThanOrEqual = "GREATERTHANOREQUAL"
+    GreaterThan = "GREATERTHAN"
+    LessThanOrEqual = "GREATERTHANOREQUAL"
+    LessThan = "LESSTHAN"
```
jselig-rigetti, line 509: Should these match the display values from the rust types? e.g, `"GE"` not `"GREATERTHANOREQUAL"`

https://github.com/rigetti/quil-rs/pull/196/files#diff-05aa3659ae4ba942340a9ef40aaf99f892949ce87121df395960513065495d36R218

MarquessV: Definitely, fixed.

Issue #199, opened 2023-05-07
Parse SWAP-PHASES


kalzoo:
`SWAP-PHASES` support was added for the Python bindings with https://github.com/rigetti/quil-rs/pull/193, but it turns out that the instruction was never added to the parser to begin with! Easy fix.

PR #200, opened 2023-05-07
fix: parse SWAP-PHASES

kalzoo:
- Adds a parser for SWAP-PHASES with a test
- Covers the parser with a test case previously limited to the lexer - a "kitchen sink" program. That had a SWAP-PHASES instruction all along, which lexed fine ... as a gate.
- This exposed two failures in parsing that program:
  - `DEFGATE Name as PERMUTATION` incorrectly had a second line. Permutation-type gate definitions have a single-line block. Fixed the test case.
  - `DELAY 0 1` exposed an edge case in parsing delays which is now handled - the `1` should be interpreted as the duration rather than a qubit index, even though the two parsers overlap on integers.
- Fixes `SWAP-PHASES` frame matching by adding `FrameMatchCondition::Or`

Closes https://github.com/rigetti/quil-rs/issues/199

### DIFF
--- a/src/parser/common.rs
+++ b/src/parser/common.rs
@@ -372,7 +372,7 @@ mod describe_skip_newlines_and_comments {
 }
 
 #[cfg(test)]
-mod tests {
+pub mod tests {
     use crate::{
         expression::Expression,
         instruction::MemoryReference,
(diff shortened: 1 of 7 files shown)


src/parser/common.rs
```diff
@@ -384,6 +384,64 @@ mod tests {
 
     use super::{parse_matrix, parse_waveform_invocation};
 
+    /// A Quil program which uses all available instructions.
+    pub const KITCHEN_SINK_QUIL: &str = "DECLARE ro BIT[1]
```
MarquessV, line 388: A+ variable naming here

src/parser/lexer/mod.rs
```diff
@@ -86,6 +86,8 @@ pub enum Command {
     SetScale,
     ShiftFrequency,
     ShiftPhase,
+    #[strum(to_string = "SWAP-PHASES")]
```
MarquessV, line 89: Is this not already the string representation of `SwapPhases`, given the enum uses `SCREAMING-KEBAB-CASE` as the default?

src/parser/instruction.rs
```diff
@@ -814,6 +817,38 @@ mod tests {
...
+                    name: String::from("rf"),
+                    qubits: vec![Qubit::Fixed(1), Qubit::Fixed(2), Qubit::Fixed(3)],
+                },
+                frame_2: FrameIdentifier {
+                    name: String::from("rf"),
+                    qubits: vec![Qubit::Fixed(4), Qubit::Fixed(5), Qubit::Fixed(6)],
+                },
+            }),
+        ];
+        assert_eq!(remainder.len(), 0);
```
Shadow53, line 848: Minor suggestion:

```suggestion
        assert!(remainder.is_empty());
```

src/parser/instruction.rs
```diff
@@ -814,6 +817,38 @@ mod tests {
...
+        assert_eq!(parsed, expected);
```
Shadow53, line 849: Should this come before the remainder assertion? If this ever breaks, I feel like it would be more helpful to first see what _did_ get parsed, rather than just that something wasn't parsed.

src/parser/lexer/mod.rs
Shadow53, line 89: Shouldn't be necessary, since the enum is annotated with `serialize_all = "SCREAMING-KEBAB-CASE"`

Shadow53 (commented):
A few minor suggestions. Don't quite follow what's going on with the `FrameMatchConfition::Or` stuff, but the rest looks fine, barring the suggestion comments.

src/parser/instruction.rs
kalzoo: Isn't it nice to know what the length is if it's not zero?

src/parser/common.rs
kalzoo: It's what I strive for. To be fair I did borrow it from somewhere in quil land

src/parser/instruction.rs
kalzoo: yep, fixed

src/program/frame.rs
```diff
@@ -77,6 +77,14 @@ impl FrameSet {
...
                 .reduce(|acc, el| acc.into_iter().filter(|&v| el.contains(v)).collect())
                 .unwrap_or_default(),
+            FrameMatchCondition::Or(conditions) => conditions
+                .into_iter()
+                .map(|c| self.get_matching_keys(c))
+                .reduce(|mut acc, el| {
+                    acc.extend(el);
+                    acc
+                })
+                .unwrap_or_default(),
```
Shadow53, line 87: I think this can be rewritten as

```suggestion
                .flatten()
                .collect(),
```

Shadow53 (approved):
One more suggestion, but otherwise LGTM

PR #201, opened 2023-05-10
Prepare to release quil-py

MarquessV:
Setting up the last few things needed to properly release this thang.

1. `knope.toml` has been updated to support both the `quil-rs` and `quil-py` packages.
    * `knope` will tag releases as `quil-rs/v*` and `quil-py/v*`, respectively
    * The supported conventional commit scopes are `rs`, `rust`, `quil-rs` for quil-rs, and `py`, `python`, and `quil-py` for `quil-py`. An unscoped commit affects both packages.
2. Created a new release workflow for `quil-py` that build wheels `macos` and `linux`, publishes a `quil` package to `pypi`, and `quil-py` as a Rust crate.
3. Updated miscellaneous details regarding package metadata and READMEs.


knope.toml
```diff
@@ -29,8 +34,7 @@ type = "PrepareRelease"
 
 [[workflows.steps]]
 type = "Command"
-command = "git add Cargo.lock && git commit -m \"chore: prepare release $version [skip ci]\""
-variables = { "$version" = "Version" }
```
MarquessV, line 33: This isn't supported for multi-package repos, so I removed it.

.github/workflows/release-quil-rs.yml
```diff
@@ -0,0 +1,26 @@
...
+on:
+  release:
+    types: [published]
+    
+jobs:
+  is-quil-rs-release:
+    runs-on: ubuntu-latest
+    if: ${{ github.event_name == 'workflow_dispatch' || startswith(github.event.release.tag_name, 'quil-rs/v') }}
+    steps:
+      - run: echo "release tag starts with quil-rs/v, proceeding with release"
```
jselig-rigetti, line 12: What the motivation/expected situation is to handle manual releases - wouldn't manually publishing a release tag give a similar effect? In a similar pyquil MR (see [comment](https://github.com/rigetti/pyquil/pull/1571/files#r1186326866)) I was hoping to align with the dual publish/manual event pattern, but didn't see a clean alternative to relying on `github.event.release.tag_name` in the manual case.

Also, is it intended that this have an `on: workflow_dispatch` section too, based on the `if` condition here?

```suggestion
on:
  release:
    types: [published]
  workflow_dispatch:
    description: "Manually publish release"
    
jobs:
  is-quil-rs-release:
    runs-on: ubuntu-latest
    if: ${{ github.event_name == 'workflow_dispatch' || startswith(github.event.release.tag_name, 'quil-rs/v') }}
    steps:
      - run: echo "release tag starts with quil-rs/v, proceeding with release"
```

knope.toml
```diff
@@ -1,6 +1,12 @@
-[package]
-versioned_files = ["Cargo.toml"]
-changelog = "CHANGELOG.md"
+[packages.quil-rs]
+versioned_files = ["quil-rs/Cargo.toml"]
+changelog = "quil-py/CHANGELOG.md"
```
Shadow53, line 3: Copy-pasta error?

```suggestion
changelog = "quil-rs/CHANGELOG.md"
```

MarquessV: Sure was, good spot.

.github/workflows/release-quil-rs.yml
MarquessV: Yeah, was intending to have a `workflow_dispatch` trigger here, good catch!

As for why it's there, it's mostly an escape hatch when something goes wrong with the release pipeline that is retry-able after knope has already committed version bumps, changelogs, and created the release tag. For example, a couple of weeks ago I released `qcs-sdk-python`, and [publishing to pypi failed on a 504](https://github.com/rigetti/qcs-sdk-rust/actions/runs/4928837074/jobs/8808162101#step:3:41). You are right in that recreating the release tag would work, but you would have to force delete the existing tag for the current release and recreate it, which seems more burdensome and error prone than the `workflow_dispatch` approach.

As for your use case, I think we could call the docker workflow at the end of a successful publish (see: [reusing workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows). That would cover the "happy path" and ensure the docker workflow ran after the package was published. Though, I think we'd also want a `workflow_dispatch` trigger on the docker workflow just in case the PyPI publish succeeds but the docker one fails on some transient error. PyPI returns an error if you try to publish an existing version, so you couldn't just re-trigger the Python release workflow to get the docker one to run after in that case. Assuming Docker also returns an error if you try to publish a duplicate tag, I don't see any harm in having the manual trigger. Does all that make sense? Hopefully I understood your question correctly.

.github/workflows/release-quil-py.yml
```diff
@@ -0,0 +1,109 @@
...
+    - uses: actions/setup-python@v4
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Build wheels
+      uses: messense/maturin-action@v1
+      with:
+        target: ${{ matrix.target }}
+        manylinux: 2014
+        args: -i ${{ matrix.python-version }} --release --manifest-path quil-py/Cargo.toml --out dist
+    - name: Upload wheels
```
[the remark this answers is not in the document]
MarquessV: It's unclear to me how much rigor is required by maturin to build a wheel, and how much validation pypi does against published wheels, but I'm assuming there is a fair amount of validation in those two steps alone, so I'm not sure how necessary the step is to begin with. I felt good about having it as a smoke test, without needing to check it everywhere. Would you like to see as part of the linux workflows?

scripts/sync_versions.sh
```diff
@@ -0,0 +1,12 @@
...
+# Get current version of quil-rs
+PACKAGE_VERSION=$(cargo metadata --format-version 1 --no-deps | grep -o '"name":"quil-rs","version":"[^"]*' | cut -d '"' -f 8)
+
+# Write it to `quil-py`s Cargo.toml
+# `awk -i inplace` would work on linux, but not on macos. This command is compatible with both.
+awk -v ver="$PACKAGE_VERSION" '/^quil-rs = /{$0="quil-rs = { path = \"../quil-rs\", version = \"" ver "\" }"}1' quil-py/Cargo.toml > quil-py/Cargo.toml.tmp
+mv quil-py/Cargo.toml.tmp quil-py/Cargo.toml
+
+# Validate that the new version of Cargo.toml is valid
+cargo check --manifest-path=quil-py/Cargo.toml
```
MarquessV, line 12: This is how I'm planning to keep the version of `quil-rs` specified in `quil-py` in sync.

`knope` will call this script after bumping the version of `quil-rs` to update the version spec in `quil-py`. The `cargo check` here returns an error for safety, just in case the awk command mangles the version number.

These are three cases I'm considering:

## A globally scoped release triggers version bumps of both `quil-rs` and `quil-py`

Everything gets version bumped, release tags get created, and the release workflows for both kick off.

## A `quil-py` scoped release triggers a version bump of just `quil-py`

Here, the script is effectively a no-op, this script will run and overwrite the spec with the same version, and the release will proceed as normal.

## A `quil-rs` scoped release triggers a version bump of just `quil-rs`

Here, `quil-rs` gets bumped and so does it's version in `quil-py`'s manifest, but the latter doesn't get released. I don't see this as a feature, since if the release is correctly scoped, the new release of `quil-rs` shouldn't need to trigger a new release of `quil-py`. If it will affect `quil-py`, the release should be scoped to both.

.github/workflows/release-quil-py.yml
```diff
@@ -0,0 +1,109 @@
...
+
+jobs:
+  is-python-release:
+    if: ${{ github.event_name == 'workflow_dispatch' || startsWith(github.event.release.tag_name, 'quil-py/v') }}
+    runs-on: ubuntu-latest
+    steps:
+      - run: echo "Release tag starts with quil-py/v, proceeding with release"
+
+  macos:
+    runs-on: macos-11
```
Shadow53, line 18: Any specific reason for `macos-11` over `macos-latest`, considering the use of `ubuntu-latest`?

.github/workflows/release-quil-py.yml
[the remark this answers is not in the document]
Shadow53: Not sure what would be different between a macos universal wheel vs a non-manylinux linux wheel, but it _could_ help catch something 🤷🏻‍♂️. Just seemed odd to do it on one system and not another.

.github/workflows/release-quil-py.yml
```diff
@@ -0,0 +1,109 @@
...
+      - name: Publish to PyPi
+        env:
+          MATURIN_USERNAME: ${{ secrets.PYPI_USERNAME }}
+          MATURIN_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
+        uses: messense/maturin-action@v1
+        with:
+          command: upload 
+          args: --skip-existing wheels/*
+
+  publish-rust-crate:
```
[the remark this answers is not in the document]
Shadow53: Why not merge the release logic for both into the same file and make releasing `quil-py` depend on the non-failure of `quil-rs`?

.github/workflows/release-quil-py.yml
```diff
@@ -0,0 +1,109 @@
...
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+          token: ${{ secrets.PAT }}
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          override: true
+      - run: cargo publish --manifest-path=quil-rs/Cargo.toml --token ${{ secrets.CRATES_IO_TOKEN }}
```
Shadow53, line 109: I think this is publishing the rust library again instead of the pyo3 one

```suggestion
      - run: cargo publish --manifest-path=quil-py/Cargo.toml --token ${{ secrets.CRATES_IO_TOKEN }}
```

Shadow53 (approved):
A couple minor things and a couple answers to questions. This looks good overall, though.

.github/workflows/release-quil-py.yml
MarquessV: Nice catch, fixed.

.github/workflows/release-quil-py.yml
MarquessV: Pinned them to `macos-12` and `ubuntu-22.04`

.github/workflows/release-quil-py.yml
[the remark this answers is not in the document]
MarquessV: I ended up adding it for `x86_64`, but not `aarch64`. The latter is built using cross compilation, so the host `x86_64` can't install it. I'm sure there's ways around it, but I don't think it's worth the squeeze.

PR #205, opened 2023-05-19
fix: The imaginary part of a complex number will now always be formatted as a floating point number.

### LINKED ISSUE #204: The imaginary part of a complex number should always be printed as a floating point number
The `format_complex` function trims trailing `.0` off of all floats, but this results in invalid Quil expressions for the imaginary part of a complex number.

Per the quil spec, if `i` is followed by a number, it has to be a floating point number:

image
("Real" is defined as a floating point number in the quil spec)

Original [discussion](https://github.com/rigetti/qcs-sdk-rust/issues/294#issuecomment-1536917461)

MarquessV:
closes #204

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -19,7 +19,7 @@ vulnerability = "deny"
 # The lint level for unmaintained crates
 unmaintained = "deny"
 # The lint level for crates that have been yanked from their source registry
-yanked = "deny"
+yanked = "warn"
 # The lint level for crates with security notices.
 notice = "deny"
 # A list of advisory IDs to ignore. Note that ignored advisories will still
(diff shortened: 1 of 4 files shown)


deny.toml
```diff
@@ -19,7 +19,7 @@ vulnerability = "deny"
 # The lint level for unmaintained crates
 unmaintained = "deny"
 # The lint level for crates that have been yanked from their source registry
-yanked = "deny"
+yanked = "warn"
```
MarquessV, line 22: Downgrading, but this is the default value IIRC.

Either way, `crossbeam-channel` yanked a version of their crate, which is a dependency of `criterion`, so we can't do anything about it until they upgrade.

jselig-rigetti (approved):

Shadow53 (approved):

PR #207, opened 2023-05-23
fix: The parser now follows the correct precedence rules for ungrouped infix expressions

### LINKED ISSUE #206: Infix expressions with ungrouped infix operands are parsed incorrectly
Infix expressions that have left and right infix operands, without any parentheses to explicitly express precedence are parsed incorrectly. For example, the expression:

```
pi/2 + 2*theta[0]
```

Is parsed as if it were:

```
pi / (2 + 2*theta[0])
```

MarquessV:
closes #206

### DIFF
--- a/src/expression.rs
+++ b/src/expression.rs
@@ -545,6 +545,32 @@ fn format_inner_expression(f: &mut fmt::Formatter, expression: &Expression) -> f
     }
 }
 
+#[cfg(test)]
+mod test {
+    use crate::{
+        expression::{Expression, InfixOperator, PrefixOperator},
+        real,
+    };
+
+    #[test]
+    fn formats_nested_expression() {
+        let expression = Expression::Infix {
+            left: Box::new(Expression::Prefix {
+                operator: PrefixOperator::Minus,
+                expression: Box::new(Expression::Number(real!(3f64))),
+            }),
+            operator: InfixOperator::Star,
+            right: Box::new(Expression::Infix {
+                left: Box::new(Expression::PiConstant),
+                operator: InfixOperator::Slash,
+                right: Box::new(Expression::Number(real!(2f64))),
+            }),
+        };
+
+        assert_eq!(expression.to_string(), "-3*(pi/2)");
+    }
+}
+
 /// A function defined within Quil syntax.
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
 #[cfg_attr(test, derive(Arbitrary))]
--- a/src/expression.rs
+++ b/src/expression.rs
@@ -513,15 +513,35 @@ impl fmt::Display for Expression {
                 left,
                 operator,
                 right,
-            } => write!(f, "({}{}{})", left, operator, right),
+            } => {
+                format_inner_expression(f, left)?;
+                write!(f, "{}", operator)?;
+                format_inner_expression(f, right)
+            }
             Number(value) => write!(f, "{}", format_complex(value)),
             PiConstant => write!(f, "pi"),
             Prefix {
                 operator,
                 expression,
-            } => write!(f, "({}{})", operator, expression),
+            } => {
+                write!(f, "{}", operator)?;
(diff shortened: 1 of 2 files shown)


src/expression.rs
```diff
@@ -513,18 +513,64 @@ impl fmt::Display for Expression {
...
+            right: Box::new(Expression::Infix {
+                left: Box::new(Expression::PiConstant),
+                operator: InfixOperator::Slash,
+                right: Box::new(Expression::Number(real!(2f64))),
+            }),
+        };
+
+        assert_eq!(expression.to_string(), "-3*(pi/2)");
+    }
+}
```
MarquessV, line 572: This is somewhat tangential to the parser fix, but I figured I'd take a stab at simplifying expression formatting here. Now, only nested infix expression will have surrounding parentheses applied. It is a lot less aggressive, and I think makes the resulting Quil easier to read. It probably would have made issues like this easier to spot.

A few more examples:

Before:
```
RZ((-pi)) 0
RZ(((-pi)/2)) 0
```

After:
```
RZ(-pi) 0
RZ(-pi/2) 0
```

Shadow53: I get the feeling this may not work well with a "minus negative" expression like `2 - (-3)`. Do we have a set of tests that verify the written expression is parsed as itself, and is a "minus negative" one of the test cases?

src/parser/expression.rs
```diff
@@ -38,17 +38,24 @@ enum Precedence {
...
             Token::LParenthesis => Precedence::Call,
             _ => Precedence::Lowest,
         }
     }
 }
 
+impl From<&Operator> for Precedence {
+    fn from(operator: &Operator) -> Self {
+        match operator {
+            Operator::Plus | Operator::Minus => Precedence::Sum,
```
Shadow53, line 52: Does quil also support one or both of these as unary operators. e.g. `+2`, `-4`?

src/parser/expression.rs
```diff
@@ -200,7 +207,7 @@ fn parse_infix(input: ParserInput, left: Expression) -> InternalParserResult<Exp
                 Operator::Slash => InfixOperator::Slash,
                 Operator::Star => InfixOperator::Star,
             };
-            let precedence = get_precedence(remainder);
```
MarquessV, line 203: This was the crux of the issue. At this point `remainder` is the input just _after_ the operator, so `get_precedence` wasn't actually matching against the right token and  always fell through to the default case, which is `Precedence::Lowest`.

The fix instead gets precedence using the operator token directly.

Shadow53 (commented):
The changes seem good for resolving this issue. I'd like more checks that the expression simplification isn't breaking anything, though

src/expression.rs
MarquessV: Functionally it is fine.

Something like:
```
-3 - -2
```

will get formatted:
```
-3--2
```

Both parse and run correctly (at least on the QVM). It isn't the prettiest, but I'd rather have this edge case then all the extra parentheses we had before. As a compromise, we could wrap nested prefix expressions in parentheses too.

src/parser/expression.rs
MarquessV: Yes, `+` and `-` are both valid unary operators in quil. Are you getting at something with that question?

src/expression.rs
MarquessV: To start, I added test cases to the expression parser for the "minus negative" case.

src/parser/expression.rs
Shadow53: I was wondering if there needed to be a separate case for `+`/`-` as unary operators, since they would need higher precedence than `*`/`/`. Though thinking about it now, I suppose the lexer could/should be handling that case, not the parser.

MarquessV: Yeah, I think that makes more sense to handle with the lexer. Either way, tangential to the fix at hand, so I'll resolve.

PR #209, opened 2023-05-24
feat: improve Expression::simplify

### LINKED ISSUE #208: Improve Expression Simplification
Currently, Expression::simplify can collapse certain expressions to simpler forms, but it's fairly limited. For example, the following expressions would be untouched:

```
0 * theta
theta / 1
(theta * 5) / 5
```

This would be useful for preprocessing of evaluations for execution on an embedded control system.

kalzoo:
Closes #208 

This should be considered proof-of-concept quality at this point.

Ideally, we'd be able to use a library to implement expression simplification; this is reinventing an oft-reinvented wheel. However, we do need some minimal utility here for a particular use case, and so I don't want perfect to be the enemy of minimally good.

Questions for reviewers:

* Is there a library you'd recommend for use here?
* Are there clever patterns we can put to work?
* How can we prevent unbounded recursion? At a minimum, we could pass a depth argument or could recognize when simplification is recursing without modification

### DIFF
--- a/src/expression.rs
+++ b/src/expression.rs
@@ -249,11 +249,75 @@ impl Expression {
             }
             Infix {
                 left,
-                operator: _,
+                operator,
                 right,
             } => {
                 left.simplify();
                 right.simplify();
+
+                let new = match (left.as_ref(), right.as_ref()) {
+                    (left, Number(right)) if right == &Complex64::from(0f64) => {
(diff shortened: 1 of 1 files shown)


kalzoo:
Superseded by https://github.com/rigetti/quil-rs/pull/210

PR #210, opened 2023-05-25
Simplify eggspressions

genos:
This ~is a DRAFT of using~ uses [`egg`](https://docs.rs/egg/latest/egg/index.html) to tackle #208.

TODO:
- [x] Incorporate `egg` in a private module to do `Expression::simplify`
- [x] Pass current test suite + more
- [x] Handle the `"(neg <number>)"` edge case in simplification, not with `if let Ok(number) = …` workaround†
- [x] Actual error handling, no `panic!` or `expect!`
- [x] Documentation

†~I haven't figured out how to manage this one; any help from reviewers would be much appreciated.~ Fixed in d4f7fbf


src/expression.rs
```diff
@@ -219,6 +219,517 @@ fn is_small(x: f64) -> bool {
...
+            .expect("Expression->Sexp->RecExpr should work by design");
+        let runner = egg::Runner::default()
+            .with_expr(&recexpr)
+            .run(&make_rules());
+        let root = runner.roots[0];
+        let (_, best) = egg::Extractor::new(&runner.egraph, egg::AstSize).find_best(root);
+        let simpler_sexp = parser::parse_str(&best.to_string())
+            .expect("RecExpr implements to_sexp, so this parse should succeed");
+        let mut simpler_exp = Expression::from(&simpler_sexp);
+        if let Ok(number) = simpler_exp.evaluate(&Default::default(), &Default::default()) {
```
genos, line 249: For some reason, the rewrite rules don't cover turning, e.g. `(neg 1)` into `-1`. Trying to take care of this without the special case.

src/expression.rs
```diff
@@ -219,6 +219,517 @@ fn is_small(x: f64) -> bool {
...
+    impl From<&Sexp> for Expression {
+        fn from(sexp: &Sexp) -> Self {
+            match sexp {
+                Sexp::Empty => panic!("Can't make an expression from an empty sexp"),
+                Sexp::String(s) => {
+                    match Expression::from_str(s)
+                        .expect("Strings encode valid Expressions by design")
+                    {
+                        // corner case, we're missing the prefix %
+                        Expression::Address(MemoryReference { name, .. }) => {
```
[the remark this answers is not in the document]
kalzoo: Nope, looks like a `quil-rs` bug to me: https://quil-lang.github.io/#3-2Syntactic-Rudiments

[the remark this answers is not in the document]
MarquessV: Yup, that's a bug. I fixed it [here](https://github.com/rigetti/quil-rs/pull/175), but it's tied up in a feature branch. Should land in main soon.

kalzoo (commented):
This is rad, @genos !! Sure, it is a little funny to simplify via stringification, but it's _a lot_ cleaner than having to rewrite all that logic using the enums directly. Two things:

* I'd be interested in simple benchmarks to track performance. I'm not too worried about regression here but would be good to know, given how often it's invoked elsewhere.
* (see other comment)

genos:
Thanks, @kalzoo! Re: 

> * I'd be interested in simple benchmarks to track performance. I'm not too worried about regression here but would be good to know, given how often it's invoked elsewhere.

What did you have in mind? A `criterion` bench with a randomly-generated-of-a-fixed-sized `Expression` to simplify?

src/expression.rs
```diff
@@ -235,47 +769,8 @@ impl Expression {
...
-                    std::mem::swap(self, &mut temp);
-                }
-            }
-            Variable(_) | Address(_) | PiConstant | Number(_) => {}
-        };
-
-        if let Ok(number) = self.evaluate(&HashMap::new(), &HashMap::new()) {
-            *self = Number(number);
+        if let Ok(simpler) = simplification::run(self) {
+            *self = simpler;
```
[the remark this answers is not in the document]
genos: Should we include memory references?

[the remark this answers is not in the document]
genos: See 2a5e7c3

kalzoo:
> Thanks, @kalzoo! Re:
> 
> > * I'd be interested in simple benchmarks to track performance. I'm not too worried about regression here but would be good to know, given how often it's invoked elsewhere.
> 
> What did you have in mind? A `criterion` bench with a randomly-generated-of-a-fixed-sized `Expression` to simplify?

That could work, so long as it's only randomly generated once and then committed - maybe a few of them for good measure. We'd want benchmark inputs to be static so that they can provide meaningful feedback to development.

src/expression.rs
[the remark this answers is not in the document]
kalzoo: I thought this sounded familiar...

genos:
> Sure, it is a little funny to simplify via stringification,

This would be easier if `egg` exposed some of the internals of `RecExpr`, specifically [parsing from `Sexp`s](https://docs.rs/egg/latest/src/egg/language.rs.html#551). I think I'll steal this, actually, rather than rely on the stringification.

src/expression.rs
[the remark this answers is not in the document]
genos: Turned out this one was on me, I needed to include the `%` when turning a variable into a symbolic expression; see b5b0753.

genos:
@kalzoo re: generating random expressions to benchmark simplifying, I've tried adding `rand` as a dependency and the following `main.rs` to generate a static file of them, but it looks like the `Expression ⇒ String ⇒ Expression` roundtrip doesn't quite work:

```
thread 'main' panicked at
'pi/cis(sin(sin(-%owlhhqxriv^cos(%mghfxnbixn)))) => error while parsing:
at line 1, column 28 (OPERATOR(^)): expected RParenthesis, found OPERATOR(^)', src/main.rs:83:23
```

<details>
<summary>main.rs </summary>

```rust
use quil_rs::{
    expression::{Expression, ExpressionFunction, InfixOperator, PrefixOperator},
    instruction::MemoryReference,
};
use rand::{rngs::StdRng, Rng, SeedableRng};
use std::{fs::File, io::Write, str::FromStr};

// rand::dists::Alphanumeric isn't what we need
fn rand_string(rng: &mut impl Rng) -> String {
    (0..10).map(|_| rng.gen_range('a'..'z')).collect()
}

fn rand_expr(rng: &mut impl Rng, max_depth: usize) -> Expression {
    if max_depth == 0 {
        Expression::PiConstant
    } else {
        match rng.gen_range(0..7) {
            0 => {
                let name = rand_string(rng);
                let index = rng.gen();
                Expression::Address(MemoryReference { name, index })
            }
            1 => {
                let function = match rng.gen_range(0..5) {
                    0 => ExpressionFunction::Cis,
                    1 => ExpressionFunction::Cosine,
                    2 => ExpressionFunction::Exponent,
                    3 => ExpressionFunction::Sine,
                    _ => ExpressionFunction::SquareRoot,
                };
                let expression = Box::new(rand_expr(rng, max_depth / 2));
                Expression::FunctionCall {
                    function,
                    expression,
                }
            }
            2 => {
                let left = Box::new(rand_expr(rng, max_depth / 2));
                let operator = match rng.gen_range(0..5) {
                    0 => InfixOperator::Caret,
                    1 => InfixOperator::Plus,
                    2 => InfixOperator::Minus,
                    3 => InfixOperator::Slash,
                    _ => InfixOperator::Star,
                };
                let right = Box::new(rand_expr(rng, max_depth / 2));
                Expression::Infix {
                    left,
                    operator,
                    right,
                }
            }
            3 => {
                let re = rng.gen::<f64>();
                let im = rng.gen::<f64>();
                Expression::Number(num_complex::Complex64 { re, im })
            }
            4 => Expression::PiConstant,
            5 => {
                let operator = if rng.gen() {
                    PrefixOperator::Plus
                } else {
                    PrefixOperator::Minus
                };
                let expression = Box::new(rand_expr(rng, max_depth / 2));
                Expression::Prefix {
                    operator,
                    expression,
                }
            }
            _ => Expression::Variable(rand_string(rng)),
        }
    }
}

fn main() {
    let mut rng = StdRng::seed_from_u64(8675309);
    let mut out = File::create("expressions.txt").expect("We should be able to create a file.");
    for _ in 0..1_000 {
        let expression = rand_expr(&mut rng, 256);
        match Expression::from_str(&expression.to_string()) {
            Ok(_) => write!(out, "{}", expression).expect("We should be able to write to the file"),
            Err(e) => panic!("{expression} => {e}"),
        }
    }
}
```
</details>

genos:
Further `proptest` play lead to another expression that fails to round trip: `"cis(AA[0]^a0[428])"`

genos:
Just merged in `master` in 7d9f1d5.

quil-rs/src/expression.rs
```diff
@@ -253,6 +253,537 @@ fn is_small(x: f64) -> bool {
...
+    pub(super) fn run(expression: &Expression) -> Result<Expression, SimplificationError> {
+        let sexp = expression.into_sexp();
+        let recexpr = parse_expr(&sexp)?;
+        let runner = egg::Runner::default()
+            .with_expr(&recexpr)
+            .run(&make_rules());
+        let root = runner.roots[0];
+        let (_, best) = egg::Extractor::new(&runner.egraph, egg::AstSize).find_best(root);
+        let simpler_sexp = parser::parse_str(&best.to_string())
+            .map_err(|e| SimplificationError::SexpFromRecExpr(format!("{e:#?}")))?;
```
Shadow53, line 284: Is there a specific reason for stringifying the error instead of including the error object?

quil-rs/src/expression.rs
```diff
@@ -253,6 +253,537 @@ fn is_small(x: f64) -> bool {
...
+    ///
+    /// Fun fact, there is no total ordering on the complex numbers; however, the derived thing
+    /// here will work for our purposes.
+    ///
+    /// https://en.wikipedia.org/wiki/Complex_number#Ordering
+    #[derive(Debug, Default, PartialEq, Eq, PartialOrd, Ord, Clone, Copy, Hash)]
+    struct C {
+        re: OrderedFloat<f64>,
+        im: OrderedFloat<f64>,
+    }
```
Shadow53, line 443: Would prefer a different name for this -- I looked at `impl C` below and got confused at why you were trying to add C bindings (mistook it for `extern "C"`)

quil-rs/src/expression.rs
```diff
@@ -253,6 +253,537 @@ fn is_small(x: f64) -> bool {
...
+            })
+        }
+
+        /// Update the graph to equate and simplify constant values.
+        fn modify(egraph: &mut EGraph, id: Id) {
+            if let Some(c) = egraph[id].data {
+                let value = match c {
+                    _ if c.close(C::PI) => Expr::Pi,
+                    _ => Expr::Number(c),
+                };
```
Shadow53, line 666: This should be the same logic, unless there's something I'm not understanding here

```suggestion
                let value = if c.close(C::PI) {
                    Expr::Pi
                } else {
                    Expr::Number(c)
                };
```

quil-rs/src/expression.rs
```diff
@@ -936,46 +1440,46 @@ mod tests {
         #[test]
         fn exponentiation_works_as_expected(left in arb_expr(), right in arb_expr()) {
             let expected = Expression::Infix (InfixExpression { left: Box::new(left.clone()), operator: InfixOperator::Caret, right: Box::new(right.clone()) } );
-            prop_assert_eq!(left.clone() ^ right.clone(), expected.clone());
+            prop_assert_eq!(&(left.clone() ^ right.clone()), &expected);
```
Shadow53, line 1443: Should be able to do this, right? I imagine the macro would automatically take the reference to each argument.

```suggestion
            prop_assert_eq!(left.clone() ^ right.clone(), expected);
```

Shadow53 (commented):
Some minor comments, mostly style. Biggest thing is wanting to avoid stringification of errors. If we're not wanting to expose an `egg` error in the public API, I'd prefer a newtype wrapper over strings.

quil-rs/src/expression.rs
genos: I thought so too, but:
1. 1.70 `clippy` yelled about me about needless `.clone` calls, but
2. removing them made tests not compile

quil-rs/src/expression.rs
genos: I wasn't sure about bubbling up a `SexpError` from `symbolic_expressions` by way of `egg`. If we don't mind that popping up, I'll change this to use `#[from] SexpError`.

quil-rs/src/expression.rs
genos: I'll split into two tests apiece, avoiding this issue.

genos:
Adding because the two are intertwined:

Closes #215

genos:
The latest issue is a mix of operator precedence and formatting/parsing complex numbers. I think, essentially, that `x^1-2i` is _supposed_ to be parsed as `X ^ (1-2i)`, but is instead being parsed as `(X^1) - (2i)`.

Shadow53:
> The latest issue is a mix of operator precedence and formatting/parsing complex numbers. I think, essentially, that `x^1-2i` is _supposed_ to be parsed as `X ^ (1-2i)`, but is instead being parsed as `(X^1) - (2i)`.

This appears to be a bug in the parser. The [Quil grammar](https://github.com/quil-lang/quil/blob/master/grammars/Quil.g4) specifies that `^`/`POWER` is right-associative, so I think you are right that `X ^ (1-2i)` is the correct parsing

genos:
> > The latest issue is a mix of operator precedence and formatting/parsing complex numbers. I think, essentially, that `x^1-2i` is _supposed_ to be parsed as `X ^ (1-2i)`, but is instead being parsed as `(X^1) - (2i)`.
> 
> This appears to be a bug in the parser. The [Quil grammar](https://github.com/quil-lang/quil/blob/master/grammars/Quil.g4) specifies that `^`/`POWER` is right-associative, so I think you are right that `X ^ (1-2i)` is the correct parsing

Thanks @Shadow53! With local testing, I may have moved past this to a different and new error, but I’ll write a more concrete test case when next at my computer.

genos:
Ok! This now passes all my local testing and is about where I think it should be. Pinging folks for re-reviews.

In order to get round-tripping to go (#215), I've added the `fn parenthesized(expression: &Expression) -> String` in the testing module. Unused elsewhere, but it ensures unambiguous behavior for our tests.

genos:
Even simpler now; I removed the `ordered-float` dependency that I originally added, plus made the custom `Complex` type just a wrapper for `num_complex::Complex64` implementing the various traits we need.

quil-rs/src/expression.rs
```diff
@@ -253,6 +254,592 @@ fn is_small(x: f64) -> bool {
...
+                Expr::Add([left, right]) => Some(*x(left)? + *x(right)?),
+                Expr::Sub([left, right]) => Some(*x(left)? - *x(right)?),
+                Expr::Address(_) | Expr::Symbol(_) => None,
+            }
+        }
+
+        /// Merge two pieces of data with the same value.
+        fn merge(&mut self, to: &mut Self::Data, from: Self::Data) -> egg::DidMerge {
+            egg::merge_option(to, from, |a, b| {
+                assert!(a.close(b), "Merged non-equal constants: {a} & {b}");
```
kalzoo, line 673: Can we avoid this panic?

quil-rs/src/expression.rs
```diff
@@ -253,6 +254,592 @@ fn is_small(x: f64) -> bool {
...
+    }
+
+    /// Rewrite terms of our [`Expr`] language by reducing with our [`Arithmetic`] analysis.
+    type Rewrite = egg::Rewrite<Expr, Arithmetic>;
+
+    /// Instantiate our rewrite rules for simplifying [`Expr`] terms.
+    static RULES: Lazy<Vec<Rewrite>> = Lazy::new(|| {
+        vec![
+            // largely copied from https://github.com/egraphs-good/egg/blob/main/tests/math.rs
+            // and https://github.com/herbie-fp/herbie/blob/main/egg-herbie/src/rules.rs
```
kalzoo, line 712:
```suggestion
            // largely copied from https://github.com/egraphs-good/egg/blob/82c00e970f0bc1fbfe90ce6dc3c3c79ee919c933/tests/math.rs
            // and https://github.com/herbie-fp/herbie/blob/2052806f2ffe0d46bc2e151dd096b127c39e12bd/egg-herbie/src/rules.rs
```

permalinks

quil-rs/src/expression.rs
```diff
@@ -675,7 +1232,8 @@ impl fmt::Display for InfixOperator {
             match self {
                 Caret => "^",
                 Plus => "+",
-                Minus => "-",
+                // NOTE: spaces included to distinguish from hyphenated identifiers
+                Minus => " - ",
```
kalzoo, line 1236: 👍

quil-rs/src/expression.rs
```diff
@@ -649,7 +1205,8 @@ impl fmt::Display for PrefixOperator {
             f,
             "{}",
             match self {
-                Plus => "+",
+                // NOTE: prefix Plus does nothing, and it causes parsing issues
+                Plus => "",
```
kalzoo, line 1209: 👍

kalzoo (commented):
Small comments - I think it's just about good to go! We can add benchmarks with a follow-on PR - it'll be good to have some focused on simplification.

quil-rs/src/expression.rs
genos: Sure! The only times it's been triggered in loads of randomized testing is when dealing with `NaN` and `inf`; those worries are currently avoided with tighter bounds on what floats are allowed in the property-based testing.

kalzoo:
good simplification in https://github.com/rigetti/quil-rs/pull/210/commits/c3f363f40c3be2572631db43b2c037197f3a87a6, @genos 

can you catch up to `main` and then we can merge?

genos:
> good simplification in [c3f363f](https://github.com/rigetti/quil-rs/commit/c3f363f40c3be2572631db43b2c037197f3a87a6), @genos
> 
> can you catch up to `main` and then we can merge?

Thanks @kalzoo! Fixed merge conflicts with `main`, but the `Deny` task is still failing. It looks like `proptest` and `criterion` are pulling two different versions of `regex_syntax`. Any thoughts on fixing that?

genos:
Perhaps related: we still have [`Cargo.lock` in our `.gitignore`](https://github.com/rigetti/quil-rs/blob/main/.gitignore#L3), but it's [in the repo](https://github.com/rigetti/quil-rs/blob/main/Cargo.lock).

kalzoo:
>  the Deny task is still failing. It looks like proptest and criterion are pulling two different versions of regex_syntax. Any thoughts on fixing that?

I'm not too worried about that, both of the warnings are safely ignorable at the moment, even if they add noise to CI results. We can fix them off `main`.

> Perhaps related: we still have [`Cargo.lock` in our `.gitignore`](https://github.com/rigetti/quil-rs/blob/main/.gitignore#L3), but it's [in the repo](https://github.com/rigetti/quil-rs/blob/main/Cargo.lock).

Ah good point, can you `rm` that here?

kalzoo (approved):

genos:
🚨 Squashing and merging 🚨

kalzoo:
> 🚨 Squashing and merging 🚨

Thanks for your great work on this @genos ! This is both more thorough and more maintainable than we would have made any simplification rules by hand. And now we're using `egg` in production!

PR #212, opened 2023-06-03
refactor: performance optimizations

Shadow53:
This PR is an in-progress attempt at improving the performance of `quil-rs` and a performances sensitive internal product. Benchmarks are run against both the crate itself and the aforementioned internal product.

As of [the latest commit](https://github.com/rigetti/quil-rs/commit/e69f4e3b970197b648da26025db081a7abadb9ff), benchmarked performance in `quil-rs` is roughly even with `main` (possibly slight improvement, but within noise levels), but consumers are seeing a performance improvement of roughly 7% or 2-3/35ms.


kalzoo:
looking good so far 👍

kalzoo (approved):
Looks good!

Make sure that the merged commit message indicates that this is a breaking change due to the return type change - I don't see that in the current commits.

PR #213, opened 2023-06-06
feat: build and return the unitary of a program

### LINKED ISSUE #187: Add a method to build and return the unitary of a program
A helpful utility that exists in `pyQuil` is the [program_unitary](https://github.com/rigetti/pyquil/blob/6ad688b366813104c2f31c37873b03cc1da1cc8f/pyquil/simulation/tools.py#L325), which builds the unitary of a program. This is done by taking the unitary of each `Gate` in the program. As such, this is only valid for programs consisting of only `Gate`s.

The pyQuil implementation should serve as a reference implementation, with one key difference:

pyQuil supports most standard gate definitions, in addition to a specialized gate not defined in the Quil spec by mapping the name of a gate to it's matrix specification. The `quil-rs` implementation should support _all_ the standard Quil gates, but not maintain anything special for Gates not defined in the Quil spec. As a start, we can just drop support for these gates, but it would be nice to support any custom gate by using `DefGate` instructions in the program.

genos:
closes #187


genos:
8702c98 adds the BSD-2 license to `deny.toml`; with that, CI is 🟢 🎉

genos:
Pinging reviewers; I'd hate for this to get stale with respect to `main` 🙏

quil-rs/src/instruction/gate.rs
```diff
@@ -138,6 +592,216 @@ impl fmt::Display for GateModifier {
     }
 }
 
+#[cfg(test)]
+/// Kind of like numpy.testing.assert_allclose for arrays of complex numbers.
+/// Loops by hand to print out position & differences and to avoid needing, e.g., ordered-float
+macro_rules! assert_allclose {
```
MarquessV, line 598: Is it possible to use `ndarray`s [`abs_diff_eq`](https://docs.rs/ndarray/latest/ndarray/struct.ArrayBase.html#method.abs_diff_eq) for this?

MarquessV (approved):
Awesome! A few minor comments, but nothing to hold up approval. I really appreciate all the tests!

quil-rs/src/instruction/gate.rs
genos: Nice find! If you don't mind me adding the `approx` crate in, I'll use that instead.

genos:
> I really appreciate all the tests!

Only possible thanks to the extensive testing in `pyquil`, I just copy-paste-modified for the most part 😄

quil-rs/src/instruction/gate.rs
MarquessV: I don't mind that at all!

MarquessV:
@genos just checking in - need anything else from me? You're latest changes look good, so my ✅ still stands.

genos:
@MarquessV thanks! I wasn't sure I could merge with the `change requested` hanging over me. 🚨 pushing the button! 🚨

PR #218, opened 2023-06-09
refactor: optimize terminate_working_block

Shadow53:
Another round of performance improvements!

This one is focused primarily on improving the performance of `ScheduledProgram::from_program` and `terminate_working_block`, a function that it calls.

Existing benchmarks `parser` and `get_frames_for_instruction` are largely unchanged with a little bit of noise in either direction. The new benchmark for `ScheduledProgram::from_program` shows no significant performance change at worst and a 16%/1.2µs improvement (from 7.4 to 6.2).

Further, an internal project shows a similar case of no significant change at worst and an improvement of 6.7%/55µs (from 814.24 to 759.69)

### DIFF
--- a/quil-rs/benches/corpus.rs
+++ b/quil-rs/benches/corpus.rs
@@ -16,7 +16,7 @@ fn bench_config_from_file(path: &Path) -> Option<QuilBenchConfig> {
         .file_name()
         .expect("path should have file name component")
         .to_str()
-        .expect("bad filename")
+        .expect("filename should be valid")
         .to_string();
 
     if quil_rs::Program::from_str(&program).is_ok() {
--- a/quil-rs/benches/corpus.rs
+++ b/quil-rs/benches/corpus.rs
@@ -1,4 +1,9 @@
-use std::{fs, path::{Path, PathBuf}, process::Command, str::FromStr};
+use std::{
+    fs,
+    path::{Path, PathBuf},
+    process::Command,
+    str::FromStr,
+};
 
 pub struct QuilBenchConfig {
     pub name: String,
@@ -10,8 +15,7 @@ fn bench_config_from_file(path: &Path) -> Option<QuilBenchConfig> {
         return None;
     }
 
-    let program =
-        fs::read_to_string(path).expect("failed to read quil program file");
+    let program = fs::read_to_string(path).expect("failed to read quil program file");
     let name = path
         .file_name()
         .expect("path should have file name component")
@@ -39,12 +43,9 @@ pub fn from_corpus() -> Vec<QuilBenchConfig> {
     let dir = fs::read_dir(corpus_dir).expect("failed to locate quil corpus directory");
 
     dir.filter_map(Result::ok)
-        .filter_map(|entry| {
-            bench_config_from_file(&entry.path())
-        })
-        .chain({
-            bench_config_from_file(&PathBuf::from(SAMPLE_CALIBRATIONS))
-        }).collect()
+        .filter_map(|entry| bench_config_from_file(&entry.path()))
+        .chain({ bench_config_from_file(&PathBuf::from(SAMPLE_CALIBRATIONS)) })
+        .collect()
 }
 
 // in the event someone wants to run the benchmarks locally, this will download the corpus of quil used
--- a/quil-rs/benches/corpus.rs
+++ b/quil-rs/benches/corpus.rs
@@ -1,15 +1,36 @@
-use std::{fs, path::Path, process::Command, str::FromStr};
+use std::{fs, path::{Path, PathBuf}, process::Command, str::FromStr};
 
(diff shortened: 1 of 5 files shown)


quil-rs/src/program/graph.rs
```diff
@@ -176,18 +176,21 @@ impl MemoryAccessQueue {
...
             }
             None => {
                 let mut edge = HashSet::new();
-                edge.insert($dependency);
-                $graph.add_edge($source.clone(), $target.clone(), edge);
+                edge.insert(dependency);
+                $graph.add_edge(source, target, edge);
             }
         }
-    };
```
Shadow53, line 190: This removes cloning of values, which is unnecessary in the places where this macro is used (values are `Copy`). This clarifies the macro and helps ensure we don't accidentally introduce expensive clones if we change the types used.

quil-rs/src/program/graph.rs
```diff
@@ -516,22 +519,19 @@ fn terminate_working_block(
         return Ok(());
     }
 
-    let block = InstructionBlock::build(working_instructions.to_vec(), terminator, program)?;
+    // This leaves working_instructions and working_label in their respective "empty" states.
+    let block = InstructionBlock::build(std::mem::take(working_instructions), terminator, program)?;
     let label = working_label
-        .clone()
+        .take()
```
Shadow53, line 525: `working_label` was previously set to `None` after processing. This avoids cloning and sets it to `None` at the same time.

This works because any error returned by this function is bubbled up past the point where invalid state could happen.

quil-rs/src/program/graph.rs
```diff
@@ -516,22 +519,19 @@ fn terminate_working_block(
...
+        return Err(ScheduleError {
             instruction_index,
             instruction: Instruction::Label(Label(label)),
             variant: ScheduleErrorVariant::DuplicateLabel,
-        }),
-        None => Ok(()),
-    }?;
-
-    working_instructions.drain(..);
-    *working_label = None;
```
Shadow53, line 534: Simplified the logic of this `match` statement while I was here.

quil-rs/src/program/graph.rs
```diff
@@ -516,22 +519,19 @@ fn terminate_working_block(
         return Ok(());
     }
 
-    let block = InstructionBlock::build(working_instructions.to_vec(), terminator, program)?;
+    // This leaves working_instructions and working_label in their respective "empty" states.
+    let block = InstructionBlock::build(std::mem::take(working_instructions), terminator, program)?;
```
Shadow53, line 523: Rather than cloning through `to_vec` and using `.drain(..)` below to empty the `Vec`, using `mem::take` takes ownership of the `Vec` and leaves an empty one in its place.

See the note on `working_label` for why this seems fine to do.

quil-rs/src/program/graph.rs
Shadow53: That said, is it possible that the code could change in the future such that we try to recover from errors? If that were to happen, this assumption would be invalidated and it could lead to weird bugs.

quil-rs/src/program/graph.rs
```diff
@@ -656,8 +653,8 @@ impl ScheduledProgram {
                     &mut working_label,
                     program,
                     instruction_index,
-                ),
-            }?;
+                )?,
+            };
```
Shadow53, line 657: This and above changes clarify the logic of the `match` statement, IMO, by returning errors when they happen instead of at the very end of a long statement.

quil-rs/benches/corpus.rs
```diff
@@ -1,15 +1,36 @@
...
+        return None;
+    }
+
+    let program =
+        fs::read_to_string(path).expect("failed to read quil program file");
+    let name = path
+        .file_name()
+        .expect("path should have file name component")
+        .to_str()
+        .expect("bad filename")
```
MarquessV, line 19:
```suggestion
        .expect("good filename")
```

quil-rs/src/program/graph.rs
Shadow53: You are correct: the logic is slightly different. I called those changes out [here](https://github.com/rigetti/quil-rs/pull/218#discussion_r1224894330) and [here](https://github.com/rigetti/quil-rs/pull/218#discussion_r1224894789). In short, I did not see any attempt to recover from the errors, so it seems safe to empty the values before erroring.

MarquessV: I see, reviewing the code more, I don't think this change in logic will introduce any problem. For future changes, maybe just add a note to the docstring that the function _will_ wipe out `working_instructions` and `working_label`.

MarquessV (approved):
Looks good, and a solid performance boost too! Approved.

kalzoo (commented):
lgtm!

Issue #216, opened 2023-06-09
Instructions should be hashable. 


MarquessV:
Most instructions implement __richcmp__ so that equality checks work the same way in Python as they do on Rust using `PartialEq/Eq`.

Implementing `__richcmp__` disables the default implementation of `__hash__`, since per the [Python data model](https://docs.python.org/3.5/reference/datamodel.html#object.__hash__), if `x` == `y`, then `hash(x) == hash(y)`, and Python can't guarantee this if a custom implementation was provided for equality.

Rust [makes this same recommendation](https://doc.rust-lang.org/std/hash/trait.Hash.html#hash-and-eq), and implements it by default if you can derive `Hash` and `Eq` on type.

PR #223, opened 2023-06-13
Instructions should be hashable

MarquessV:
An attempt at closing #216

### DIFF
--- a/quil-rs/src/hash.rs
+++ b/quil-rs/src/hash.rs
@@ -1,7 +1,4 @@
-use std::{
-    collections::{hash_map::DefaultHasher, HashMap},
-    hash::{Hash, Hasher},
-};
+use std::hash::{Hash, Hasher};
 
 /// Hashes a f64 using its u64 representation.
 ///
@@ -20,55 +17,15 @@ pub(crate) fn hash_f64<H: Hasher>(value: f64, state: &mut H) {
     }
 }
 
-/// Hashes a HashMap by taking the XOR sum of each key value pair.
-///
-/// This function uses XOR, because as a commutative operation, we get the same,
-/// unique XOR sum for each unique set of key value pairs, regardless of their order.
-#[inline]
-pub(crate) fn hash_hashmap<H: Hasher, K: Hash, V: Hash>(map: &HashMap<K, V>, state: &mut H) {
-    let mut final_hash = 0;
-    for element in map {
-        let mut hasher = DefaultHasher::new();
-        element.hash(&mut hasher);
-        final_hash ^= hasher.finish();
-    }
-
-    state.write_u64(final_hash)
-}
-
 #[cfg(test)]
 mod test {
-    use proptest::prelude::*;
-
-    use super::{hash_f64, hash_hashmap};
-    use std::{
-        collections::{hash_map::DefaultHasher, HashMap},
-        hash::Hasher,
-    };
-
-    fn get_map_hash(map: &HashMap<String, u32>) -> u64 {
-        let mut hasher = DefaultHasher::new();
-        hash_hashmap(map, &mut hasher);
-        hasher.finish()
-    }
+    use super::hash_f64;
+    use std::{collections::hash_map::DefaultHasher, hash::Hasher};
 
     fn get_f64_hash(float: f64) -> u64 {
         let mut hasher = DefaultHasher::new();
         hash_f64(float, &mut hasher);
         hasher.finish()
-    }
-
-    proptest! {
-        #[test]
-        fn test_hash_hashmap((map1, map2) in (any::<HashMap::<String, u32>>(), any::<HashMap::<String, u32>>())) {
-            let hash1 = get_map_hash(&map1);
-            let hash2 = get_map_hash(&map2);
-            if map1 == map2 {
-                assert_eq!(hash1, hash2);
-            } else {
-                assert_ne!(hash1, hash2);
-            }
-        }
     }
 
     #[test]
--- a/quil-rs/src/hash.rs
+++ b/quil-rs/src/hash.rs
@@ -1,0 +1,78 @@
+use std::{
+    collections::{hash_map::DefaultHasher, HashMap},
+    hash::{Hash, Hasher},
+};
+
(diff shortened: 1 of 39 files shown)


quil-rs/src/instruction/classical.rs
```diff
@@ -36,6 +36,21 @@ pub enum ArithmeticOperand {
...
+            Self::MemoryReference(operand) => operand.hash(state),
+        }
+    }
+}
+
+#[inline]
+fn hash_float(value: &f64, state: &mut impl std::hash::Hasher) {
+    std::hash::Hash::hash(&format!("{value}"), state)
+}
+
```
MarquessV, line 53: We can't derive `Hash` on structs containing a float because they don't have `Hash` implementations. This is in part because it's possible to write functions `f(a) != f(b)` for `a` and `b`, where `a == b`. This is mostly because of the semantics around `NaN`s and `+0` and `-0`. The quil spec isn't clear on what these values mean in the language, but I don't think they are generally cases we are too concerned with. Though, I could be wrong on that.

My thought, is that, for the purposes of `quil-rs`, if it formats two floats in the same way in a program, then it is effectively the same float. The downsides here are boilerplate and efficiency. However, in my survey of the issue, it seems that serialization  is among the top recommendations for handling this. 



Alternatively, there crates like [ordered-float](https://crates.io/crates/ordered-float) that wrap floats and offer implementations of `Hash`. However, I'm not sure if introducing another dependency, breaking changes, and the negative impact on ergonomics are worth it. 


For what it's worth, an experimental crate, [objecthash](https://crates.io/crates/objecthash) exists. It offers hash implementations for deeply nested structures, which can include float. From what I can understand, this also uses a fancy, cryptographically secure form of serialization.

jselig-rigetti: I think that the output of formatting being the hasher seems reasonable, since that covers the `+0` vs `-0` case (if a distinction is even necessary), and I'm not sure `NaN` is even allowed, just looking at the [grammar](https://github.com/quil-lang/quil/blob/289bc212017772c8e33f5d1a683f3dd4b8d83d42/grammars/Quil.g4#L230) it seems like no.

genos: I'm a fan of `ordered-float` for other reasons too, but I understand not wanting to pull in another dependency. For what it's worth, [here's](https://github.com/rigetti/quil-rs/blob/main/quil-rs/src/expression.rs#L155-L166) how we did it for `quil_rs::Expression::Number`; since the docs claim that `f64.to_bits()` is very portable, we just turned the floats to bits and hashed those.

MarquessV: Ah, thanks for pointing me to that!. I forgot we had hashable expressions already. That looks promising, and I would rather have a consistent way of hashing floats across the board. I was worried about the portability of `to_bits`, despite what the docs say, but if it's been in use already, that calms the worries I had. I think what I'll do is lift that logic up so it can be shared with the instruction module.

quil-rs/src/hash.rs
```diff
@@ -0,0 +1,78 @@
...
+pub(crate) fn hash_hashmap<H: Hasher, K: Hash, V: Hash>(map: &HashMap<K, V>, state: &mut H) {
+    let mut final_hash = 0;
+    for element in map {
+        let mut hasher = DefaultHasher::new();
+        element.hash(&mut hasher);
+        final_hash ^= hasher.finish();
+    }
+
+    state.write_u64(final_hash)
+}
```
MarquessV, line 37: This approach takes inspiration from how the rustc internals, and STLs in other languages hash unordered collections like `HashMap`s. Unfortunately, this algorithm conflicts with the default SipHash algorithm Rust currently uses for `HashMap`s, where the order of values is fundamentally important for achieving the security it offers. If we ignore the order with a commutative sum like this, it is theoretically possible to generate a set of values that will cause collisions and such a set could be used to attempt a DoS attack if this hashing algorithm is used somewhere inside of a service not expecting it. If hashing instructions is mostly a user utility then this seems like a non-issue, but as a library it isn't possible to discern how the feature will be used in all circumstances.

An alternative algorithm would be to take the hash of each key and value pair, sort them, then iterate over that list and use them to update the hasher state. This wouldn't betray the underlying SipHash algorithm, but it wouldn't have the expected runtime characteristics of a hashing function because of the sort and allocations needed.

We can also not allow any instructions containing a `HashMap` or similar to be hashed. This would impact `WaveformInvocation`s, `Capture`s, `Pulse`s, `FrameDefinition`s, and because they are all members of the `Instruction` enum, it also couldn't be hashed.

BatmanAoD: I'm not really familiar with how hash-DoS attacks work or how SipHash helps protect against them, but why is the order important for the security guarantee?

genos: https://stackoverflow.com/questions/52184366/why-does-hashmap-need-a-cryptographically-secure-hashing-function

TL;DR specifically crafted input can cause lots of hash collisions, leading to your hash map runtime exploding and DOSing the service.

As for commutativity, if order isn’t important, then that makes it even easier to have hash collisionss.

MarquessV: A couple of additional thoughts, after stewing and discussing with a few folks offline:

We're not sure how widely used this would be, and it may not be worth supporting this at all. There is an immediate use case for hashing instructions in pyQuil, but it may not need these specific instructions. In that case, we may omit support some instructions. If someone asks for it, we can take the opportunity to understand the use case and circle back. There are also other measure we could take to well to partially support hashing the `Instruction` type in Python if necessary.

Also, @BatmanAoD pulled up some benchmarks showing that, for some datasets, `BTreeMap` is actually faster. This may be an interesting thing for @Shadow53 to benchmark, and if it was indeed faster for our use case, it would just so happen to solve this problem for us.

MarquessV: I've reviewed the pyQuil tests and a/b tested them with and without having the hashable instructions in question and confirmed they aren't needed. I've gone ahead and removed the Hash implementation for those instructions.

kalzoo (approved):
#️⃣

PR #230, opened 2023-06-21
ci: ignore Cargo.lock in prerelease

kalzoo:
This PR follows the advice from the [broken build step in the pipeline](https://github.com/rigetti/quil-rs/actions/runs/5328135255/jobs/9652408753).

### DIFF
--- a/knope.toml
+++ b/knope.toml
@@ -26,7 +26,7 @@ command = "sh ./scripts/sync_versions.sh"
 
 [[workflows.steps]]
 type = "Command"
-command = "git add Cargo.lock quil-py/Cargo.toml"
+command = "git add quil-py/Cargo.toml"
 
 [[workflows.steps]]
 type = "Command"
@@ -50,7 +50,7 @@ command = "sh ./scripts/sync_versions.sh"
 
 [[workflows.steps]]
 type = "Command"
-command = "git add Cargo.lock quil-py/Cargo.toml"
+command = "git add quil-py/Cargo.toml"
 
 [[workflows.steps]]
 type = "Command"
--- a/knope.toml
+++ b/knope.toml
@@ -1,17 +1,11 @@
 [packages.quil-rs]
 versioned_files = ["quil-rs/Cargo.toml"]
-# Temprorarily disabling this so we can manually write the changelog entry
-# for 0.17.0 since knope doesn't handle generating the changelog for a large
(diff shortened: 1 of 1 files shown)


MarquessV:
I'm good with this, one other thing I'll throw out there is that maybe `cargo deny` should only be configured to throw warnings? It will complain about the lockfile, despite authors having no direct control over it. Sometimes we can fix those issues by editing `Cargo.toml`, but that isn't always desirable.

kalzoo:
> I'm good with this, one other thing I'll throw out there is that maybe `cargo deny` should only be configured to throw only warnings? It will complain about the lockfile, despite authors having no direct control over it. Sometimes we can fix those issues by editing `Cargo.toml`, but that isn't always desirable.

True, and quickly relevant in this PR 😄 In principle this makes sense to me though:

- If there were a lockfile it might not expose issues which downstream consumers would then see, lacking that lockfile's constraints
- It encourages the "correct" response, which is changing bounds in Cargo.toml, rather than papering them over in the lockfile
- The author _can_ at least reproduce (and thus resolve) warnings by running `cargo update` and then `cargo deny` themselves (although there are probably edge cases due to architecture)

Thoughts?

MarquessV (approved):

MarquessV:
> > I'm good with this, one other thing I'll throw out there is that maybe `cargo deny` should only be configured to throw only warnings? It will complain about the lockfile, despite authors having no direct control over it. Sometimes we can fix those issues by editing `Cargo.toml`, but that isn't always desirable.
> 
> True, and quickly relevant in this PR 😄 In principle this makes sense to me though:
> 
>     * If there were a lockfile it might not expose issues which downstream consumers would then see, lacking that lockfile's constraints
> 
>     * It encourages the "correct" response, which is changing bounds in Cargo.toml, rather than papering them over in the lockfile
> 
>     * The author _can_ at least reproduce (and thus resolve) warnings by running `cargo update` and then `cargo deny` themselves (although there are probably edge cases due to architecture)
> 
> 
> Thoughts?

Your points are all valid, and I don't think it's worth hanging up this PR, we can always adjust the approach later. Maybe PRs are the best place to put the check, since it will lead to quicker action, but in almost every case what `cargo deny` complains about is tangential to the content of the PR and now there is another layer of misdirection in solving it. Still, as you say, there are ways around it, and it probably is better for the health of the project to enforce action, even if it can be a bit confusing.

PR #232, opened 2023-06-22
perf!: optimize Program::get_frames_for_instruction

### LINKED ISSUE #231: Improve performance of get_frames_for_instruction
The `Program::get_frames_for_instruction` function shows up as a meaningful portion of the computation time of live programs. This seems to be unnecessary; what it's doing is simple. The current codebase is inefficient in terms of allocation and also in redundant calls to this method resulting in rework.

Goals:

* Reduce the number of calls to `get_frames_for_instruction` in a given code path
* Use `Option` in place of empty `HashSet` to reduce allocation and dropping

kalzoo:
* Call this method less frequently
* Return `Option::None` rather than `HashSet::new` when nothing matches

Evaluation against a yet-to-be-published benchmark indicates a 50% performance increase in `ScheduledProgram::from_program`

Closes #231

### DIFF
--- a/quil-rs/src/program/frame.rs
+++ b/quil-rs/src/program/frame.rs
@@ -207,10 +207,13 @@ pub struct MatchedFrames<'a> {
 }
 
 impl<'a> MatchedFrames<'a> {
+    /// Which concrete frames are blocked and not used.
+    /// This set is mutually exclusive with `used`.
     pub fn blocked(&self) -> &HashSet<&'a FrameIdentifier> {
         &self.blocked
     }
 
+    /// Which concrete frames are used by the [`Instruction`]
     pub fn used(&self) -> &HashSet<&'a FrameIdentifier> {
         &self.used
     }
--- a/quil-rs/src/program/frame.rs
+++ b/quil-rs/src/program/frame.rs
@@ -190,9 +190,9 @@ pub(crate) struct FrameMatchConditions<'a> {
     /// are blocked by an [`Instruction`]. A "blocked" frame is one which is not used by the
     /// `Instruction` but is not available for use by other instructions while this one executes.
     ///
-    /// **Note**: for efficiency in computation, this may or include frames also matched by `used`.
+    /// **Note**: for efficiency in computation, this may match frames also matched by `used`.
     /// In order to query which frames are _blocked but not used_, both conditions must first
-    /// be evaluated in the scope of the available frames, and
+    /// be evaluated in the scope of the available frames.
     pub blocked: Option<FrameMatchCondition<'a>>,
 }
 
--- a/quil-rs/src/program/frame.rs
+++ b/quil-rs/src/program/frame.rs
@@ -45,12 +45,12 @@ impl FrameSet {
     ) -> MatchedFrames<'s> {
         let used = condition
             .used
-            .map(|c| self.get_matching_keys_for_condition(c));
+            .map_or_else(HashSet::new, |c| self.get_matching_keys_for_condition(c));
 
-        let blocked = condition.blocked.map(|c| {
+        let blocked = condition.blocked.map_or_else(HashSet::new, |c| {
             let mut blocked = self.get_matching_keys_for_condition(c);
 
-            if let Some(used) = &used {
+            if !used.is_empty() {
                 blocked.retain(|&f| !used.contains(&f));
             }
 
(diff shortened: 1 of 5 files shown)


quil-rs/src/program/frame.rs
```diff
@@ -154,3 +175,33 @@ pub(crate) enum FrameMatchCondition<'a> {
...
+
+/// A pair of conditions to match frames within a [`crate::Program`] (or another scope).
+///
+/// This allows for deferred evaluation of matching an instruction against available frames.
+pub(crate) struct FrameMatchConditions<'a> {
+    /// A condition to identify which frames within a [`crate::Program`] (or another scope)
+    /// are actively used by an [`Instruction`].
+    ///
+    /// If `None`, then this [`Instruction`] does not use any frames, regardless of which are available.
+    pub used: Option<FrameMatchCondition<'a>>,
```
genos, line 187: If the struct is `pub(crate)`, should these fields be as well?

quil-rs/src/program/frame.rs
```diff
@@ -154,3 +175,33 @@ pub(crate) enum FrameMatchCondition<'a> {
...
+
+    /// A condition to identify which frames within a [`crate::Program`] (or another scope)
+    /// are blocked by an [`Instruction`]. A "blocked" frame is one which is not used by the
+    /// `Instruction` but is not available for use by other instructions while this one executes.
+    ///
+    /// **Note**: for efficiency in computation, this may or include frames also matched by `used`.
```
genos, line 193:
```suggestion
    /// **Note**: for efficiency in computation, this may include frames also matched by `used`.
```

quil-rs/src/program/frame.rs
kalzoo: Is there value to doing that? I tend to avoid that for readability. In contrast to nested modules, the `pub(crate)` is visible in proximity, and so my take is that the developer would not be surprised to find that that field is not accessible outside the crate

quil-rs/src/program/frame.rs
kalzoo: Thank you, fixed

quil-rs/src/program/frame.rs
genos: Fair enough! Just saw the mismatch and thought I'd ask.

quil-rs/src/program/frame.rs
```diff
@@ -154,3 +175,33 @@ pub(crate) enum FrameMatchCondition<'a> {
...
+
+/// The product of evaluating  [`FrameMatchConditions`] in the scope of available frames (such as within a [`crate::Program`]).
+pub struct MatchedFrames<'a> {
+    /// Which concrete frames are used by the [`Instruction`]
+    pub used: Option<HashSet<&'a FrameIdentifier>>,
+
+    /// Which concrete frames are blocked and not used.
+    /// This set must be mutually exclusive with `used`.
+    pub blocked: Option<HashSet<&'a FrameIdentifier>>,
+}
```
Shadow53, line 207: In order to ensure that the mutual exclusivity is upheld, should these fields be private with immutable getters?

Alternatively, would it make more sense to have a single `HashMap<&'a FrameIdentifier, FrameMatchType>` where

```rust
enum FrameMatchType {
    Borrowed,
    Used,
}
```

(I like the separation of these fields, I'm just looking at how we can ensure the exclusivity is always true)

quil-rs/src/program/graph.rs
```diff
@@ -327,56 +327,54 @@ impl InstructionBlock {
...
+
+                                let previous_node_ids = last_instruction_by_frame
+                                    .entry((*frame).clone())
+                                    .or_default()
+                                    .get_dependencies_for_next_user(node);
+
+                                for previous_node_id in previous_node_ids {
+                                    add_dependency!(graph, previous_node_id => node, ExecutionDependency::StableOrdering);
+                                }
                             }
```
Shadow53, line 354: I think it makes as much sense, if not more, to avoid the use of `Option` and use empty `HashSet`s for the cases where there are no matching frames.

quil-rs/src/program/frame.rs
kalzoo: > should these fields be private with immutable getters?

Yes, that's a good idea - since this struct is meant to be derived in an analysis step and then remain immutable.

> would it make more sense to have a single

In practice, these two groups are iterated over, which would become less efficient if we had to iterate over both with mutually exclusive filters for each.

quil-rs/src/program/graph.rs
kalzoo: Performance benchmarks seem to indicate a 5% regression from this change alone, but I'll bump the sample size up and see that I can reproduce this.

kalzoo: nvm - cannot reproduce now, it must have been due to something environmental. That's good news, since this is an ergonomic improvement.

kalzoo:
Preliminary internal benchmark run indicates that this change in isolation reduces a key metric from 7.2s -> 5.7s (~20% improvement). Internal tests are passing.

quil-rs/src/program/frame.rs
```diff
@@ -154,3 +175,43 @@ pub(crate) enum FrameMatchCondition<'a> {
...
+
+/// The product of evaluating  [`FrameMatchConditions`] in the scope of available frames (such as within a [`crate::Program`]).
+pub struct MatchedFrames<'a> {
+    /// Which concrete frames are blocked and not used.
+    /// This set is mutually exclusive with `used`.
+    blocked: HashSet<&'a FrameIdentifier>,
+
+    /// Which concrete frames are used by the [`Instruction`]
+    used: HashSet<&'a FrameIdentifier>,
+}
```
Shadow53, line 207: These doc comments should be copied/moved to the public methods

Shadow53 (approved):
Looks good! Just one note about moving doc comments from a private location to the equivalent public location.

PR #233, opened 2023-06-23
refactor!: more optimizations

Shadow53:
There are two primary changes here:

1. The qubits used by a `Program` is now tracked when adding instructions using `add_instruction(s)`. This removes the need to built the `HashSet` whenever the set of qubits is requested.
    - The downside is that the `instructions` field should no longer be `pub` to avoid accidentally working around this behavior. This is my next step.
    - This change introduced an improvement of 5-7% in the relevant benchmarks, and no difference otherwise.
2. `ScheduledProgram::from_program` used to take a `&Program` and clone the parts it needed. It now instead takes references to those things, removing the excess clones.
    - With real-world programs, this saw a ~14% improvement in benchmarking. A few benchmark programs saw improvements of ~60%! (I'm not sure how much I trust that number, but everything saw _some_ improvement.

### DIFF
--- a/quil-rs/src/instruction/mod.rs
+++ b/quil-rs/src/instruction/mod.rs
@@ -481,7 +481,9 @@ impl Instruction {
             } else {
                 FrameMatchCondition::And(vec![
                     FrameMatchCondition::ExactQubits(qubits.iter().collect()),
-                    FrameMatchCondition::AnyOfNames(frame_names.iter().collect()),
+                    FrameMatchCondition::AnyOfNames(
+                        frame_names.iter().map(String::as_str).collect(),
+                    ),
                 ])
             }),
             Instruction::Fence(Fence { qubits }) => {
--- a/quil-rs/src/instruction/mod.rs
+++ b/quil-rs/src/instruction/mod.rs
@@ -647,7 +647,7 @@ RX(2) 0",
         )
         .unwrap();
         let closure = |expr: &mut Expression| *expr = Expression::Variable(String::from("a"));
-        for instruction in program.instructions.iter_mut() {
+        for instruction in program.body_instructions_mut() {
             instruction.apply_to_expressions(closure);
         }
 
--- a/quil-rs/src/program/graph.rs
+++ b/quil-rs/src/program/graph.rs
@@ -199,13 +199,13 @@ pub type DependencyGraph = GraphMap<ScheduledGraphNode, HashSet<ExecutionDepende
 /// which include no control flow instructions aside from an (optional) terminating control
 /// flow instruction.
 #[derive(Clone, Debug)]
-pub struct InstructionBlock {
-    pub instructions: Vec<Instruction>,
+pub struct InstructionBlock<'a> {
+    pub instructions: Vec<&'a Instruction>,
     pub(super) graph: DependencyGraph,
-    pub terminator: BlockTerminator,
+    pub terminator: BlockTerminator<'a>,
 }
 
-impl Default for InstructionBlock {
+impl<'a> Default for InstructionBlock<'a> {
     fn default() -> Self {
         Self {
             instructions: Default::default(),
@@ -293,11 +293,12 @@ impl PreviousNodes {
     }
 }
 
-impl InstructionBlock {
+impl<'a> InstructionBlock<'a> {
     pub fn build(
-        instructions: Vec<Instruction>,
-        terminator: Option<BlockTerminator>,
-        program: &Program,
+        // The set of instructions in the block, a subset of the `program`.
+        instructions: Vec<&'a Instruction>,
+        terminator: Option<BlockTerminator<'a>>,
+        program: &'a Program,
     ) -> ScheduleResult<Self> {
         let mut graph: DependencyGraph = GraphMap::new();
         // Root node
@@ -306,6 +307,7 @@ impl InstructionBlock {
         let mut last_classical_instruction = ScheduledGraphNode::BlockStart;
 
         // Store the instruction index of the last instruction to block that frame
+        // TODO: with_capacity program.frames.len()
         let mut last_instruction_by_frame: HashMap<FrameIdentifier, PreviousNodes> = HashMap::new();
         let mut last_timed_instruction_by_frame: HashMap<FrameIdentifier, PreviousNodes> =
             HashMap::new();
@@ -316,6 +318,7 @@ impl InstructionBlock {
 
         for (index, instruction) in instructions.iter().enumerate() {
             let node = graph.add_node(ScheduledGraphNode::InstructionIndex(index));
+            let instruction = *instruction;
 
             let instruction_role = InstructionRole::from(instruction);
             match instruction_role {
@@ -461,7 +464,7 @@ impl InstructionBlock {
 
     /// Return a particular-indexed instruction (if present).
     pub fn get_instruction(&self, node_id: usize) -> Option<&Instruction> {
-        self.instructions.get(node_id)
+        self.instructions.get(node_id).copied()
     }
 
     /// Return the count of executable instructions in this block.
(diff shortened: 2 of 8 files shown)


quil-rs/src/instruction/mod.rs
```diff
@@ -502,7 +502,7 @@ impl Instruction {
                         set.insert(qubit);
                         set
                     }
-                    None => qubits_available,
+                    None => qubits_available.iter().collect(),
```
Shadow53, line 505: The function this is in used to take a `HashSet<&Qubit>` as input and output. Since it now takes a `& HashSet<Qubit>` as input, we have to convert it to `HashSet<&Qubit>` in the one case we want to return that set.

quil-rs/src/program/graph.rs
```diff
@@ -316,6 +317,7 @@ impl InstructionBlock {
 
         for (index, instruction) in instructions.iter().enumerate() {
             let node = graph.add_node(ScheduledGraphNode::InstructionIndex(index));
+            let instruction = *instruction;
```
Shadow53, line 320: Was easier to dereference the `&&Instruction` here than to do `(*instruction)` multiple times below.

quil-rs/src/program/graph.rs
```diff
@@ -461,7 +463,7 @@ impl InstructionBlock {
 
     /// Return a particular-indexed instruction (if present).
     pub fn get_instruction(&self, node_id: usize) -> Option<&Instruction> {
-        self.instructions.get(node_id)
+        self.instructions.get(node_id).copied()
```
Shadow53, line 466: Another case of `&&Instruction`. Since references are `Copy`, this is effectively a single dereference.

quil-rs/src/program/graph.rs
```diff
@@ -474,29 +476,29 @@ impl InstructionBlock {
         self.instructions.is_empty()
     }
 
-    pub fn set_exit_condition(&mut self, terminator: BlockTerminator) {
-        self.terminator = terminator
+    pub fn set_exit_condition(&mut self, terminator: BlockTerminator<'a>) {
+        self.terminator = terminator;
```
Shadow53, line 480: Added a semicolon for consistency

quil-rs/src/program/graph.rs
```diff
@@ -528,6 +530,7 @@ fn terminate_working_block(
     let block = InstructionBlock::build(std::mem::take(working_instructions), terminator, program)?;
     let label = working_label
         .take()
+        .cloned()
         .unwrap_or_else(|| ScheduledProgram::generate_autoincremented_label(blocks));
```
Shadow53, line 534: `working_label` changed from `Option<String>` to `Option<&String>`, but the `unwrap_or_else` must return a `String`. So, we must clone the string if `working_label.is_some()`

quil-rs/src/program/graph.rs
```diff
@@ -541,17 +544,17 @@ fn terminate_working_block(
...
         let mut working_label = None;
-        let mut working_instructions: Vec<Instruction> = vec![];
+        let mut working_instructions: Vec<&'a Instruction> = vec![];
         let mut blocks = IndexMap::new();
 
-        let instructions = program.to_instructions();
+        // let instructions = program.to_instructions();
 
-        for (index, instruction) in instructions.into_iter().enumerate() {
+        for (index, instruction) in program.instructions.iter().enumerate() {
```
Shadow53, line 557: Per @kalzoo, the "headers" (everything added to the `Vec<Instruction>` returned by `to_instructions()` that is _not_ part of `program.instructions`) are not used here, so we can iterate over the `instructions` field directly.

quil-rs/src/program/memory.rs
```diff
@@ -315,12 +315,9 @@ impl Expression {
...
-
-        for expression in self.parameters.values() {
-            result.extend(expression.get_memory_references());
-        }
-
-        result
+        self.parameters
+            .values()
+            .flat_map(Expression::get_memory_references)
+            .collect()
```
Shadow53, line 321: Made this change to see if it made any performance difference, left it because it didn't hurt anything and is more readable, IMO

quil-rs/src/program/mod.rs
```diff
@@ -65,6 +65,9 @@ pub struct Program {
     pub memory_regions: BTreeMap<String, MemoryRegion>,
     pub waveforms: BTreeMap<String, Waveform>,
     pub instructions: Vec<Instruction>,
+    // private fields used for caching operations
+    // TODO: will require the above fields to also be private
+    used_qubits: HashSet<Qubit>,
```
Shadow53, line 70: Recognize the TODO, will address tomorrow

quil-rs/src/program/graph.rs
```diff
@@ -541,17 +544,17 @@ fn terminate_working_block(
...
     #[allow(unused_assignments)]
-    pub fn from_program(program: &Program) -> ScheduleResult<Self> {
+    pub fn from_program(program: &'a Program) -> ScheduleResult<Self> {
         let mut working_label = None;
-        let mut working_instructions: Vec<Instruction> = vec![];
+        let mut working_instructions: Vec<&'a Instruction> = vec![];
         let mut blocks = IndexMap::new();
 
-        let instructions = program.to_instructions();
+        // let instructions = program.to_instructions();
```
Shadow53, line 555: Delete

quil-rs/src/program/mod.rs
```diff
@@ -104,6 +108,41 @@ impl Program {
...
+                self.instructions.push(Instruction::Capture(capture));
+            }
+            Instruction::Pulse(pulse) => {
+                self.used_qubits.extend(pulse.frame.qubits.clone());
+                self.instructions.push(Instruction::Pulse(pulse));
+            }
+            Instruction::RawCapture(raw_capture) => {
+                self.used_qubits.extend(raw_capture.frame.qubits.clone());
+                self.instructions.push(Instruction::RawCapture(raw_capture));
+            }
```
Shadow53, line 145: Moved from `get_used_qubits` with the addition of adding each instruction to the instructions list as well. I decided this was better than

```rust
other => {
    match &other {
        // Add qubits to set
        _ => {
            // Has no qubits, ignore
        }
    }
    self.instructions.push(other);
}
```

because it makes it more difficult for a branch to slip through the cracks, e.g. a new instruction that should be handled specially but isn't caught by the compiler because it falls through both default cases.

quil-rs/src/program/mod.rs
```diff
@@ -640,7 +664,7 @@ I 0
         let expected_owned = vec![Qubit::Fixed(0), Qubit::Variable("q".to_string())];
         let expected = expected_owned.iter().collect::<HashSet<_>>();
         let actual = program.get_used_qubits();
-        assert_eq!(expected, actual);
+        assert_eq!(expected, actual.iter().collect());
```
Shadow53, line 667: Another side effect of switching `HashSet<&Qubit>` to `&HashSet<Qubit>`

Shadow53 (commented):
Left some comments on the changes and noted two places where more work needs to be done. I think this is ready for preliminary review, though, and the remaining changes are basically cleanup.

quil-rs/src/program/graph.rs
```diff
@@ -474,29 +476,29 @@ impl InstructionBlock {
...
 }
 
 #[derive(Clone, Debug)]
-pub enum BlockTerminator {
+pub enum BlockTerminator<'a> {
     Conditional {
-        condition: MemoryReference,
-        target: String,
+        condition: &'a MemoryReference,
+        target: &'a String,
```
genos, line 488: If you're including lifetimes now, does it make sense/help at all to move from `&'a String` to `&'a str`?

quil-rs/src/program/graph.rs
kalzoo: yep. We'll need to document this intent somewhere - all instructions within all blocks must be completely contained within `Program.instructions` (the "body" of the program) in order for this to be valid.

I think a (breaking) rename of `.instructions` to `.body` or `.instruction_body` might help make this clear, together with making it [crate-]private and then exposing a getter.

PR #234, opened 2023-06-23
refactor!: cache used qubits on Program

Shadow53:
Split out from the original PR, #233.

This changes the logic of `get_used_qubits` from calculating the used qubits at call time to calculating them as instructions are added to the `Program`.


quil-rs/src/instruction/mod.rs
```diff
@@ -502,7 +502,7 @@ impl Instruction {
                         set.insert(qubit);
                         set
                     }
-                    None => qubits_available,
+                    None => qubits_available.iter().collect(),
```
Shadow53, line 505: The function this is in used to take a `HashSet<&Qubit>` as input and output. Since it now takes a `& HashSet<Qubit>` as input, we have to convert it to `HashSet<&Qubit>` in the one case we want to return that set.

quil-rs/src/program/mod.rs
```diff
@@ -104,6 +108,41 @@ impl Program {
...
+                self.instructions.push(Instruction::Capture(capture));
+            }
+            Instruction::Pulse(pulse) => {
+                self.used_qubits.extend(pulse.frame.qubits.clone());
+                self.instructions.push(Instruction::Pulse(pulse));
+            }
+            Instruction::RawCapture(raw_capture) => {
+                self.used_qubits.extend(raw_capture.frame.qubits.clone());
+                self.instructions.push(Instruction::RawCapture(raw_capture));
+            }
```
Shadow53, line 145: Moved from `get_used_qubits` with the addition of adding each instruction to the instructions list as well. I decided this was better than

```rust
other => {
    match &other {
        // Add qubits to set
        _ => {
            // Has no qubits, ignore
        }
    }
    self.instructions.push(other);
}
```

because it makes it more difficult for a branch to slip through the cracks, e.g. a new instruction that should be handled specially but isn't caught by the compiler because it falls through both default cases.

quil-rs/src/program/mod.rs
```diff
@@ -640,7 +664,7 @@ I 0
         let expected_owned = vec![Qubit::Fixed(0), Qubit::Variable("q".to_string())];
         let expected = expected_owned.iter().collect::<HashSet<_>>();
         let actual = program.get_used_qubits();
-        assert_eq!(expected, actual);
+        assert_eq!(expected, actual.iter().collect());
```
Shadow53, line 667: Another side effect of switching `HashSet<&Qubit>` to `&HashSet<Qubit>`

Shadow53 (commented):
Copied comments from the original PR.

quil-rs/src/program/mod.rs
```diff
@@ -74,6 +74,8 @@ pub struct Program {
     pub memory_regions: BTreeMap<String, MemoryRegion>,
     pub waveforms: BTreeMap<String, Waveform>,
     instructions: Vec<Instruction>,
+    // private field used for caching operations
+    used_qubits: HashSet<Qubit>,
```
kalzoo, line 78:
```suggestion
    /// The qubits used by all of the instructions; that is, the qubits on which the program operates.
    /// Note that this does not include all qubits _referenced_ by the program (such as those within frames defined but never used)
    used_qubits: HashSet<Qubit>,
```

The other fields may not be `///` documented, but this one can be the first 😄

kalzoo (approved):
marked as draft, but looks good to go to me 👍

PR #236, opened 2023-06-26
feat: add clone_without_body_instructions to Program

### LINKED ISSUE #235: Add clone_without_body_instructions to Program
`pyquil` has a `copy_everything_except_instructions`. Since we use this as a backing for `pyquil` programs, we should implement that behavior natively instead of just in the wrapper.

Shadow53:
Resolves #235.

I noticed a couple of tests that were not testing the correct thing, so I fixed them while I was here.

### DIFF
--- a/quil-rs/src/program/mod.rs
+++ b/quil-rs/src/program/mod.rs
@@ -101,6 +101,17 @@ impl Program {
         self.instructions.iter_mut()
     }
 
+    /// Like `Clone`, but does not clone the body instructions.
+    pub fn clone_without_body_instructions(&self) -> Self {
+        Self {
+            instructions: Vec::new(),
+            calibrations: self.calibrations.clone(),
+            frames: self.frames.clone(),
+            memory_regions: self.memory_regions.clone(),
+            waveforms: self.waveforms.clone(),
+        }
+    }
+
     /// Add an instruction to the end of the program.
     pub fn add_instruction(&mut self, instruction: Instruction) {
         match instruction {
--- a/quil-rs/src/program/mod.rs
+++ b/quil-rs/src/program/mod.rs
@@ -418,12 +418,12 @@ DEFCAL I 0:
     DELAY 0 1.0
 DEFFRAME 0 \"rx\":
     HARDWARE-OBJECT: \"hardware\"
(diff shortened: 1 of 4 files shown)


quil-rs/src/program/mod.rs
```diff
@@ -407,12 +418,12 @@ DEFCAL I 0:
     DELAY 0 1.0
 DEFFRAME 0 \"rx\":
     HARDWARE-OBJECT: \"hardware\"
-DEFWAVEFORM custom 6.0:
+DEFWAVEFORM custom:
```
Shadow53, line 421: This created a parsing error

quil-rs/src/program/mod.rs
```diff
@@ -407,12 +418,12 @@ DEFCAL I 0:
...
     1,2
 I 0
 ";
-        let a = Program::from_str(input);
-        let b = Program::from_str(input);
+        let a = Program::from_str(input).unwrap();
+        let b = Program::from_str(input).unwrap();
```
Shadow53, line 426: The parsing error was not caught because we were inadvertently testing that the errors were equal, not the parsed programs.

Shadow53:
Is [this fix](https://github.com/rigetti/pyquil/pull/1600) something that needs to be applied here as well?

MarquessV:
> Is [this fix](https://github.com/rigetti/pyquil/pull/1600) something that needs to be applied here as well?

I don't think so. In `pyQuil` a `Program` maintains a `declaration` mapping separately from the instructions in the program. The bug there had to do with copying the `declarations` mapping without the corresponding instructions, leading to a `Program` in an invalid state.

In `quil-rs`, the instructions in the program are derived directly from each member of `Program`. When a `DECLARE` instruction is added to a `Program`, it gets added to the `memory_regions` field and when a user calls `to_instructions()`, `memory_regions` is used to derive the corresponding `Instruction`s. With this approach the internal state of the program and the instructions it outputs can't differ like they can in pyQuil.

kalzoo (approved):
good catch!

Looks good, would like @MarquessV 's stamp of approval as well.

MarquessV (approved):
Forgot to approve w/ my original comment. Looks good! 🚀

PR #237, opened 2023-06-28
fix: hash PiConstant like Expression::Number

genos:
We've noticed some issues in downstream consumers of `quil-rs` pertaining to, e.g., calibrations. If a user has a `DEFCAL RX(pi) 0 …` somewhere, `quil-rs` will keep the `Expression::PiConstant` around, but "simplify" it in the calibration to `RX(std::f64::consts::PI) 0` in practice. This will _not_ match the user-given calibration, causing sorrow and strife.

Since we have the `PartialEq` and `Eq` traits for `Expression` based on its `Hash` implementation, this PR changes the hash step for `PIConstant` to act like the equivalent `Expression::Number`'s hash.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1061,7 +1061,7 @@ checksum = "a1d01941d82fa2ab50be1e79e6714289dd7cde78eba4c074bc5a4374f650dfe0"
 
 [[package]]
 name = "quil-py"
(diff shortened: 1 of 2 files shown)


genos:
Superseded by #240

PR #238, opened 2023-06-28
fix: Expand all analog control instructions

### LINKED ISSUE #227: Calibration expansion doesn't account for parameterized qubits
Consider the following program:

```
RZ(%theta) %q
    SHIFT-PHASE %q \"rf\" -%theta
RZ(pi) 0
```

After expanding calibrations, the `RZ` instruction should be re-written as:

```
SHIFT-PHASE 0 "rf" -pi
```

However, quil-rs discounts the qubit parameter and prints:

```
SHIFT-PHASE q "rf" -pi
```

MarquessV:
Closes #227 - calibration expansion was actually working fine, we just weren't account for every type of instruction that could be in the calibration.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -31,7 +31,14 @@ ignore = []
 # https://embarkstudios.github.io/cargo-deny/checks/licenses/cfg.html
 [licenses]
 unlicensed = "deny"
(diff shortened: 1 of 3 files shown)


kalzoo (approved):
nice!

PR #240, opened 2023-06-29
fix(rs): 'simplify' pi expression to floating-point form

kalzoo:
This change makes expression simplification consistent between the following:

* `pi` is the root expression
* `pi` is a member of the expression

Previously, `pi/2` simplified to its floating point form (`1.57...`) while `pi` did not. This amends that, so that `simplify(pi) == 3.14...`).

### DIFF
--- a/quil-rs/src/expression/mod.rs
+++ b/quil-rs/src/expression/mod.rs
@@ -268,10 +268,10 @@ impl Expression {
     /// ```
     pub fn simplify(&mut self) {
         match self {
-            Expression::Address(_)
-            | Expression::Number(_)
-            | Expression::PiConstant
-            | Expression::Variable(_) => {}
+            Expression::Address(_) | Expression::Number(_) | Expression::Variable(_) => {}
+            Expression::PiConstant => {
+                *self = Expression::Number(Complex64::from(PI));
+            }
             _ => {
                 if let Ok(simpler) = simplification::run(self) {
                     *self = simpler;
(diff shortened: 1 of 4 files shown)


quil-rs/src/expression/mod.rs
```diff
@@ -1016,16 +1016,44 @@ mod tests {
...
+                simple_e=simple_e
+            );
         }
 
     }
 
+    #[test]
+    fn specific_simplification_tests() {
+        for (input, expected) in vec![
+            // ("pi", Expression::Number(PI.into())),
```
genos, line 1046: Uncomment?

quil-rs/src/expression/mod.rs
```diff
@@ -1016,16 +1016,44 @@ mod tests {
             prop_assert_eq!(x, expected);
         }
 
+        // Redundant clone: clippy does not correctly introspect the prop_assert_eq! macro
```
genos, line 1019: Nice find! I ended up splitting out a bunch of the proptests because of `clippy` arguing with me about clones.

quil-rs/src/expression/simplification.rs
```diff
@@ -268,7 +269,7 @@ fn expression_to_recexpr(expression: &Expression) -> RecExpr<Expr> {
 fn recexpr_to_expression(recexpr: RecExpr<Expr>) -> Result<Expression, SimplificationError> {
     fn helper(nodes: &[Expr], i: usize) -> Result<Expression, SimplificationError> {
         match nodes[i] {
-            Expr::Pi => Ok(Expression::PiConstant),
+            Expr::Pi => Ok(Expression::Number(PI.into())),
```
genos, line 272: 👏

genos (approved):
One commented out test, but other than that 👍 👍 👍

quil-rs/src/expression/simplification.rs
kalzoo: that's the one, thank you for the pointers

PR #242, opened 2023-06-29
feat(rs): Program::into_instructions

### LINKED ISSUE #241: Program::into_instructions
Currently, `Program` and some of its sub-structs have a `to_instructions` method which clones the inner instructions to populate a Vec. When the caller has ownership of the `Program` and wishes to consume it, there's no option for that.

That's where a `Program::into_instructions(self)` is needed, as it is for `CalibrationSet` and `FrameSet`.

kalzoo:
Closes #241 

Simple enough.

I did not see a neat way to share code between `Program::into_instructions` and `Program::to_instructions` without cloning more than strictly needs to be cloned in the latter.

I'd like to be able to support `as_instructions(&self) -> Vec<&Instruction>` but many of the owned instructions are created within the method (and not just referenced). If `CalibrationSet` et al were to be restructured so as to maintain their backing instructions in `Instruction` form (rather than decomposing them), we could do that (in another PR).

### DIFF
--- a/quil-rs/src/program/calibration.rs
+++ b/quil-rs/src/program/calibration.rs
@@ -336,6 +336,19 @@ impl CalibrationSet {
         self.measure_calibrations.extend(other.measure_calibrations);
     }
 
+    /// Return the Quil instructions which describe the contained calibrations, consuming the [`CalibrationSet`].
+    pub fn into_instructions(self) -> Vec<Instruction> {
+        self.calibrations
+            .into_iter()
+            .map(Instruction::CalibrationDefinition)
+            .chain(
+                self.measure_calibrations
+                    .into_iter()
+                    .map(Instruction::MeasureCalibrationDefinition),
+            )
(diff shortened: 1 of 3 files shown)


genos (approved):
LGTM!

One thought: would it make sense to turn `into_instructions` into an `IntoInstructions` trait that these things implement?

kalzoo:
> LGTM!
> 
> One thought: would it make sense to turn `into_instructions` into an `IntoInstructions` trait that these things implement?

I wondered that as well, but I wasn’t sure what that would buy us beyond the ability to be generic over Program, CalibrationSet, and FrameSet. Would that be useful?

genos:
> I wondered that as well, but I wasn’t sure what that would buy us beyond the ability to be generic over Program, CalibrationSet, and FrameSet. Would that be useful?

I guess not, unless something new comes along later that should also have an `into_instructions`.

PR #243, opened 2023-06-29
fix: Correct precedence rules are followed when matching measure calibrations.

### LINKED ISSUE #160: `get_match_for_measurement` should follow precedence rules of matching qubit
Per [this discussion](https://github.com/rigetti/quil-rs/pull/156#discussion_r1132712623) the precedence for determining the best match for a `Measurement` to a vector of `MeasureCalibrationDefinition` is as follows:

>    1. Match fixed qubit
>    2. Match variable qubit
>    3. Match no qubit (I don't recall this case but it's supported here)
> 
> The tiebreaker among calibrations of equal precedence is "last wins."

MarquessV:
closes #160


quil-rs/Cargo.toml
```diff
@@ -13,6 +13,7 @@ approx = { version = "0.5.1", features = ["num-complex"] }
 dot-writer = { version = "0.1.2", optional = true }
 egg = { version = "0.9.4", features = ["deterministic"] }
 indexmap = "1.6.1"
+itertools = "0.11.0"
```
genos, line 16: 🎮 `itertools` has entered the game 🎮

genos (approved):
👍

Issue #244, opened 2023-07-05
Expression simplification: associative addition and multiplication


kalzoo:
Small bug/missing feature: `(a[0] * 2 * pi) / (2 * pi)` does not simplify except to convert `pi` to floating point form.

PR #245, opened 2023-07-05
feat(rs): associate & commute in expression simplification

kalzoo:
Adds rules to the `egg` ruleset with tests to validate behavior.

Closes https://github.com/rigetti/quil-rs/issues/244

### DIFF
--- a/quil-rs/src/expression/mod.rs
+++ b/quil-rs/src/expression/mod.rs
@@ -1046,6 +1046,13 @@ mod tests {
             ("pi", Expression::Number(PI.into())),
             ("pi/2", Expression::Number((PI / 2.0).into())),
             ("pi * pi", Expression::Number((PI.powi(2)).into())),
+            (
+                "(a[0]*2*pi)/6.283185307179586",
+                Expression::Address(MemoryReference {
+                    name: String::from("a"),
+                    index: 0,
+                }),
+            ),
         ] {
             assert_eq!(
                 Expression::from_str(input).unwrap().into_simplified(),
--- a/quil-rs/src/expression/simplification.rs
+++ b/quil-rs/src/expression/simplification.rs
@@ -467,15 +467,17 @@ static RULES: Lazy<Vec<Rewrite>> = Lazy::new(|| {
         rw!("zero-add"      ; "(+ 0 ?a)"                => "?a"),
         rw!("cancel-sub"    ; "(- ?a ?a)"               => "0"),
         // multiplication & division
-        rw!("mul-zero"      ; "(* ?a 0)"                => "0"),
         rw!("zero-mul"      ; "(* 0 ?a)"                => "0"),
         rw!("one-mul"       ; "(* 1 ?a)"                => "?a"),
-        rw!("mul-one"       ; "(* ?a 1)"                => "?a"),
         rw!("one-div"       ; "(/ ?a 1)"                => "?a"),
         rw!("cancel-div"    ; "(/ ?a ?a)"               => "1" if is_not_zero("?a")),
         // + & *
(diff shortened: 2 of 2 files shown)


quil-rs/src/expression/simplification.rs
```diff
@@ -476,6 +476,8 @@ static RULES: Lazy<Vec<Rewrite>> = Lazy::new(|| {
         // + & *
         rw!("distribute"    ; "(* ?a (+ ?b ?c))"        => "(+ (* ?a ?b) (* ?a ?c))"),
         rw!("factor"        ; "(+ (* ?a ?b) (* ?a ?c))" => "(* ?a (+ ?b ?c))"),
+        rw!("commute-add"   ; "(+ (+ ?a ?b) ?c)"        => "(+ ?a (+ ?b ?c))"),
```
BatmanAoD, line 479: Nit: this is associativity, not commutativity. Commutativity would be `(+ ?a ?b)` => `(+ ?b ?a)`.

BatmanAoD (approved):

quil-rs/src/expression/mod.rs
```diff
@@ -1046,6 +1046,13 @@ mod tests {
             ("pi", Expression::Number(PI.into())),
             ("pi/2", Expression::Number((PI / 2.0).into())),
             ("pi * pi", Expression::Number((PI.powi(2)).into())),
+            (
+                "(a[0]*2*pi)/6.283185307179586",
```
genos, line 1050: 👍

quil-rs/src/expression/simplification.rs
genos: ⬆️

genos: Same below

quil-rs/src/expression/simplification.rs
```diff
@@ -563,4 +565,22 @@ mod tests {
         &RULES,
         "(^ (neg 9.48e42i) A[9])" => "(^ -9.48e42i A[9]))"
     }
+
+    egg::test_fn! {
```
genos, line 569: Great tests!

genos (commented):
Agree with @BatmanAoD; `(◊ (◊ ?a ?b) ?c) == (◊ ?a (◊ ?b ?c))` (where `◊` here is one of `{+, *}`) is associativity, not commutativity. Other than the rule naming, LGTM!

If you do put in commutativity with `(◊ ?a ?b) => (◊ ?b ?a)`, you could conceivably drop a number of other rules; for instance, we'd arguably only need one of `"add-zero"` and `"zero-add"`, as we'd get the other one "for free."

quil-rs/src/expression/mod.rs
Shadow53: Does this also cover `2*pi*a[0]`? Also curious if this can be generalized into `a[0]*x/x`

genos: I think there may be a way to put that together with the current rules + the changes here, but maybe adding commutativity (& trimming the ruleset, see my previous comment) might make sense.

As for your `a[0]*x/x` example, I think the `"div-mul-2"` rule covers it as long as `x` is nonzero:

```
rw!("div-mul-2" ; "(/ (* ?b ?a) ?a)" => "?b" if is_not_zero("?a")),
```

quil-rs/src/expression/simplification.rs
kalzoo: doh. fixed

quil-rs/src/expression/simplification.rs
kalzoo: thanks!

BatmanAoD (approved):

BatmanAoD:
We're gettin' all the axioms in here

Issue #229, opened 2023-06-20
Add owned Programs?


Shadow53:
`src/program/mod.rs` implements `Add` for `&Program` as such:

```rust
impl<'a, 'b> ops::Add<&'b Program> for &'a Program {
    type Output = Program;

    fn add(self, rhs: &'b Program) -> Program {
        let mut new = self.clone();
        new.calibrations.extend(rhs.calibrations.clone());
        new.frames.merge(rhs.frames.clone());
        new.memory_regions.extend(rhs.memory_regions.clone());
        new.waveforms.extend(rhs.waveforms.clone());
        new.instructions.extend(rhs.instructions.clone());
        new
    }
}
```

Was there a specific reason for this? Seems like it would be more efficient to implement adding two owned values and require them to be cloned at the call side instead, if necessary.

MarquessV:
This was originally motivated by needing to expose an __add__ method for pyQuil, where the programs are stored references. There's no reason I can think of to not clone them at the call site instead, so I think you're good to make this change.

MarquessV:
Closed via #249

PR #249, opened 2023-07-11
fix!: When adding two Programs, the resulting Program will have a correct used qubit cache.

### LINKED ISSUE #248: Adding programs doesn't update the used_qubits cache
When adding two programs, the `used_qubits` cache isn't merged and set in the new program, which can lead to an invalid cache state when two programs are added together.

MarquessV:
Closes #248 and #229


Shadow53 (approved):
LGTM

PR #253, opened 2023-07-12
Specify minimum rust version to improve compile-failure error message

BatmanAoD:
We recently added code that only compiles with Rust 1.70, leading to a confusing error message ([reported in Slack](https://rigetti.slack.com/archives/C02BBP9A60J/p1689182934952119)).

This sets the minimum required Rust version to 1.70 for `quil-rs`. For now, no change is made to `quil-py`. If we frequently break Rust version compatibility for users, we should consider using a tool to automate setting the version; [this one](https://github.com/foresterre/cargo-msrv) looks promising.


kalzoo:
> [this one](https://github.com/foresterre/cargo-msrv) looks promising.

That's cool! Agreed

kalzoo (approved):

PR #260, opened 2023-07-19
feat!: allow overriding Instruction getters

Shadow53:
This adds a couple simple getters that are needed for an internal project and should be generally useful for any consumers of `quil-rs`.

Edit: this expanded from the initial PR, but with the same goal of allowing custom handling of instructions when scheduling, especially for `PRAGMA` instructions.

### DIFF
--- a/quil-rs/src/program/graph.rs
+++ b/quil-rs/src/program/graph.rs
@@ -26,7 +26,6 @@ use crate::instruction::{
 };
 use crate::{instruction::InstructionRole, program::Program};
 
-use super::frame::MatchedFrames;
 pub use super::memory::MemoryAccessType;
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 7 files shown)


quil-rs/src/program/graph.rs
```diff
@@ -587,9 +599,7 @@ impl<'a> ScheduledProgram<'a> {
                 | Instruction::Declaration(_)
                 | Instruction::GateDefinition(_)
                 | Instruction::FrameDefinition(_)
-                | Instruction::MeasureCalibrationDefinition(MeasureCalibrationDefinition {
-                    ..
```
BatmanAoD, line 591: Any idea why this variant was written differently from all the others?

quil-rs/src/program/graph.rs
```diff
@@ -329,11 +329,13 @@ impl<'a> InstructionBlock<'a> {
                     Ok(())
                 }
                 InstructionRole::RFControl => {
-                    let matched_frames = program.get_frames_for_instruction(instruction);
+                    let matched_frames: Option<MatchedFrames<'_>> =
```
BatmanAoD, line 332: Does the custom handler make the explicit type here necessary, or did you add it for some other reason?

BatmanAoD (approved):

quil-rs/src/program/graph.rs
Shadow53: No idea, probably just leftover from previous development.

PR #264, opened 2023-07-24
feat(py): Add setters for `Program` calibrations, waveforms, frames, and memory_regions

### LINKED ISSUE #263: Python: Add setters for Program fields
The PyProgram bindings do not include setters for the various public fields on a Program

MarquessV:
closes #263


kalzoo:
genuine question - is this a good thing to offer? My impression of those fields within the Rust library was that they were just views into instructions which had already been added, but maybe that's moved since I first wrote it.

MarquessV:
> genuine question - is this a good thing to offer? My impression of those fields within the Rust library was that they were just views into instructions which had already been added, but maybe that's moved since I first wrote it.

I'm not sure what the intent was, but I believe these fields have always been public. I think they count as more than a view too, since they actually maintain the state of the program, it's not just a query into the program. As it exists today, the fields being public expose all the mutable methods in `CalibrationSet` and `FrameSet`, which I assume were added because they are useful in some way. In `pyQuil`s `Program` at least, it's not uncommon to operate on these properties, and I've learned it's useful to be able to think of these things as in different "buckets". I think the ultimate target would be a `Program` that really is just instructions, with a configurable way of getting and/or mutating a "bucket" of instructions in the program defined by a filter of some sort. This gets us out of the business of defining what those buckets are. That said, `CalibrationSet` and `FrameSet` have specialized purposes that would probably be missed. 

In absence of revamping all the access patterns `quil-rs` currently uses, then I think we need a way to modify them. For example, say we want to remove a single calibration. If the field was private, then the workaround requires a fair amount of boilerplate. Create a new `Program`, iterate over every instruction in the original program, adding each one to the new `Program` except the calibration you want to skip. Though I'm generally hesitant to make fields public, I'm not sure what clamping down on the visibility gets us as long as we make sure `CalibrationSet` and the like can be constructed and modified in a sound way. Having them as a writable field on a `Program` gives the type more modularity. I can see the case that the public fields we have today could be generically useful across multiple programs, and making them public allows you to mix and match between programs.

PR #266, opened 2023-07-26
feat!: add qubit and label placeholders

kalzoo:
Supersedes #164 , which got way out of sync following crate reorganization. This is a complete re-do which re-uses much of the code.

BREAKING: New variant Qubit::Placeholder
BREAKING: Label is now an enum, not a struct
BREAKING: std::fmt::Display no longer implemented for `Program`, `Instruction`, `Expression`, or the various instruction types.

Guide for reviewers:

* New `Quil` trait:
  * `write(impl Write, fall_back_to_debug bool)` allows for either a fallible write or an "infallible" one. I quote "infallible" because it can still fail with a `std::fmt::Result`; it's only infallible in that _quil syntax errors_ should not cause it to fail in writing. That's because any component which _would_ fail to write now instead writes the debug representation of whatever can't be represented as valid Quil. Today (this PR), that's only qubit and label placeholders.
  * `to_quil` exposes that fallible branch and will return an error if any part of its receiver cannot be formatted as valid Quil
  * `to_quil_or_debug` exposes that infallible branch and returns a String in all cases. **Note: this takes after the Rust core implementation of `to_string` on `Display`, which _also_ ignores `std::fmt::Result`.
* `Display` implementations have been removed for _almost_ everything, so as not to confuse users who may expect `thing.to_string()` to return valid Quil. This is following discussions with @Shadow53 and @MarquessV . They are not (yet) removed for some of the `Expression` subtypes, because Expressions feel like a compartment within Quil itself. I would have preferred to leave `impl Display for Expression` alone as well, but it contains a `MemoryReference` which now also does not implement `Display`. `MemoryReference` stringification is not itself fallible so it may make sense to restore `Display` on it; that's used in a few places today.
* ~~Note the new `impl_quil!` macro, which steals from the `Display` implementation. This means some inefficiency in Quil output in the form of String allocation and dropping, but it makes for neater code. I see a way to rework this so that's avoided, by pulling the `fmt` body into the macro invocation. This is the key difference from #164.~~
* ~~`impl_quil_for_ref!` handles the case where we need `Quil` for `&T` and not just `T`, in the case of iterators~~
* `Quil::to_quil_or_debug` helps with the case where we need a string infallibly, such as in error messages. It would be better if it were recursive via a `write_quil_or_debug(...) -> ()`, so then only subsets of the object would fall back on Debug rather than the whole top-level object (such as a `Program`)

Questions:

* Does it make sense for us to use `Debug` here? That's a Rust-specific concept. Should we have a different fallback for placeholders?

I am still uncomfortable with this API, which seems to compact two different concerns into one set of types:

- A Quil program builder, for which placeholders are a key component, but which is not necessarily valid Quil
- A Quil program literal representation, which represents a valid Quil program and in which placeholders are not valid.

Packing the two together into `Program` and `Instruction`, etc, makes for some awkward situations, such as:

- the parser parses a `Qubit`, but now that must be handled as if it may be a `Qubit::Placeholder`, which it cannot be (that is not parseable Quil, it has no Quil representation). Same for `Label`, which now has to be an `enum` even though it is naturally just a `String` wrapper in Quil.

TODO:

- [x] `quil-py` implementation
- [x] ~~There's a flaky test, `program::tests::placeholder_replacement`, which appears to depend on iteration order. I've replaced the containers that _should_ affect that order with `IndexSet`, which fixed `LabelPlaceholder` stability but not `QubitPlaceholder`. If you can spot the bug, let me know! I'll give it a fresh look when I can.~~ Straightforward with fresh eyes; it needed `get_used_qubits` to be stably ordered. Rather than change that return type, I just have `resolve_placeholders` sort prior to iteration. Opinions welcome.


quil-rs/src/instruction/mod.rs
```diff
@@ -186,162 +224,165 @@ impl From<&Instruction> for InstructionRole {
...
 
-fn write_qubits(f: &mut fmt::Formatter, qubits: &[Qubit]) -> fmt::Result {
+fn write_qubits(
+    f: &mut impl std::fmt::Write,
+    fall_back_to_debug: bool,
+    qubits: &[Qubit],
+) -> crate::quil::ToQuilResult<()> {
     for qubit in qubits {
-        write!(f, " {qubit}")?;
+        write!(f, " ")?;
```
Shadow53, line 270: Maybe a doc comment indicating that this produces a leading space?

quil-rs/src/instruction/mod.rs
```diff
@@ -560,6 +604,44 @@ impl Instruction {
         }
     }
 
+    /// Return immutable references to the [`Qubit`]s contained within an instruction
+    /// TODO: replace the logic in Program::get_used_qubits with this
```
Shadow53, line 608: TODO

quil-rs/src/program/mod.rs
```diff
@@ -348,6 +360,90 @@ impl Program {
...
+
+            for qubit in sorted_qubits {
+                match qubit {
+                    Qubit::Fixed(index) => {
+                        qubits_used.insert(*index);
+                    }
+                    Qubit::Placeholder(placeholder) => {
+                        qubit_placeholders.insert(placeholder.clone());
+                    }
+                    _ => {}
```
Shadow53, line 393: I would prefer specifying the specific remaining variant here instead of `_`.

quil-rs/src/program/mod.rs
```diff
@@ -348,6 +360,90 @@ impl Program {
...
+                }
+            }
+
+            let mut qubit_iterator = (0u64..).filter(|index| !qubits_used.contains(index));
+            let qubit_resolutions: HashMap<QubitPlaceholder, u64> = qubit_placeholders
+                .into_iter()
+                .map(|p| (p, qubit_iterator.next().unwrap()))
+                .collect();
```
Shadow53, line 401:
```suggestion
            let qubit_resolutions: HashMap<QubitPlaceholder, u64> = qubit_placeholders
                .into_iter()
                .zip(0u64..)
                .collect();
```

quil-rs/src/quil.rs
```diff
@@ -0,0 +1,71 @@
...
+        self.write(&mut buffer, false)?;
+        Ok(buffer)
+    }
+
+    /// Return a string in valid Quil syntax if possible. Any individual component of this object
+    /// which cannot be represented in Quil will be replaced with a `Debug` representation of that
+    /// component.
+    fn to_quil_or_debug(&self) -> String
+    where
+        Self: std::fmt::Debug,
```
Shadow53, line 19: Having the `Debug` restriction on the method feels weird. I would prefer it on the trait itself rather than the method:

```suggestion
pub trait Quil: std::fmt::Debug {
    /// Return a string in valid Quil syntax or an error if the item cannot be represented with valid Quil.
    fn to_quil(&self) -> Result<String, ToQuilError> {
        let mut buffer = String::new();
        self.write(&mut buffer, false)?;
        Ok(buffer)
    }

    /// Return a string in valid Quil syntax if possible. Any individual component of this object
    /// which cannot be represented in Quil will be replaced with a `Debug` representation of that
    /// component.
    fn to_quil_or_debug(&self) -> String
```

quil-rs/src/program/mod.rs
kalzoo: Not equivalent, because of the lack of filter. Could still zip the qubit iterator though

quil-rs/src/program/mod.rs
```diff
@@ -348,6 +360,90 @@ impl Program {
...
+    /// If your placeholder returns `None` for a particular placeholder, it will not be replaced but will be left as a placeholder.
+    ///
+    /// If you do not provide a resolver for a placeholder, a default resolver will be used which will generate a unique value
+    /// for that placeholder within the scope of the program using an auto-incrementing value (for qubit) or suffix (for label)
+    /// while ensuring that unique value is not already in use within the program.
+    #[allow(clippy::type_complexity)]
+    pub fn resolve_placeholders(
+        &mut self,
+        label_resolver: Option<Box<dyn Fn(&LabelPlaceholder) -> Option<String>>>,
+        qubit_resolver: Option<Box<dyn Fn(&QubitPlaceholder) -> Option<u64>>>,
```
MarquessV, line 376: Thoughts on splitting this up into two methods?

```rs
resolve_placeholders(&mut self)

resolve_placeholders_with_resolvers(
    &mut self,
    label_resolver: Box<dyn Fn(&LabelPlaceholder)> -> Option<String>>>,
    qubit_resolver: Box<dyn Fn(&QubitPlaceholder)> -> Option<u64>>>)
)
```

Some benefits of this:

1. The default behavior (`resolve_placeholders(None, None)`) seems like it would be a  common way to use this method and splitting it allows it to be used without specifying two `None`s each time.
2. Eliminates optional parameters where `None` actually means use some default behavior.
3. Less likely, but the return type for `qubit_resolver` is `Option<u64>`, if we ever change the type for a fixed qubit we'd have to break the signature. If we split them out, then at least users using the "default" one won't be broken.

quil-rs/src/instruction/calibration.rs
```diff
@@ -104,7 +109,7 @@ mod test_measure_calibration_definition {
         insta::with_settings!({
             snapshot_suffix => description,
         }, {
-            assert_snapshot!(measure_cal_def.to_string())
+            assert_snapshot!(measure_cal_def.to_quil_or_debug())
```
MarquessV, line 112: If this test suite isn't going to have a placeholder case, should this just be `to_quil()`?

quil-rs/src/instruction/gate.rs
```diff
@@ -72,14 +72,17 @@ pub enum GateError {
         parameters: Vec<Expression>,
     },
 
-    #[error("cannot produce a matrix for gate `{name}` with variable qubit {qubit}")]
+    #[error("cannot produce a matrix for gate `{name}` with variable qubit {qubit}", qubit=.qubit.to_quil_or_debug())]
```
MarquessV, line 75: `qubit=.qubit.to_quil_or_debug()`

Is that extra period a mistake, or special thiserror syntax?

kalzoo: It's correct, that's `thiserror`

quil-rs/src/instruction/calibration.rs
kalzoo: I thought this might be easier to debug - it'll just show up in the diff rather than a terminal dump with `unwrap()`

quil-rs/src/program/mod.rs
kalzoo: Yep, on board with this, except that your new signature doesn't allow for one resolver to be provided without the other. Is that intended?

MarquessV: My thought was that there could be a default implementation that a user could pull in and explicitly use instead of using `None` resolving to that default implementation. Could also just split this into another set of methods: `resolve_placeholders_with_qubit_resolver` and ``resolve_placeholders_with_label_resolver`

kalzoo: Ah, I see - we could indeed vend those defaults outside of the function body. That makes sense. Then the user can mix and match and wrap.

Otherwise - from your note - do you think we should replace the `Option<u64>` return type with `Option<Qubit>`

MarquessV: I think `u64` is _probably_ fine. I'm not sure if getting a placeholder for a variable qubit is particularly useful (pyQuil at least only supports replacing placheolder with indexed qubits). I wish we could restrict the type to `Qubit::Fixed` specifically, but `u64` seems unlikely to ever be incompatible with a fixed qubit. Changing the signature to `Option<Qubit>` opens up the user to replacing a placeholder with a placeholder, which would be invalid, so it comes with it's own problems.

MarquessV:
Updates:

1. Replaced implementations of `__str__` on instruction types with `to_quil`
2. Fixed a bug with `LabelPlaceholders` only being resolved on `Label` instructions, when they should be resolved in instructions containing label too (i.e. `JUMP`).
3. Fixed a similar bug with `Program.get_labels`, and by consequence, the default label resolver w/ a test to prevent regression.
4. Renamed the private `label` module to `control_flow` and added the `Jump*` instructions to it.
5. Expose `default_label/qubit_resolver`s so `resolve_placeholders_with_custom_resolvers` doesn't need to take `Option` params anymore.
6. Python bindings to all control flow instructions, placeholder variants, and related methods.

quil-py/src/program/mod.rs
```diff
@@ -218,8 +225,75 @@ impl PyProgram {
             .to_owned())
     }
 
-    pub fn __str__(&self) -> String {
-        self.as_inner().to_string()
+    pub fn resolve_placeholders(&mut self) {
+        self.as_inner_mut().resolve_placeholders();
+    }
+
+    // TODO: Pre-resolve placeholders into a map so a non-panic error can be raised?
```
MarquessV, line 232: If the given function raises an exception or returns an incompatible type, we can't bubble up an error from the resolver function, which is why the current implementation uses `assert!`. Because these are pyo3 bindings, that isn't as bad as it would be in a pure Rust context. `pyo3` wraps Rust panics in a `PanicException`, that gets raised like any other exception. The context string from the `assert!` gets included in the exception too, so the error is actually pretty clear if you can get past the `PanicException`.

Alternatively, we can use the given function to populate a pair of `HashMap` for the placeholders and use the resulting `HashMap`s to build the resolver functions. If populating the `HashMap`s fail for any reason we can bubble up the error in the usual `PyResult`. This requires methods that don't exist on `Program` yet, and I'm not sure if it would make sense to make such methods public either.

quil-rs/src/program/mod.rs
MarquessV: I added those defaults and removed the option from the signature.

quil-rs/src/quil.rs
```diff
@@ -0,0 +1,68 @@
...
+    joiner: &str,
+    prefix: &str,
+) -> Result<(), ToQuilError>
+where
+    I: IntoIterator<Item = &'i T>,
+    T: Quil + 'i,
+{
+    let mut iter = values.into_iter();
+    if let Some(first) = iter.next() {
+        write!(writer, "{}", prefix)?;
```
notmgsk, line 59:
```suggestion
        write!(writer, "{prefix}")?;
```

quil-rs/src/quil.rs
```diff
@@ -0,0 +1,68 @@
...
+pub enum ToQuilError {
+    #[error("Failed to write Quil: {0}")]
+    FormatError(#[from] std::fmt::Error),
+    #[error("Label has not yet been resolved")]
+    UnresolvedLabelPlaceholder,
+    #[error("Qubit has not yet been resolved")]
+    UnresolvedQubitPlaceholder,
+}
+
+/// Write a sequencer of Quil items to the given writer, joined with the provided `joiner`.
```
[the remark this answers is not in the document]
notmgsk: sequencers on the mind

quil-py/src/program/mod.rs
```diff
@@ -234,12 +308,12 @@ impl PyProgram {
         }
     }
 
-    // This is infallible now, but will raise an error once placeholders are
-    // supported. This is because placeholders can't be converted to valid Quil,
-    // nor be reliably serialized by something like serde using the current
-    // quil-rs data model.
-    pub fn __getstate__<'a>(&self, py: Python<'a>) -> &'a PyBytes {
-        PyBytes::new(py, self.as_inner().to_string().as_bytes())
+    // This will raise an error if the program contains placeholders any
```
notmgsk, line 311:
```suggestion
    // This will raise an error if the program contains any
```

quil-py/src/program/mod.rs
```diff
@@ -218,8 +225,75 @@ impl PyProgram {
...
+        label_resolver: Option<Py<PyFunction>>,
+        qubit_resolver: Option<Py<PyFunction>>,
+    ) {
+        #[allow(clippy::type_complexity)]
+        let rs_qubit_resolver: Box<dyn Fn(&QubitPlaceholder) -> Option<u64>> =
+            if let Some(resolver) = qubit_resolver {
+                Box::new(move |placeholder: &QubitPlaceholder| -> Option<u64> {
+                    Python::with_gil(|py| {
+                        let resolved_qubit =
+                            resolver.call1(py, (placeholder.to_python(py).unwrap(),));
```
notmgsk, line 245: Any concern about using `unwrap` here (and elsewhere)? or is this covered by your comment elsewhere about pyo3 wrapping this in a nice Python exception?

quil-py/src/program/mod.rs
```diff
@@ -218,8 +225,75 @@ impl PyProgram {
...
+                        assert!(
+                            resolved_qubit.is_ok(),
+                            "qubit_resolver returned an error: {resolved_qubit:?}"
+                        );
+
+                        let resolved_qubit: PyResult<Option<u64>> = resolved_qubit
+                            .expect("asserted that resolved_qubit is ok")
```
notmgsk, line 252: I suppose you can just `unwrap()` here since, if there is a problem, it will have been caught above (i.e. this `expect` will never fail).

quil-py/src/program/mod.rs
notmgsk: Admittedly, I don't quite understand your proposed alternative, but if the current implementation gives a useful Python exception I would say that's enough.

quil-rs/src/program/mod.rs
```diff
@@ -348,6 +356,108 @@ impl Program {
...
+                    next_suffix += 1;
+                }
+            })
+            .collect();
+
+        Box::new(move |key| label_resolutions.get(key).cloned())
+    }
+
+    /// The default qubit resolver will resolve each [`QubitPlaceholder`] in the program to
+    /// a unique fixed qubit by incrementing to the next available index.
```
notmgsk, line 431:
```suggestion
    /// a unique fixed qubit index by incrementing to the next available index.
```

quil-rs/src/instruction/qubit.rs
```diff
@@ -1,19 +1,122 @@
...
+                } else {
+                    Err(ToQuilError::UnresolvedQubitPlaceholder)
+                }
+            }
+            Variable(value) => write!(writer, "{value}").map_err(Into::into),
         }
     }
 }
+
+type QubitPlaceholderInner = Arc<()>;
```
Shadow53, line 46: The comments on [this StackOverflow question](https://stackoverflow.com/questions/54658894/how-do-i-create-a-unique-id-for-a-rust-object) seem to indicate that this is a problematic way to do things.

The better option, I think, is using [`AtomicUsize`](https://doc.rust-lang.org/stable/std/sync/atomic/struct.AtomicUsize.html) or [`AtomicU64`](https://doc.rust-lang.org/stable/std/sync/atomic/struct.AtomicU64.html) as a (semi-)global incrementor for integer ids.

quil-rs/src/program/mod.rs
```diff
@@ -348,6 +356,110 @@ impl Program {
...
+    /// If your placeholder returns `None` for a particular placeholder, it will not be replaced but will be left as a placeholder.
+    ///
+    /// If you wish to provide a resolver for either labels or qubits, but want to rely on the
+    /// default behavior for the other, considering using either
+    /// [`default_qubit_resolver`](Self::default_qubit_resolver) or [`default_target_resolver`](Self::default_target_resolver).
+    #[allow(clippy::type_complexity)]
+    pub fn resolve_placeholders_with_custom_resolvers(
+        &mut self,
+        target_resolver: Box<dyn Fn(&TargetPlaceholder) -> Option<String>>,
+        qubit_resolver: Box<dyn Fn(&QubitPlaceholder) -> Option<u64>>,
```
Shadow53, line 384: Is there a specific reason for using `Box<dyn Fn()>` instead of `impl Fn()`?

quil-rs/src/program/mod.rs
```diff
@@ -348,6 +356,110 @@ impl Program {
...
+    ) {
+        for instruction in &mut self.instructions {
+            instruction.resolve_placeholders(&target_resolver, &qubit_resolver);
+        }
+    }
+
+    /// The default target resolver will resolve each [`TargetPlaceholder`] in the program to a unique target
+    /// by applying an auto-incrementing suffix to the base target.
+    #[allow(clippy::type_complexity)]
+    pub fn default_target_resolver(&self) -> Box<dyn Fn(&TargetPlaceholder) -> Option<String>> {
```
Shadow53, line 394: Same question here re: `Box` vs `impl`

quil-rs/src/program/mod.rs
```diff
@@ -348,6 +356,108 @@ impl Program {
...
+            .map(|p| {
+                let base_label = p.as_inner();
+                let mut next_suffix = 0;
+
+                loop {
+                    let next_label = format!("{base_label}_{next_suffix}");
+
+                    if !fixed_labels.contains(&next_label) {
+                        fixed_labels.insert(next_label.clone());
+                        break (p, next_label);
```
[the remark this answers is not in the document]
Shadow53: I don't think there is. IMO `return` would be clearer here.

Also, it might be clearer to do something like

```rust
for next_suffix in 0.. {
    let next_label = ...;
    if !fixed_labels.contains(&next_label) {
        // ...
    }
}
```

quil-rs/src/program/mod.rs
```diff
@@ -348,6 +356,110 @@ impl Program {
...
+                    Qubit::Variable(_) => {}
+                }
+            }
+        }
+
+        let mut qubit_iterator = (0u64..).filter(|index| !qubits_used.contains(index));
+        let qubit_resolutions: HashMap<QubitPlaceholder, u64> = qubit_placeholders
+            .into_iter()
+            .map(|p| (p, qubit_iterator.next().unwrap()))
+            .collect();
```
Shadow53, line 458:
```suggestion
        let qubit_resolutions: HashMap<QubitPlaceholder, u64> = qubit_placeholders
            .into_iter()
            .zip(qubit_iterator)
            .collect();
```

quil-rs/src/quil.rs
```diff
@@ -0,0 +1,68 @@
...
+        self.write(&mut buffer, false)?;
+        Ok(buffer)
+    }
+
+    /// Return a string in valid Quil syntax if possible. Any individual component of this object
+    /// which cannot be represented in Quil will be replaced with a `Debug` representation of that
+    /// component.
+    fn to_quil_or_debug(&self) -> String {
+        let mut buffer = String::new();
+        self.write(&mut buffer, true).ok();
```
Shadow53, line 19: I believe this syntax is preferred for ignoring a result

```suggestion
        let _ = self.write(&mut buffer, true);
```

quil-rs/src/quil.rs
```diff
@@ -0,0 +1,68 @@
...
+        buffer
+    }
+
+    /// Write the Quil representation of the item to the given writer.
```
Shadow53, line 23: Should have a note here that it _must not_ error if `fall_back_to_debug` is `true`.

quil-rs/src/quil.rs
```diff
@@ -0,0 +1,68 @@
+/// A trait to wrap items which represent some construct within the Quil language.
+///
+/// If you want to serialize an object to string and fail if it can't be represented as valid Quil, then use
+/// `to_quil()`. If you want to serialize an object to string infallibly, and can tolerate invalid Quil, then
+/// use `to_quil_or_debug()`.
+pub trait Quil: std::fmt::Debug {
```
Shadow53, line 6: We should consider using a [sealed trait](https://rust-lang.github.io/api-guidelines/future-proofing.html) to prevent issues with library consumers implementing faulty versions of `Quil` for their custom types. The [`sealed`](https://crates.io/crates/sealed) crate seems popular, or it can be done manually

quil-rs/src/quil.rs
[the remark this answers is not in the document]
Shadow53: Maybe the intent was `sequence`?

Shadow53 (changes requested):

quil-rs/src/program/mod.rs
MarquessV: Not sure, @kalzoo do you remember why you went this route?

quil-rs/src/instruction/qubit.rs
MarquessV: Can you expand on how it might be problematic? I can see how the question is similar, but it seems different in that they are using the address of a value, where we are using the address to a value an Arc points to. For an `Arc` to work, I don't see how that could change without every copy of the `Arc` being able to see that change, so I think we are good. The [docs on `ptr_eq`](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.ptr_eq) do callout a gotcha with `dyn Trait`, but that doesn't apply here.

quil-rs/src/program/mod.rs
```diff
@@ -281,6 +275,20 @@ impl Program {
             .map(|condition| self.frames.get_matching_keys_for_conditions(condition))
     }
 
+    /// Return references to all targets used in the program.
+    fn get_targets(&self) -> Vec<&Target> {
+        self.instructions
```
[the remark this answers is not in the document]
MarquessV: Possibly to have deterministic ordering for a given program. Maybe something different? @kalzoo would know more. Could use a `BTreeSet` if we add `PartialOrd`, `Ord` to `Target`.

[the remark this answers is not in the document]
MarquessV: Thinking about it more, building a `Vec` in this way ensures the target suffixes are incremented sequentially based on the order of appearance in the program, which will likely be easier for users to follow.

quil-rs/src/quil.rs
MarquessV: Any reason? Other than not doing the work to convert it to an Option?

quil-rs/examples/generate_test_expressions.rs
```diff
@@ -66,7 +67,7 @@ fn parenthesized(expression: &Expression) -> String {
             operator,
             parenthesized(right)
         ),
-        Number(_) => format!("({})", expression),
+        Number(_) => format!("({})", expression.to_quil_or_debug()),
```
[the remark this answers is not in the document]
MarquessV: `n` would be a `Complex64`, which doesn't have a `Quil` impl. I believe the intent is to just add the parentheses correctly, but defer to `Expression` for actually formatting what goes inside of it.

quil-py/src/program/mod.rs
MarquessV: Yup, this would also be covered by pyo3 wrapping mechanisms for panics. However, I will change this one to an `expect` because the expectation for it to not fail isn't obvious, but `to_python` here should never fail for a `QubitPlaceholder`

quil-rs/src/program/mod.rs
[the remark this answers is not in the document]
MarquessV: Took a different approach that I think expresses the sentinel condition the most clearly.

quil-rs/src/quil.rs
MarquessV: Will consider and come back to this.

Shadow53 (approved):
Looks good! I feel like there's a better way to generate new labels, but I don't want to bikeshed on that.

quil-py/src/program/mod.rs
```diff
@@ -218,8 +229,72 @@ impl PyProgram {
...
+    ) {
+        #[allow(clippy::type_complexity)]
+        let rs_qubit_resolver: Box<dyn Fn(&QubitPlaceholder) -> Option<u64>> =
+            if let Some(resolver) = qubit_resolver {
+                Box::new(move |placeholder: &QubitPlaceholder| -> Option<u64> {
+                    Python::with_gil(|py| {
+                        let resolved_qubit = resolver
+                            .call1(
+                                py,
+                                (placeholder
```
jselig-rigetti, line 254: This is really cool! I wonder what happens if the user-provided function calls back into rust, So Py->Rust->Py->Rust... - would there be GIL deadlock causing `Python::with_gil` to block?

jselig-rigetti (commented):
Just a few questions, but overall very nice!

quil-py/quil/expression/__init__.pyi
```diff
@@ -129,12 +120,22 @@ class Expression:
...
+        """
+        Attempt to convert the instruction to a valid Quil string. Raises
+        an exception if the instruction can't be converted to valid Quil.
+        """
+        ...
+    def to_quil_or_debug(self) -> str:
+        """
+        Convert the instruction to a Quil string. If any part of
+        the instruction can't be converted to valid Quil, a debug
+        implementation will be used.
```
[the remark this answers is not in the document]
MarquessV: Yeah, I think I see what you are saying. It does come immediately after `if any part of the instruction can't be converted to _valid_ Quil`, but in hindsight debug implementation is a bit vague so I updated it.

quil-py/src/program/mod.rs
MarquessV: Good question! The pyQuil counterpart to this PR actually does just that and nothing locks up.

Issue #259, opened 2023-07-14
Hashing Expressions should be fast


kalzoo:
This expression appears to take over 2 minutes to hash on a decent laptop using the default Rust SipHasher:

```
[...]

PR #267, opened 2023-07-27
Clean up simplification rules

genos:
This came about from messing around with #259. Unfortunately, I couldn't conquer the pathologically large expression in that issue, but I did emerge (after much theft from [`herbie`](https://github.com/herbie-fp/herbie/blob/main/egg-herbie/src/rules.rs)) with a cleaner and condensed simplification ruleset. Note: this adds `velcro` as a dependency for the convenience of "splatting" vectors with the beefed-up `vec![ .. ]` macro for bidirectional rewrite rules.


quil-rs/src/expression/simplification.rs
```diff
@@ -462,43 +463,58 @@ static RULES: Lazy<Vec<Rewrite>> = Lazy::new(|| {
         // Largely copied from https://github.com/egraphs-good/egg/blob/82c00e970f0bc1fbfe90ce6dc3c3c79ee919c933/tests/math.rs
         // and https://github.com/herbie-fp/herbie/blob/2052806f2ffe0d46bc2e151dd096b127c39e12bd/egg-herbie/src/rules.rs
 
+        // pos and neg
+        rw!("pos canon"     ; "(pos ?a)"                => "?a"),
```
kalzoo, line 467: What do you think about removing `pos` (and perhaps `-` subtraction) in the name of simplifying both the e-graphs and the ruleset?

kalzoo:
The changes look good, but we are somewhat in the dark on performance impact or boost here, right?

kalzoo (approved):

genos:
> … we are somewhat in the dark on performance impact or boost here, right?

Yes indeed. Test cases all passed, as did some (non-pathological) extra simplification tests I had around at the time. But I don't have before/after deltas to examine, sorry.

quil-rs/src/expression/simplification.rs
genos: I'd love to remove the prefix unary ops if we could, but they're part of the Quil spec, right? Either way, I think these `<X> canon` rules fire almost only at the first iteration of simplification, from anecdotally looking at `RUST_LOG=egg=debug` runs of things, and then both `pos` and `neg` are pretty much never heard from again—barring when something simplifies a great deal.

genos: Oh! But if we change how `Expression`s get turned into `RecExpr<Expr>`s, we of course could simplify the `Expr` language, couldn't we? That'd be nice.

genos:
Per internal discussion, this should wait until we have favorable data from benchmarking in #270.

genos:
No longer needed after #273.

PR #268, opened 2023-07-28
chore: test that similar calibrations with different sites are not filtered

### LINKED ISSUE #192: Add test cases for calibration filtering
To prevent a re-regression of [pyQuil#1561](https://github.com/rigetti/pyquil/issues/1561) we should build up test suites that make sure calibration filtering work as expected. In particular, calibrations should only be recognized as duplicates if the name, parameters _and_ qubits match of the `DEFCAL` match

MarquessV:
closes #192


kalzoo (approved):
prefix says `wip` but offline you said this was ready; lgtm, just needs the right commit syntax. Good thing to bring forward from pyQuil v3.

PR #272, opened 2023-08-03
fix(rs): Simplify by hand

genos:
Trying to get as much simplification of expressions done by hand without turning to `egg`.


genos:
Closing in favor of #273

Issue #27, opened 2021-08-12
Support equality between commutative expressions


dbanty:
Expressions which are equal by the commutative property should evaluate to equal and hash the same. As an example, an expression `1 + theta` should be equivalent to `theta + 1` both when compared directly (e.g. `assert_eq!`) and when hashed (e.g. added to a `HashSet`).

Per [this comment](https://github.com/rigetti/quil-rust/pull/25#issuecomment-897342062) we have some internal code in the QCS backend which creates a "canonical" or "normal" form of expressions which we can likely reuse.

genos:
Trying to dive into this, I naïvely thought, "No problem, we just want to sort the operands when dealing with a commutative infix operator!" completely forgetting that the operands are themselves `Expression`s, which would mean I'd need to `impl Ord`…

PR #276, opened 2023-08-09
wip: use arena-style cache when simplifying Expressions (DO NOT MERGE)

Shadow53:
This PR is meant as a reference and not to be merged (yet).

It is based on @genos' `simplify-by-hand` branch and attempts to improve performance by using an arena-based cache during `Expression` simplification.

Some internal benchmarks seem to show a good improvement (~17%) in runtime when the cache is actually useful, but an overhead of 1-5% when not useful (no repeated expressions). This may be conflated with the item below, which also helps performance -- I did not benchmark them separately.

One thing that may be useful enough to pull into its own PR is the change to not use hashing in the implementation of `PartialEq` for `Expression`, which also helps speed things up.


quil-rs/src/expression/mod.rs
```diff
@@ -135,15 +135,15 @@ impl Hash for Expression {
...
-                        // commutative, so put left & right in decreasing order by hash value
-                        let (a, b) = (
-                            min_by_key(&left, &right, hash_to_u64),
-                            max_by_key(&left, &right, hash_to_u64),
-                        );
-                        a.hash(state);
-                        b.hash(state);
-                    }
+                    //InfixOperator::Plus | InfixOperator::Star => {
+                    //    // commutative, so put left & right in decreasing order by hash value
```
genos, line 139: I agree that this is costly, but we put it in for #27. Are we deciding to no longer ensure that, e.g., `1 + 2 == 2 + 1`?

quil-rs/src/expression/mod.rs
```diff
@@ -274,9 +284,8 @@ impl Expression {
                 *self = Expression::Number(Complex64::from(PI));
             }
             _ => {
-                if let Ok(simpler) = simplification::run(self) {
-                    *self = simpler;
-                }
+                let temp = std::mem::replace(self, Expression::PiConstant);
```
genos, line 287: I really need to remember this trick 👍

quil-rs/src/expression/mod.rs
genos: Alternatively, should we just derive `Hash`, `PartialEq,` and `Eq` and call it a day?

genos: In fact, I think this is the way to go. #27 only handles commutativity, but not other rules that lead to equality between expressions. Since we're simplifying expressions with these rules, enforcing some but not all in the hashing & equality implementations seems wrong and wasteful. I'll open a new PR.

PR #277, opened 2023-08-10
feat!: Decouple expression hashing and equality

genos:
BREAKING: Expressions no longer use hashing for implementing equality
BREAKING: Expression hashing and equality no longer take commutativity into account

In #276, @Shadow53 noted

> One thing that may be useful enough to pull into its own PR is the change to not use hashing in the implementation of `PartialEq` for `Expression`, which also helps speed things up.

We originally put this together in #27 to ensure that equality held in the face of commutativity, e.g., `1 + x == x + 1`. In addition to the performance benefits of decoupling the hashing and equality implementations, it makes sense to remove any special status for commutative operations in light of all the work we're doing on expression simplification. If we wished to ensure expressions are `Eq` if and only if they represent the same mathematical expression, we'd have to have equality contingent upon simplification, which would be even **more** costly.


kalzoo (approved):

kalzoo:
Good to see it!

quil-rs/src/expression/mod.rs
```diff
@@ -106,50 +108,52 @@ impl PrefixExpression {
...
 impl Hash for Expression {
     // Implemented by hand since we can't derive with f64s hidden inside.
-    // Also to understand when things should be the same, like with commutativity (`1 + 2 == 2 + 1`).
-    // See https://github.com/rigetti/quil-rust/issues/27
     fn hash<H: Hasher>(&self, state: &mut H) {
-        use std::cmp::{max_by_key, min_by_key};
-        use Expression::*;
         match self {
-            Address(m) => {
+            Expression::Address(m) => {
```
Shadow53, line 134: Small style nit, but while we're here: should use `Self` instead of `Expression`.

(It's fine not to make this change if there isn't anything more substantial being made. It's just a style thing that is triggered by a certain clippy lint)

Shadow53 (approved):
Looks good! Thanks for pulling this out of the other PR. Left a style nit, but it's fine to merge without.

quil-rs/src/expression/mod.rs
genos: Thanks! 

Yikes, I just ran `cargo clippy -- -W clippy::pedantic` 😮

PR #273, opened 2023-08-04
new(rs): fully by-hand expression simplification

genos:
~This combines the work in #267 and #272; putting together multiple passes of `by_hand(by_egg(expression))` yields dramatically simpler results.~

~`cargo test --release` on the _big_ expression in #259 yielded~

```
…<SNIP>…
yielded `-14553.919984548369+(gamma[0]*-0.6379644761225252)` instead of the expected `-0.637964476122525*gamma[0] - 14553.9199845484`', quil-rs/src/expression/simplification/mod.rs:173:5
```

~which looks pretty darn close.~

~This is just a proof of concept right now. Paying for multiple roundtrips to & from a format `egg` likes is inefficient.~

This now relies _solely_ on by-hand simplification, and does rather well:
- one test fails (`affine`, expecting `ax + b + cx + d` to simplify to `(a + c)x + (b + d)`, I can't quite get around a syntax error in trying to cover this case;
- one test hangs (the big one from #259);

The by-hand simplification has a maximum limit of times it's allowed to run, to avoid some "test `XXX` has overflowed its stack" issues stemming from infinite recursions.

PR #278, opened 2023-08-30
Don't rely on egg to simplify pi/c for constant c

mac01021:
This patch makes quil-rs simplify `pi/c` expressions without invoking `egg`.

It reduces the runtime of `expand_calibrations` on my RB program by 90%


mac01021:
It looks like maybe https://github.com/rigetti/quil-rs/pull/272 could supercede this?

genos:
Superceded by #273.

PR #280, opened 2023-09-05
feat(python): Build & publish wheels for Windows

MarquessV:
Builds and publishes wheels for x86 and x64 based windows systems. Created #281 to track creating Windows for ARM wheels when it's better supported.


MarquessV:
Successful run here: https://github.com/rigetti/quil-rs/actions/runs/6102526563/job/16561173399

quil-py/quil/expression/__init__.pyi
```diff
@@ -63,17 +63,17 @@ class Expression:
     @staticmethod
     def new_pi() -> "Expression": ...
     @staticmethod
-    def from_address(memory_reference: MemoryReference) -> "Expression": ...
+    def from_address(inner: MemoryReference) -> "Expression": ...
```
jselig-rigetti, line 66: Per https://github.com/rigetti/rigetti-pyo3/blob/b43e780532fb7b2adf0e69e4957aa7fce935c4c9/src/wrappers.rs#L583 looks like `inner` is the right name 👍

jselig-rigetti (approved):

.github/workflows/publish-quil-py.yml
```diff
@@ -81,10 +81,40 @@ jobs:
...
+          python-version: ${{ matrix.python-version }} 
+          architecture: ${{ matrix.target }} 
+      - name: Build wheels 
+        uses: messense/maturin-action@v1 
+        env:
+          RUSTFLAGS: "-Z macro-backtrace"
+        with: 
+          target: ${{ matrix.target }} 
+          rust-toolchain: nightly
+          args: -i python --release --manifest-path quil-py/Cargo.toml --out dist
```
Shadow53, line 104: Why are we using `nightly` here?

quil-py/quil/expression/__init__.pyi
MarquessV: Yup! One of the nice things about updating pyo3 is that the `__text_signature__` property is now derived automatically from the Rust function signature (or the `#[pyo3(signature = (...))]` attribute). `stubtest` checks __text_signature__ and flags any hints that don't align with it, meaning we now get `stubtest` failures when a signature uses the wrong argument names or types whereas before it really only failed if a function was missing. That's what prompted all these fixes in the type hints.

.github/workflows/publish-quil-py.yml
MarquessV: I switched to nightly so I could apply the `-Z macro-backtrace` flag and see where the x86 build was failing. I meant to switch it back, but missed it. I'll test with the stable toolchain and restore that step to how it was.

MarquessV:
Here it is passing with the stable toolchain: https://github.com/rigetti/quil-rs/actions/runs/6103513055/job/16564034952?pr=280

kalzoo (approved):

PR #283, opened 2023-09-14
feat(python): Instruction classes now implement `__copy__`, and `__deepcopy__`, making them compatible with Python's `copy` module.

MarquessV:
It's not uncommon to want to create a copy of something using Python's copy module, so this makes all instruction types compatible with it.

### DIFF
--- a/quil-py/src/instruction/mod.rs
+++ b/quil-py/src/instruction/mod.rs
@@ -1,7 +1,7 @@
 use quil_rs::instruction::Instruction;
 use rigetti_pyo3::{
     create_init_submodule, impl_repr, py_wrap_union_enum,
-    pyo3::{pyclass::CompareOp, pymethods, IntoPy, PyObject, Python},
+    pyo3::{pyclass::CompareOp, pymethods, types::PyDict, IntoPy, PyObject, PyResult, Python},
     PyWrapper,
 };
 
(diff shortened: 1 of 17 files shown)


quil-py/src/instruction/mod.rs
```diff
@@ -178,3 +204,56 @@ create_init_submodule! {
...
 }
+
+/// Implements a copy method, as well as __copy__ and __deepcopy__ for use with Python's copy
+/// module, on any type that is a variant of the PyInstruction class.
+///
+/// Both `copy` and `__deepcopy__` are deep copies. They check the instruction for
+/// [`QubitPlaceholder`]s and create new placeholders in place of them in the copy.
+///
+/// The `__copy__` method [`Clone`]s the instruction. This means resolving [`QubitPlaceholder`]s
+/// in a copy created with Python's ``copy.copy`` will also resolve them in the original.
```
MarquessV, line 215: Thoughts on this? I'm considering not providing `__copy__` at all because Python's `copy.copy(x)` should return a _shallow_ copy of x, but a `clone()` will almost always result in a deep copy.

MarquessV: After chatting offline @BatmanAoD, we learned that `Clone` on a `pyo3` object will actually just return a reference, so it can be used to create a shallow copy. I've updated the tests to check that works as we expect.

I went ahead and removed the `copy` method as well since it was redundant.

notmgsk (approved):

Issue #285, opened 2023-09-17
Extraneous whitespace in a matrix specification leads to parsing error


MarquessV:
[pyQuil#1658](https://github.com/rigetti/pyquil/issues/1658) reports an issue parsing this DEFGATE:

```
DEFGATE CPHASE00(%alpha):
    cis(%alpha), 0, 0, 0
    0,           1, 0, 0
    0,           0, 1, 0
    0,           0, 0, 1    
```

The error occurs because the extra padding is picked up by the lexer as indentation tokens, which the parser doesn't expect when processing matrix specifications.

PR #286, opened 2023-09-17
fix: Allow whitespace to delimit matrix specifications, better support parsing  identifiers followed by a minus operator.

### LINKED ISSUE #284: Identifiers minus expressions fail to parse without whitespace
[pyQuil#1658](https://github.com/rigetti/pyquil/issues/1658) reported a couple of issues with parsing. One of them occurs when parsing expressions in this DEFGATE:

```
DEFGATE CAN(%alpha, %beta, %gamma):
    (cis((%alpha+%beta-%gamma)/2)+cis((%alpha-%beta+%gamma)/2))/2, 0, 0, (cis((%alpha-%beta+%gamma)/2)-cis((%alpha+%beta-%gamma)/2))/2
    0, (cis((%alpha+%beta+%gamma)/(-2))+cis((%beta+%gamma-%alpha)/2))/2, (cis((%alpha+%beta+%gamma)/(-2))-cis((%beta+%gamma-%alpha)/2))/2, 0
    0, (cis((%alpha+%beta+%gamma)/(-2))-cis((%beta+%gamma-%alpha)/2))/2, (cis((%alpha+%beta+%gamma)/(-2))+cis((%beta+%gamma-%alpha)/2))/2, 0
    (cis((%alpha-%beta+%gamma)/2)-cis((%alpha+%beta-%gamma)/2))/2, 0, 0, (cis((%alpha+%beta-%gamma)/2)+cis((%alpha-%beta+%gamma)/2))/2
```

In particular, the expression `%beta-%gamma` fails to pass the lexer.

In Quil `-` is the minus operator and is also valid to use in an identifier, as long as it isn't the first or last character in the identifier. In this case, it should be lexed as a minus operator, but it is instead lexes it as part of the `beta` identifier. Since that results in an invalid identifier "beta-", the lexer returns an error.

MarquessV:
closes #284 and #285

### DIFF
--- a/quil-rs/src/parser/common.rs
+++ b/quil-rs/src/parser/common.rs
@@ -178,7 +178,10 @@ pub(crate) fn parse_matrix<'a>(
             token!(NewLine),
             preceded(
                 token!(Indentation),
-                separated_list0(token!(Comma), parse_expression),
+                separated_list0(
+                    pair(token!(Comma), many0(token!(Indentation))),
+                    parse_expression,
+                ),
             ),
         ),
     )(input)
@@ -577,6 +580,24 @@ SWAP-PHASES 2 3 \"xy\" 3 4 \"xy\"";
             "expected remainder to be empty, got {remainder:?}"
         );
         assert_eq!(matrix.len(), 2);
+    }
+
+    #[test]
+    fn test_parse_matrix_inline_whitespace() {
+        let input = LocatedSpan::new(
+            "
+    cis(%alpha), 0, 0, 0
+    0,           1, 0, 0
+    0,           0, 1, 0
+    0,           0, 0, 1",
+        );
+        let lexed = lex(input).unwrap();
+        let (remainder, matrix) = parse_matrix(&lexed).unwrap();
+        assert!(
+            remainder.is_empty(),
+            "expected remainder to be empty, got {remainder:?}"
+        );
+        assert_eq!(matrix.len(), 4);
     }
 
     #[test]
--- a/quil-rs/src/parser/lexer/mod.rs
+++ b/quil-rs/src/parser/lexer/mod.rs
@@ -22,7 +22,7 @@ use nom::{
     combinator::{all_consuming, map, recognize, value},
     multi::many0,
     number::complete::double,
-    sequence::{preceded, terminated, tuple},
+    sequence::{pair, preceded, terminated, tuple},
     Finish, IResult,
 };
 use nom_locate::LocatedSpan;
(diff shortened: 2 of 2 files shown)


quil-rs/src/parser/common.rs
```diff
@@ -178,7 +178,10 @@ pub(crate) fn parse_matrix<'a>(
             token!(NewLine),
             preceded(
                 token!(Indentation),
-                separated_list0(token!(Comma), parse_expression),
+                separated_list0(
+                    pair(token!(Comma), many0(token!(Indentation))),
```
MarquessV, line 182: This fix might be a little too targeted, but changing how indentation is handled is a much more heavy handed fix.

The Quil spec states that indentation is exactly four spaces and can only occur at the start of a line. The new line is part of the definition of indentation, which means that indenting more than once in a single line isn't possible. It's unclear if this is intentional or not. Nothing in the Quil spec states that nested definitions are invalid, but they would require multiple levels of indentation.

quil-rs/src/parser/lexer/mod.rs
```diff
@@ -262,21 +262,17 @@ fn lex_identifier_raw(input: LexInput) -> InternalLexResult<String> {
...
+                take_while(is_valid_identifier_end_character),
+                recognize(many0(pair(
+                    take_while1(is_dash),
+                    take_while1(is_valid_identifier_end_character),
+                ))),
             )),
-            |(left, right)| format!("{left}{right}"),
+            |(leading, middle, trailing_dash_vars)| {
+                format!("{leading}{middle}{trailing_dash_vars}")
+            },
```
MarquessV, line 273: The idea here is to parse any valid identifier up to any dashes. After that point, any dash should be followed by at least one valid end character for an identifier. Any such pairs are returned and appended to the resulting identifier.

quil-rs/src/parser/common.rs
jselig-rigetti: Is this only when someone wants to align the non-leftmost items within a matrix? Would it necessarily be a `Indentation` token or could it be any amount of whitespace?

quil-rs/src/parser/lexer/mod.rs
jselig-rigetti: > After that point, any dash should be followed by at least one valid end character for an identifier

"any dash" or "any dash sequence"? If not the latter, then https://github.com/rigetti/quil-rs/pull/286/files#diff-a5bcc7a39fd5fe75b07e825fc96eeaa32b6cbed091f5936b09d97f9e9112769eR267 seems off

MarquessV: Yeah, the latter, a singular dash or a sequence of dashes are both valid, as long a valid end character comes after.

quil-rs/src/parser/common.rs
MarquessV: This use case is just for aligning items in the matrix. The lexer conisders a tab character or four consecutive spaces as a single indentation token. Other whitespace is ignored.

jselig-rigetti (approved):

quil-rs/src/parser/common.rs
jselig-rigetti: oops - I'll unresolve since this was your thread originally

PR #289, opened 2023-09-22
fix(py): Implement not equal comparisons

MarquessV:
`__richcmp__` needs to have the behavior for the not equals comparator defined explicitly, it doesn't just implicitly derive it from the equality check. This implements that with a macro, so the behavior is consistent across all the bindings.


notmgsk (approved):
Good catch

quil-py/src/lib.rs
```diff
@@ -44,3 +44,25 @@ macro_rules! impl_to_quil {
...
+                &self,
+                py: pyo3::Python<'_>,
+                other: &Self,
+                op: pyo3::pyclass::CompareOp,
+            ) -> pyo3::PyObject {
+                use pyo3::IntoPy;
+                match op {
+                    pyo3::pyclass::CompareOp::Eq => (self == other).into_py(py),
+                    pyo3::pyclass::CompareOp::Ne => (self != other).into_py(py),
+                    _ => py.NotImplemented(),
```
Shadow53, line 63: Out of curiosity, is there a specific reason you went for `#[derive(PartialEq)]` and `self == other` instead of the `self.as_inner() == other.as_inner()` that the replaced code had?

Shadow53 (approved):

quil-py/src/lib.rs
MarquessV: Mostly because I don't see there being much of a reason to tie the implementation to `PyWrapper` in that way. This makes the macro more generally applicable to anything that can derive `PartialEq`, rather than anything that implements `PyWrapper` on some inner type that derives `PartalEq`.

PR #292, opened 2023-09-29
ci: Don't automatically publish wheels for pre-releases

MarquessV:


.github/test_events/README.md
```diff
@@ -0,0 +1,11 @@
+# GitHub Test Events
```
jselig-rigetti, line 1: Awesome thing to include, thanks!

.github/workflows/publish-quil-py.yml
```diff
@@ -83,10 +93,11 @@ jobs:
 
   windows: 
     runs-on: windows-latest 
+    needs: [ is-python-release, should-publish-wheels ]
     strategy: 
       matrix: 
         python-version: ['3.8', '3.9', '3.10', '3.11']
```
jselig-rigetti, line 99: Should we add `3.12` now that it's stable?

jselig-rigetti (approved):

.github/workflows/publish-quil-py.yml
MarquessV: Here are those issues:
https://github.com/rigetti/rigetti-pyo3/issues/32
https://github.com/rigetti/quil-rs/issues/305
https://github.com/rigetti/qcs-sdk-rust/issues/382

PR #293, opened 2023-10-03
fix: no percent symbol in variable qubit outputs

genos:
In [this pyQuil issue](https://github.com/rigetti/pyquil/issues/1673), we saw that the v3 version of pyQuil can't parse `DEFCALS` with a variable qubit; e.g.

```
DEFCAL I %q:
	DELAY q 4e-8
```

would cause pyQuil to crash, complaining about the `%` symbol. While the [spec claims](https://quil-lang.github.io/#12-6Defining-Calibrations) this is a valid way of writing variable qubits for calibrations, the percent symbol prefix is only otherwise used for parameters (e.g. angles for rotational gates), not for qubits themselves. This PR changes how variable qubits are printed, removing the `%` prefix.

If this change is merged into `quil-rs`, chances are we probably want to update the spec accordingly as well.

### DIFF
--- a/quil-rs/src/instruction/mod.rs
+++ b/quil-rs/src/instruction/mod.rs
@@ -219,23 +219,16 @@ fn write_qubits(
     Ok(())
 }
 
-/// Write qubits as a Quil parameter list, where each variable qubit must be prefixed with a `%`
-/// and all are prefixed with ` `. Return an error if any is a qubit placeholder.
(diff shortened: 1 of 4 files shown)


notmgsk:
https://github.com/quil-lang/quil/issues/66

Shadow53 (approved):
LGTM. Hopefully the spec is able to be updated soon.

PR #295, opened 2023-10-11
fix: implement PartialOrd correctly for types implementing Ord

### LINKED ISSUE #294: Invalid PartialOrd implementation
Newer versions of `clippy` surface a lint error relating to how `PartialOrd` is manually implemented on a couple of types that implement `Ord`. This should be fixed so that actions succeed.

Shadow53:
Resolves #294.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1046,7 +1046,7 @@ checksum = "a1d01941d82fa2ab50be1e79e6714289dd7cde78eba4c074bc5a4374f650dfe0"
 
 [[package]]
 name = "quil-py"
-version = "0.5.1-rc.1"
+version = "0.5.6-rc.0"
 dependencies = [
  "ndarray",
  "numpy",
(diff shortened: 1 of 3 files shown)


genos:
👍 LGTM

jselig-rigetti (approved):
If something implements `Ord`, does it automatically implement `PartialOrd`, or is that implemented here because it's required for `Ord`? Confusing that it it then using the `Ord` implementation, but /shrug

Shadow53:
> If something implements `Ord`, does it automatically implement `PartialOrd`, or is that implemented here because it's required for `Ord`? Confusing that it it then using the `Ord` implementation, but /shrug

It is not an automatic impl, but `PartialOrd` is required to be implemented if `Ord` is, similar to `PartialEq` and `Eq`. Since `Ord` means you can always compare two items, the convention is to return its result wrapped in `Some` for `PartialOrd` to ensure consistency.

PR #300, opened 2023-10-19
fix(quil-rs): match exactly one qubit for DELAYs without frame specifier

### LINKED ISSUE #299: DELAY implementation does not match specification
The [quil specification states](https://quil-lang.github.io/#12-7Timing-and-Synchronization)
> If frame names are specified, then the delay instruction affects those frames on those qubits. If no frame names are specified, all frames on precisely those qubits are affected (1).
 > > Note: Note: this excludes frames which intersect the specified qubits but involve others (2). For example, DELAY 0 1.0 delays one qubit frames on 0, such as 0 "xy", but leaves other frames, such as 0 1 "cz", unaffected.

The current implementation however matches the instruction `DELAY 0 1 1.0` with the frame `0 1 "2q"`. This is at odds with (1) and (2) above. Per the specification, `DELAY 0 1 1.0` should match frames using _only_ `0` or _only_ `1`.

notmgsk:
Closes #299. 

The incorrect behavior is caused by the use of `FrameMatchCondition::ExactQubits` when querying the frames used by a `DELAY` instruction which specifies only qubits (e.g. `DELAY 0 1 0.1`): in this case, each used frame must use all of the instruction's qubits. However, according to the Quil spec, this form of `DELAY` uses frames associated with any of the instruction's qubits. That is, in the incorrect implementation `DELAY 0 1 1.0` would use the frame `0 1 "2q"` but neither `0 "1q"` nor `1 "1q"`. In the correct implementation `DELAY 0 1 1.0` would use the frame `0 "1q"`, `1 "2q"`, and `0 1 "2q"`, but not `0 1 2 "3q"`.

The fix here is to check that the frame's qubits are a subset of the `DELAY`'s qubits.

### DIFF
--- a/quil-rs/src/program/frame.rs
+++ b/quil-rs/src/program/frame.rs
@@ -77,11 +77,8 @@ impl FrameSet {
             FrameMatchCondition::AnyOfQubits(qubits) => keys
                 .filter(|&f| f.qubits.iter().any(|q| qubits.contains(&q)))
                 .collect(),
-            FrameMatchCondition::ExactQubit(qubits) => keys
-                .filter(|&f| f.qubits.len() == 1 && qubits.contains(&f.qubits[0]))
-                .collect(),
             FrameMatchCondition::ExactQubits(qubits) => keys
-                .filter(|&f| f.qubits.iter().collect::<HashSet<_>>() == qubits)
+                .filter(|&f| f.qubits.iter().all(|q| qubits.contains(q)))
                 .collect(),
             FrameMatchCondition::Specific(frame) => {
                 // This unusual pattern (fetch key & value by key, discard value) allows us to return
@@ -179,9 +176,6 @@ pub(crate) enum FrameMatchCondition<'a> {
 
     /// Match all frames which contain any of these qubits
     AnyOfQubits(HashSet<&'a Qubit>),
-
-    /// Match all frames which use exactly one of these qubits
-    ExactQubit(HashSet<&'a Qubit>),
 
     /// Match all frames which contain exactly these qubits
     ExactQubits(HashSet<&'a Qubit>),
--- a/quil-rs/src/program/frame.rs
+++ b/quil-rs/src/program/frame.rs
@@ -77,6 +77,9 @@ impl FrameSet {
             FrameMatchCondition::AnyOfQubits(qubits) => keys
                 .filter(|&f| f.qubits.iter().any(|q| qubits.contains(&q)))
                 .collect(),
+            FrameMatchCondition::ExactQubit(qubits) => keys
+                .filter(|&f| f.qubits.len() == 1 && qubits.contains(&f.qubits[0]))
+                .collect(),
             FrameMatchCondition::ExactQubits(qubits) => keys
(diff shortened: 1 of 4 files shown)


quil-rs/src/program/frame.rs
```diff
@@ -77,6 +77,9 @@ impl FrameSet {
             FrameMatchCondition::AnyOfQubits(qubits) => keys
                 .filter(|&f| f.qubits.iter().any(|q| qubits.contains(&q)))
                 .collect(),
+            FrameMatchCondition::ExactQubit(qubits) => keys
+                .filter(|&f| f.qubits.len() == 1 && qubits.contains(&f.qubits[0]))
```
notmgsk, line 81: I think this could be simpler

Shadow53:
I think the changes you've made correctly do what you are wanting them to, but I'm not sure your interpretation of the spec is correct:

> If frame names are specified, then the delay instruction affects those frames on those qubits. If no frame names are specified, all frames on precisely those qubits are affected (1).
>
> > Note: Note: this excludes frames which intersect the specified qubits but involve others (2). For example, DELAY 0 1.0 delays one qubit frames on 0, such as 0 "xy", but leaves other frames, such as 0 1 "cz", unaffected.

I read this as saying that `DELAY 0 1 1.0` should match _all three_ of `0 "1q"`, `1 "2q"`, and `0 1 "2q"`, but not a `0 1 2 "3q"`, as the first three's qubits are a complete subset of the qubits listed in the `DELAY` instruction, but the last one does not match, as it has qubit `2` which is not in the `DELAY` instruction.

In other words, I think that the necessary change is to turn

```rust
.filter(|&f| f.qubits.iter().collect::<HashSet<_>>() == qubits)
```

into

```rust
.filter(|&f| f.qubits.iter().all(|qubit| qubits.contains(qubit)))
```

notmgsk:
> I think the changes you've made correctly do what you are wanting them to, but I'm not sure your interpretation of the spec is correct:
> 
> > If frame names are specified, then the delay instruction affects those frames on those qubits. If no frame names are specified, all frames on precisely those qubits are affected (1).
> > > Note: Note: this excludes frames which intersect the specified qubits but involve others (2). For example, DELAY 0 1.0 delays one qubit frames on 0, such as 0 "xy", but leaves other frames, such as 0 1 "cz", unaffected.
> 
> I read this as saying that `DELAY 0 1 1.0` should match _all three_ of `0 "1q"`, `1 "2q"`, and `0 1 "2q"`, but not a `0 1 2 "3q"`, as the first three's qubits are a complete subset of the qubits listed in the `DELAY` instruction, but the last one does not match, as it has qubit `2` which is not in the `DELAY` instruction.
> 
> In other words, I think that the necessary change is to turn
> 
> ```rust
> .filter(|&f| f.qubits.iter().collect::<HashSet<_>>() == qubits)
> ```
> 
> into
> 
> ```rust
> .filter(|&f| f.qubits.iter().all(|qubit| qubits.contains(qubit)))
> ```

Good spot. Thanks.

Shadow53 (approved):
Looks good! I would suggest renaming the PR/commit to `fix: ` so that `quil-py` also gets bumped by this change

Issue #301, opened 2023-10-24
Program::from_str fails for PRAGMA DELAY 0 "0.1"


jselig-rigetti:
This test fails:

```rs
    #[rstest]
    #[case(r#"PRAGMA DELAY 0 "0.1""#)]
    fn compile_delays(#[case] quil_input: &str) {
        let program = Program::from_str(quil_input).expect("Should compile.");
        let quil_output = program.to_quil().expect("Should have valid quil output.");

        assert_eq!(quil_input, quil_output.trim());
    }
```

With the following error:

```
ParsingError(
  Syntax(
    ParseError(
      Error {
        line: 1,
        column: 1,
        snippet: "COMMAND(PRAGMA)",
        kind: Other(InvalidCommand { command: Pragma }),
        previous: Some(Error {
          line: 1,
          column: 8,
          snippet: "COMMAND(DELAY)",
          kind: Other(ExpectedToken {
            actual: COMMAND(DELAY),
            expected: "Identifier"
          }),
          previous: None
        })
      }
    )
  )
)
```

jselig-rigetti:
Causes https://github.com/rigetti/forest-benchmarking/pull/237#discussion_r1369481046

kalzoo:
I believe this is intended, PRAGMA DELAY is not consistent with the Quil spec (because DELAY is a reserved word) and support has been dropped from pyQuil v4 et al

jselig-rigetti:
Closed as wontfix

PR #302, opened 2023-10-24
fix: PRAGMA DELAY parse error

jselig-rigetti:
Fixes https://github.com/rigetti/quil-rs/issues/301


jselig-rigetti:
Ah - the issue isn't in the parser, it's in the lexer. It's seeing `PRAGMA DELAY` as two unrelated tokens, so `PRAGMA` becomes the command and then `DELAY` becomes another command, since it is a reserved command word.

PR #304, opened 2023-10-26
fix: add missing instruction memory accesses

### LINKED ISSUE #303: Get instruction memory access bugs
I stumbled across what I believe are bugs in [Instruction::get_memory_accesses](https://github.com/rigetti/quil-rs/blob/3108fb555908e5e29095c86917b076b46ecb3422/quil-rs/src/program/memory.rs#L97). 

1. Set/ shift frequency instructions do not return any memory accesses, despite the fact that `frequency` is an `Expression`.
2. `STORE` does not include its offset in the memory reads (`LOAD` does ✅ ).
3. `EXCHANGE` includes both operands in its write accesses, but neither in the read accesses (both operands are both read and written).
4. `CONVERT` does not include its destination as a write, nor its source as a read.

erichulburd:
Fixes #303

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1046,7 +1046,7 @@ checksum = "a1d01941d82fa2ab50be1e79e6714289dd7cde78eba4c074bc5a4374f650dfe0"
 
 [[package]]
 name = "quil-py"
-version = "0.5.6"
+version = "0.5.8"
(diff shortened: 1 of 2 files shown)


kalzoo (approved):
good fix, thanks!

PR #306, opened 2023-10-31
feat!: Program now has a gate_definitions property that stores all DEFGATEs in a program. These instructions will no longer appear in body_instructions.

### LINKED ISSUE #288: Gate definitions should be treated like other definitions
Gate definitions should not be treated differently from frame or calibration definitions. They should have their own property, be removed from body instructions, and consistently printed at the top of programs when they are converted to Quil.

MarquessV:
closes #288, tested here and through rigetti/pyquil#1688


quil-rs/src/program/mod.rs
```diff
@@ -74,6 +74,7 @@ pub struct Program {
     pub frames: FrameSet,
     pub memory_regions: BTreeMap<String, MemoryRegion>,
     pub waveforms: BTreeMap<String, Waveform>,
+    pub gate_definitions: BTreeMap<String, GateDefinition>,
```
MarquessV, line 77: `BTreeMap` is one of a couple options here:

* `HashMap` - Maybe faster for large enough maps, but isn't hashable, and program stringification would be non-deterministic. This wouldn't be new, since `FrameSet` already uses a `HashMap` under the hood, but I still think it's worth considering from a UX perspective. In any case, we're no longer appending gate definitions to a Vector, so overall performance should be more consistent, and potentially faster for large enough programs.
* `HashSet`/`BTreeSet` - Technically, `GateDefinition`s are hashable, so it isn't required we key them on `name` in a map, but it's overall a better UX IMO to do so. This also makes de-duplication of `DEFGATE`s trivial, which, based on previous discussions, may or may not be desired.
* A custom type a la `FrameSet` or `CalibrationSet`. Felt like unneeded complexity, but could insulate against future breaking changes if we happen to think of something that requires wrapping the map.

`BTreeMap` is my preferred option here, but open to ideas. If nothing else, how do we feel about de-duping `DEFGATE`s? The Quil spec doesn't have any rules regarding duplicate definitions, but in talking with Bram and others, it seems like a meaningful UX improvement to have. This also keeps `Program`s lean in the sense that a Program can only have one definition per identifier, which is realistically all that can be executed, despite not having an explicit rule in the Quil spec.

quil-py/src/program/mod.rs
```diff
@@ -159,17 +159,22 @@ impl PyProgram {
     }
 
     #[getter]
-    pub fn defined_gates(&self, py: Python<'_>) -> PyResult<Vec<PyGateDefinition>> {
```
MarquessV, line 162: Since this is already a breaking change, I'm removing this property. It's a vestige from one of the first pushes to implement `quil-py`, and superseded by the new property.

quil-rs/src/program/mod.rs
```diff
@@ -1048,6 +1084,30 @@ I 0
         assert_abs_diff_eq!(matrix.as_ref().unwrap(), expected);
     }
 
+    /// Tests that the various methods of getting the instructions from a Program produce
+    /// consistent results.
+    #[test]
+    fn test_to_instructions() {
```
MarquessV, line 1090: Added some more tests because it's too easy to miss updating some of these methods when adding something like this.

kalzoo (approved):
looks good!

quil-rs/src/program/mod.rs
kalzoo: BTreeMap seems right to me, and consistent with others. The catch: it is possible for the key to fall out of sync with the name within the `GateDefinition` itself

Shadow53: > it is possible for the key to fall out of sync with the name within the GateDefinition itself

That's _arguably_ a reason for a wrapping type, no? Even just as a newtype?

One other possibility (though perhaps a bit more work) is an ordered map type based on a pair of `Vec`s. For some internal work I've been doing, it seems to be ~20% faster than a `BTreeMap` and maintains sorted ordering.

quil-py/quil/program/__init__.pyi
```diff
@@ -35,6 +35,10 @@ class Program:
     @waveforms.setter
     def waveforms(self, waveforms: Dict[str, Waveform]): ...
     @property
+    def gate_definitions(self) -> Dict[str, GateDefinition]: ...
+    @gate_definitions.setter
+    def gate_definitions(self, gaet_definitions: Dict[str, GateDefinition]): ...
```
Shadow53, line 40:
```suggestion
    def gate_definitions(self, gate_definitions: Dict[str, GateDefinition]): ...
```

Shadow53 (changes requested):

PR #314, opened 2023-11-22
doc(quil-py): Publish documentation on GitHub Pages

MarquessV:
Use [pdoc](https://pdoc.dev/) and GitHub pages to build and deploy a [live site](https://rigetti.github.io/quil-rs/quil.html) with our `quil-py` type hints as documentation. 

Due to some issue between pdoc, pyo3, and type hints, nested submodules don't render. This may be [related to this issue](https://github.com/mitmproxy/pdoc/issues/633).


MarquessV:
GH pages doesn't yet support previewing documentation from a PR, so I had to test the workflow by deploying from this branch. That means the [docs are live](https://rigetti.github.io/quil-rs/quil.html)!

As a todo, I'll need to remove this branch from the allowlist for GH pages deployments, and refactor the workflow so that it only triggers on a full release.

quil-py/2
Shadow53, line 1: Is this intended to exist? Looks like a potential mistyped `> &2`

quil-py/build/docs/quil.html
Shadow53, line 1: I think this directory was supposed to be removed/ignored

quil-py/docs/expression.html
Shadow53, line 1: Should this directory also be ignored?

quil-py/quil/.null-ls_220677___init__.py
Shadow53, line 1: Should be deleted, I think

Shadow53 (commented):
Seems like some stuff was committed by accident, but otherwise looks good.

PR #317, opened 2023-11-23
fix: use internal QuotedString wrapper to quote Quil strings correctly

### LINKED ISSUE #315: String serialization output does not match spec.
The [quil spec](quil-lang.github.io) specifies that a String is made up of a series of characters inside of double quotation marks, with escape characters `\"` and `\\`.

`quil-rs` currently uses a debug-printed string to handle the quotation marks and escaping, but it escapes too much.

This is a valid string in Quil:

```
"   This is
 a multi-line
string
"
```

But `quil-rs` will encode it as `"   This is\n a multi-line\nstring\n"`, which can cause errors when parsing it later in downstream projects.

Shadow53:
Fixes #315.

This uses a new `QuotedString` wrapper with a custom `Display` impl to only escape quotation marks and backslashes in strings, and nothing else.

### DIFF
--- a/quil-rs/src/instruction/mod.rs
+++ b/quil-rs/src/instruction/mod.rs
@@ -319,6 +319,25 @@ impl Quil for Instruction {
     }
 }
 
+pub(crate) struct QuotedString<S>(pub(crate) S);
+
+impl<S> fmt::Display for QuotedString<S>
+where
+    S: AsRef<str>,
+{
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "\"")?;
+        for c in self.0.as_ref().chars() {
+            match c {
+                '"' => write!(f, "\\\"")?,
+                '\\' => write!(f, "\\\\")?,
+                c => write!(f, "{}", c)?,
+            }
+        }
+        write!(f, "\"")
+    }
+}
+
 #[cfg(test)]
 mod test_instruction_display {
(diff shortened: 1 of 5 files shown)


kalzoo:
Thanks for the clear writeup and the fix with a test!

kalzoo (approved):

quil-rs/src/instruction/mod.rs
```diff
@@ -319,6 +319,25 @@ impl Quil for Instruction {
...
+{
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "\"")?;
+        for c in self.0.as_ref().chars() {
+            match c {
+                '"' => write!(f, "\\\"")?,
+                '\\' => write!(f, "\\\\")?,
+                c => write!(f, "{}", c)?,
+            }
+        }
```
BatmanAoD, line 336: I suspect `String::replace` would be faster than this, but we'd need to benchmark it. I don't know how much the performance of this conversion matters, though.

Shadow53: This is specifically inside of `Display`, so I think this might actually be faster. `String::replace` would be creating a new owned `String` and then calling _its_ `Display`, which probably looks similar to this loop.

BatmanAoD: Ah, fair point. 👍🏻

PR #318, opened 2023-11-23
fix!: make NaN == NaN inside Expression

### LINKED ISSUE #316: Expressions containing NaN are not equal when everything else matches
This was surfaced by `proptest` fuzzing just now. The expression `"aa[0]/(pi - pi)"` fails the simplification test because `NaN != NaN`. This follows the IEEE standard, but I think for our purposes it may make sense to make them equal for the purposes of `PartialEq for Expression`.

Shadow53:
This is the easiest way to resolve #316 in a way that doesn't break the API (the other way being to use `OrderedFloat`).

Fixes #316.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1139,7 +1139,7 @@ checksum = "a1d01941d82fa2ab50be1e79e6714289dd7cde78eba4c074bc5a4374f650dfe0"
 
 [[package]]
 name = "quil-cli"
-version = "0.6.3-rc.0"
+version = "0.7.0-rc.0"
 dependencies = [
  "anyhow",
  "clap",
@@ -1148,7 +1148,7 @@ dependencies = [
 
 [[package]]
 name = "quil-py"
-version = "0.13.3-rc.1"
+version = "0.14.0-rc.0"
 dependencies = [
  "indexmap",
  "ndarray",
@@ -1163,7 +1163,7 @@ dependencies = [
 
 [[package]]
 name = "quil-rs"
-version = "0.29.3-rc.0"
+version = "0.30.0-rc.0"
 dependencies = [
  "approx",
  "clap",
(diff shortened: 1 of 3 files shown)


BatmanAoD (commented):
Looks correct to me; I don't think I'm the one to make the call on actually deciding we want `NaN==NaN` for our users, though.

notmgsk:
Hm. Not sure how to feel about this one. Part of me wants something like `Expression::Nan`.

antalsz:
This showed up again in #432

antalsz:
In discussion with @erichulburd today, it seems like the best approach is to take this method, but make it available under a different name rather than `PartialEq`, and have the `round_trip` test use that.  This should resolve the concern with NaN behaving unexpectedly politely while still allowing us to get the polite behavior.

antalsz:
On second thought, I'm warming to the approach here actually being correct.  We claim that `Expression`s are `Eq`; that requires promising that equality is an equivalence relation, which is not currently true.  It will be true if NaNs compare equal; maybe all that we need to do is document that well and then roll with this change.  Thoughts?  cc. @erichulburd, @Shadow53, @kalzoo

antalsz:
@Shadow53 I hope you don't mind, I rebased this branch, updated the proptest seed, and added the documentation I mentioned.  If this looks good to folks, I think this is ready to merge (subject to CI).

erichulburd:
> On second thought, I'm warming to the approach here actually being correct. We claim that `Expression`s are `Eq`; that requires promising that equality is an equivalence relation, which is not currently true. It will be true if NaNs compare equal; maybe all that we need to do is document that well and then roll with this change. Thoughts? cc. @erichulburd, @Shadow53, @kalzoo

I don't feel too strongly either way and am happy to defer to you and @Shadow53 on this.

PR #321, opened 2023-12-13
feat: Add `Program.wrap_in_loop()` method

### LINKED ISSUE #320: Add method to `Program` that will wrap it in a Quil based num-shots loop
The QVM and QCS API accept a `trials`/`num_shots` parameter that allows users to configure how many times their programs run. However, it is possible, and sometimes desirable, to do this by just using Quil's classical instructions. We can add a method to `Program` that applies a generic template to the contained program to wrap it in a "num shots" loop.

Example template:
```
# User declarations go here
DECLARE shot_count INTEGER[1]
MOVE shot_count[0] 100 # Configurable via method paramter
LABEL @START
# Rest of the user program goes here
SUB shot_count[0] 1
JUMP-UNLESS @END shot_count[0]
JUMP @START
LABEL @END
```

We can then use this utility to implement rigetti/pyQuil#1712 upstream.

MarquessV:
closes #320

### DIFF
--- a/quil-py/quil/program/__init__.pyi
+++ b/quil-py/quil/program/__init__.pyi
@@ -15,6 +15,7 @@ from quil.instructions import (
     MemoryReference,
     Qubit,
     Sharing,
+    Target,
     Vector,
     Waveform,
     TargetPlaceholder,
@@ -121,22 +122,24 @@ class Program:
         Convert the instruction to a Quil string. If any part of the instruction can't
         be converted to valid Quil, it will be printed in a human-readable debug format.
         """
-    def wrap_in_loop(self, loop_count_reference: MemoryReference, iterations: int) -> "Program":
+    def wrap_in_loop(
+        self, loop_count_reference: MemoryReference, start_target: Target, end_target: Target, iterations: int
+    ) -> "Program":
         """
         Return a copy of the `Program` wrapped in a loop that repeats ``iterations`` times.
 
         The loop is constructed by wrapping the body of the program in classical Quil instructions.
         The given ``loop_count_reference`` must refer to an INTEGER memory region. The value at the
-        reference given will be set to `iterations` and decremented in the loop. The loop will
+        reference given will be set to ``iterations`` and decremented in the loop. The loop will
         terminate when the reference reaches 0. For this reason your program should not itself
         modify the value at the reference unless you intend to modify the remaining number of
-        iterations (e.g. to break the loop).
+        iterations (i.e. to break the loop).
 
-        The loop is constructed with `quil.instructions.TargetPlaceholder`s in order to prevent
-        conflicts with existing label names in the program. These placeholders are resolved before
-        returning the program.
+        The given ``start_target`` and ``end_target`` will be used as the entry and exit points for
+        the loop, respectively. You should provide unique `quil.instructions.Target`s that won't be
+        used elsewhere in the program.
 
-        If ``iterations`` is 0, then a copy of the program is returned without any changes.
+        If `iterations` is 0, then a copy of the program is returned without any changes.
         """
         ...
     def resolve_placeholders(self) -> None:
--- a/quil-py/quil/program/__init__.pyi
+++ b/quil-py/quil/program/__init__.pyi
@@ -12,6 +12,7 @@ from quil.instructions import (
     Instruction,
     MeasureCalibrationDefinition,
     Measurement,
+    MemoryReference,
     Qubit,
     Sharing,
     Vector,
@@ -120,6 +121,24 @@ class Program:
         Convert the instruction to a Quil string. If any part of the instruction can't
         be converted to valid Quil, it will be printed in a human-readable debug format.
         """
+    def wrap_in_loop(self, loop_count_reference: MemoryReference, iterations: int) -> "Program":
+        """
+        Return a copy of the `Program` wrapped in a loop that repeats ``iterations`` times.
+
+        The loop is constructed by wrapping the body of the program in classical Quil instructions.
+        The given ``loop_count_reference`` must refer to an INTEGER memory region. The value at the
+        reference given will be set to `iterations` and decremented in the loop. The loop will
+        terminate when the reference reaches 0. For this reason your program should not itself
+        modify the value at the reference unless you intend to modify the remaining number of
+        iterations (e.g. to break the loop).
+
+        The loop is constructed with `quil.instructions.TargetPlaceholder`s in order to prevent
+        conflicts with existing label names in the program. These placeholders are resolved before
+        returning the program.
+
+        If ``iterations`` is 0, then a copy of the program is returned without any changes.
+        """
+        ...
     def resolve_placeholders(self) -> None:
         """
         Resolve ``TargetPlaceholder``s and ``QubitPlaceholder``s within the program using default resolvers.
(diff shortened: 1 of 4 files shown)


quil-py/quil/program/__init__.pyi
```diff
@@ -120,6 +121,24 @@ class Program:
         Convert the instruction to a Quil string. If any part of the instruction can't
         be converted to valid Quil, it will be printed in a human-readable debug format.
         """
+    def wrap_in_loop(self, loop_count_reference: MemoryReference, iterations: int) -> "Program":
+        """
+        Return a copy of the `Program` wrapped in a loop that repeats ``iterations`` times.
+
+        The loop is constructed by wrapping the body of the program in classical Quil instructions.
+        The given ``loop_count_reference`` must refer to an INTEGER memory region. The value at the
+        reference given will be set to `iterations` and decremented in the loop. The loop will
```
jselig-rigetti, line 130: Should this be double-backtick'd?

```suggestion
        reference given will be set to ``iterations`` and decremented in the loop. The loop will
```

quil-rs/src/program/mod.rs
```diff
@@ -379,6 +381,74 @@ impl Program {
...
+    ///
+    /// If `iterations` is 0, then a copy of the program is returned without any changes.
+    pub fn wrap_in_loop(&self, loop_count_reference: MemoryReference, iterations: u32) -> Self {
+        if iterations == 0 {
+            return self.clone();
+        }
+
+        let mut looped_program = self.clone_without_body_instructions();
+        let start_target = TargetPlaceholder::new("__loop-start".to_string());
+        let end_target = TargetPlaceholder::new("__loop-end".to_string());
```
jselig-rigetti, line 405: Do these need to be unique at all? What happens if you call this multiple times e.g. `wrap_in_loop().wrap_in_loop()` or otherwise have nesting?

quil-rs/src/program/mod.rs
```diff
@@ -379,6 +381,74 @@ impl Program {
...
+                    condition: loop_count_reference,
+                }),
+                Instruction::Jump(Jump {
+                    target: Target::Placeholder(start_target),
+                }),
+            ])
+            .collect::<Vec<Instruction>>(),
+        );
+
+        looped_program.resolve_placeholders();
```
kalzoo, line 447: wise to think of this!

the catch: it will also resolve any existing placeholders existing in the program, which the user may not want. Would it make sense to have the user provide the Label alongside the MemoryReference? That seems like it may be more consistent.

kalzoo (changes requested):
This looks great, aside from the thought about placeholders in the comment.

quil-rs/src/program/mod.rs
MarquessV: Yeah, `pdoc` uses a markdown extension that makes it such that paths wrapped in single backticks links to that path in the documentation. Double backticks is normal inline code.

Not sure why the action hasn't updated the deployment notification on this PR, but [you can see how it renders in the preview here](https://rigetti.github.io/quil-rs/pr-preview/pr-321/quil/program.html#Program.wrap_in_loop).

quil-py/quil/program/__init__.pyi
MarquessV: Yeah, `pdoc` uses a markdown extension that makes it such that paths wrapped in single backticks links to that path in the documentation. Double backticks is normal inline code.

Not sure why the action hasn't updated the deployment notification on this PR, but [you can see how it renders in the preview here](https://rigetti.github.io/quil-rs/pr-preview/pr-321/quil/program.html#Program.wrap_in_loop).

quil-rs/src/program/mod.rs
MarquessV: By default, the placeholders resolve by checking for conflicting targets and appending a unique integer to the end, so I believe that should have worked as expected.

Though, that is a moot point now that this method makes the user provide `Target`s per Kalan's comment.

quil-rs/src/program/mod.rs
MarquessV: Ah yeah, good point. I guess we could use the custom resolver to only resolve the placeholders we created, but I think you make a good point on consistency between how the targets and loop counter are specified. This approach biases towards giving the user control.

quil-py/quil/program/__init__.pyi
```diff
@@ -120,6 +122,26 @@ class Program:
...
+        terminate when the reference reaches 0. For this reason your program should not itself
+        modify the value at the reference unless you intend to modify the remaining number of
+        iterations (i.e. to break the loop).
+
+        The given ``start_target`` and ``end_target`` will be used as the entry and exit points for
+        the loop, respectively. You should provide unique `quil.instructions.Target`s that won't be
+        used elsewhere in the program.
+
+        If `iterations` is 0, then a copy of the program is returned without any changes.
+        """
```
notmgsk, line 143: Might also be worth noting that `iterations` must be non-negative

quil-rs/src/program/mod.rs
```diff
@@ -379,6 +381,76 @@ impl Program {
...
+                    target: start_target.clone(),
+                }),
+            ]
+            .into_iter()
+            .chain(self.body_instructions().cloned())
+            .chain(vec![
+                Instruction::Arithmetic(Arithmetic {
+                    operator: ArithmeticOperator::Subtract,
+                    destination: ArithmeticOperand::MemoryReference(MemoryReference {
+                        name: "__loop-count".to_string(),
```
notmgsk, line 435: This could be a problem if the user wants to use multiple loops (e.g. inner and outer)

quil-rs/src/program/snapshots/quil_rs__program__tests__wrap_in_loop.snap
```diff
@@ -0,0 +1,23 @@
...
+DEFCAL I 0:
+	DELAY 0 1
+MOVE shot_count[0] 10
+LABEL @loop-start
+MEASURE q ro[0]
+JUMP-UNLESS @end-reset ro[0]
+X q
+LABEL @end-reset
+I 0
+SUB __loop-count[0] 1
```
notmgsk, line 20: This doesn't look right. `__loop-count` is neither declared nor used elsewhere.

quil-rs/src/program/mod.rs
notmgsk: Shouldn't this be `loop_count_reference.name`?

quil-py/quil/program/__init__.pyi
MarquessV: Good call, added a note.

quil-rs/src/program/mod.rs
MarquessV: Yup, that explains the snapshot issue too, good catch.

quil-rs/src/program/snapshots/quil_rs__program__tests__wrap_in_loop.snap
```diff
@@ -0,0 +1,23 @@
+---
+source: quil-rs/src/program/mod.rs
+expression: program.to_quil().unwrap()
+---
+DECLARE ro BIT[1]
```
notmgsk, line 5: An idea for the future: we could use `libquil-sys` to validate the behaviour of stuff like this, rather than eye-balling the logic.

notmgsk (approved):
LGTM 🚢

PR #323, opened 2023-12-14
feat: Add methods for identifying Quil-T instructions and filtering instructions from Programs

### LINKED ISSUE #322: Add method for identifying Quil-T instructions and a utility for filtering programs.
It is a common task to want to create a copy of a program without some set of instructions. One common example is creating a copy of a program that doesn't have any Quil-T instructions. Quil-T instructions are well defined in the spec, so we should also add a utility that makes identifying them easier.

MarquessV:
Closes #322 

I think the `Instruction.is_quil_t` method is the most key addition in these changes. The filtering methods border on being too simple of a wrapper to keep. I think they are convenient shortcuts for a common task, but I'd be open to removing them if we think they aren't worth it.


notmgsk (approved):
LGTM!

quil-rs/src/program/mod.rs
```diff
@@ -1317,6 +1328,25 @@ CNOT 2 3
...
+H 1
+CNOT 2 3";
+
+        let program = Program::from_str(input).unwrap();
+        let program_without_quil_t =
+            program.filter_instructions(|instruction| !instruction.is_quil_t());
+        assert_snapshot!(program_without_quil_t.to_quil().unwrap())
+    }
+
     fn test_wrap_in_loop() {
```
notmgsk, line 1350:
```suggestion
    #[test]
    fn test_wrap_in_loop() {
```

PR #324, opened 2023-12-20
Reduce number of classical instruction edges in InstructionBlock::gra…

erichulburd:
The ultimate goal of this PR is to simplify the `InstructionBlock::graph` and thereby improve the ability of a hardware scheduler to schedule a Quil program by reducing the number of edges between classical instructions and block start/end.

Two types of edges in the `InstructionBlock::graph` are deemed extraneous in this work:

1. An edge from two consecutive classical instructions that read from / write to different memory regions. Rather than insist on classical instruction execute in the particular order they are written in a Quil program, _only_ represent ordering of classical Quil instructions based upon the memory regions they read and write to (this is already done with `MemoryAccessQueue`).
2. Edges from BlockStart, if the node has other incoming edges, or to BlockEnd, if the node has other outgoing edges. _Note that while this could apply equally to `InstructionRole::RfControl`, this change set deliberately only impacts edges to and from classical instructions._

```quil
DECLARE params1 REAL[10]
DECLARE params2 REAL[10]
DECLARE params3 REAL[10]
DECLARE params4 REAL[10]
DECLARE integer1 INTEGER[10]

LOAD params1[0] params2 integers[0] # LOAD_1 in graphs below
SHIFT-PHASE 0 "rf" params1[0]
LOAD params3[0] params4 integers[0] # LOAD_2 in graphs below
```

Old representation of `InstructionBlock::graph`.

```mermaid
flowchart TD
    BlockStart --> LOAD_1
    BlockStart --> SHIFT-PHASE
    LOAD_1 --> SHIFT-PHASE
    LOAD_1 --> LOAD_2
    LOAD_1 --> BlockEnd
    LOAD_2 --> BlockEnd
    SHIFT-PHASE --> BlockEnd
```
_In the original algorithm, there is an edge from LOAD_1 to BlockEnd because there are pending reads on `params1` and `integers`._

New representation of `InstructionBlock::graph`.

```mermaid
flowchart TD
    BlockStart --> LOAD_1
    BlockStart --> LOAD_2
    LOAD_1 --> SHIFT-PHASE
    LOAD_2 --> BlockEnd
    SHIFT-PHASE --> BlockEnd
```

There are three invariants of the `InstructionBlock::graph` that remain unchanged:

1. There is always a path connecting the BlockStart to any classical node in the graph.
1. There is always a path connecting any classical node in the graph to the BlockEnd.
1. These invariants pertain to the BlockStart and BlockEnd as well. In other words, there is always a connecting path from the BlockStart to the BlockEnd regardless as to whether there are any instructions in the block.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,5 @@ __pycache__
 # quil-py documentation
 quil-py/build
 
+# unversioned developer notes
+scratch/
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1047,7 +1047,7 @@ checksum = "a1d01941d82fa2ab50be1e79e6714289dd7cde78eba4c074bc5a4374f650dfe0"
 
 [[package]]
 name = "quil-py"
-version = "0.6.1-rc.0"
+version = "0.6.3"
 dependencies = [
  "ndarray",
  "numpy",
@@ -1060,7 +1060,7 @@ dependencies = [
 
 [[package]]
 name = "quil-rs"
-version = "0.22.1-rc.0"
+version = "0.22.3"
 dependencies = [
  "approx",
  "clap",
--- a/quil-rs/Cargo.toml
+++ b/quil-rs/Cargo.toml
@@ -5,7 +5,7 @@ version = "0.22.3"
 edition = "2021"
 rust-version = "1.70"
 license = "Apache-2.0"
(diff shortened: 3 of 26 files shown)


Shadow53:
Closing in favor of #350 with its fixes

PR #336, opened 2024-02-02
feat!: #334: program stats

### LINKED ISSUE #334: Add analysis package
We would like a package for analyzing Quil programs to determine various characteristics without running or simulating them.

BatmanAoD:
Closes #334


kalzoo:
- many of these methods only make sense within the context of a block. They only appear to make sense on a `Program` when that program consists of one block. 

These should be implemented on a  block, and then perhaps implemented as well on a program with an error about multiple blocks when present. Like ExecutionGraph, this can also be computationally expensive and could be done lazily and cached, or eagerly on construction.
- I see a todo about testing the path_fold. I’ll have to review that on my next pass.

quil-rs/src/stats/execution_graph.rs
```diff
@@ -0,0 +1,200 @@
...
+                if let Some(last_instruction) = last_instruction_for_qubit.insert(qubit, node) {
+                    graph.add_edge(last_instruction, node, ());
+                }
+            }
+        }
+
+        Ok(Self { graph })
+    }
+
+    // TODO: TEST THIS WELL
```
Shadow53, line 59: This is implicitly tested via tests on public methods that use it. Can probably remove the TODO.

quil-rs/src/stats/execution_graph.rs
```diff
@@ -0,0 +1,200 @@
...
+    /// 2. `X 0` is visited with accumulator `A`, and a new accumulator `B` is returned from `f`.
+    /// 3. `H 1` is visited with accumulator `A`, and a new accumulator `C` is returned from `f`.
+    /// 4. `CNOT 1 0` is visited with accumulator `B`, and a result value `D` is returned from `f`.
+    /// 5. `CNOT 1 0` is visited with accumulator `C`, and a result value `E` is returned from `f`.
+    /// 5. The result values are collected into a [`Vec`] and returned as `[D, E]`.
+    ///
+    /// # Errors
+    ///
+    /// Any error returned from a call to `f` will be returned immediately.
+    // TBD: visibility
```
Shadow53, line 103: I think it is safe and preferable to leave this private.

quil-rs/src/stats/execution_graph.rs
```diff
@@ -0,0 +1,200 @@
...
+                        .collect(),
+                ));
+            }
+        }
+
+        Ok(result)
+    }
+
+    /// Returns the longest path from an initial instruction (one with no prerequisite instructions) to a final instruction (one with no dependent instructions).
+    pub fn gate_depth(&self) -> Result<usize, Error> {
```
Shadow53, line 135: We had this return `Result` when we wanted to error on non-gate instructions. We now handle this better, so we can maybe make this infallible?

`path_fold` only returns a `Result` because the closures might return one. It looks like we are safe to make that infallible as well. If there's a good reason not to, we can also assert in this function that the result is `Result<_, Infallible>` so we get a compile error if we introduce a real error at some point.

quil-rs/src/stats/mod.rs
```diff
@@ -0,0 +1,270 @@
+mod execution_graph;
+
+#[cfg(test)]
+pub(crate) mod test_programs;
+
+// Use quil_rs::program::graph for pulse-level programs, but not qubit programs.
```
Shadow53, line 6:
```suggestion
// NOTE: Use quil_rs::program::graph for pulse-level programs, but not qubit programs.
```

Shadow53: (Or just delete, but this seems useful to record somewhere)

quil-rs/src/stats/mod.rs
```diff
@@ -0,0 +1,270 @@
...
+
+use crate::{
+    instruction::{Instruction, Qubit},
+    Program,
+};
+use execution_graph::{Error as ExecutionGraphError, ExecutionGraph};
+
+pub trait InstructionsSource {
+    fn body_instructions(&self) -> impl Iterator<Item = &Instruction> + '_;
+    fn num_instructions(&self) -> usize;
```
Shadow53, line 19: Same note as on `Program`

quil-rs/src/stats/mod.rs
```diff
@@ -0,0 +1,270 @@
...
+    fn get_used_qubits(&self) -> impl IntoIterator<Item = &Qubit>;
+}
```
Shadow53, line 21: `impl Trait` as the return type in traits is a very recent thing, IIRC. we might consider using associated types (i.e. `type IntoInstructionIter = &[Instruction]`) for better backwards compatibility.

quil-rs/src/stats/mod.rs
```diff
@@ -0,0 +1,270 @@
...
+    // See note on `Program` impl above.
+    #[allow(refining_impl_trait)]
+    fn get_used_qubits(&self) -> &HashSet<Qubit> {
+        (*self).get_used_qubits()
+    }
+}
+
+pub struct ProgramStats<S: InstructionsSource> {
+    source: S,
+    execution_graph: ExecutionGraph,
```
Shadow53, line 59: Still think it might be a good idea to have the graph contain references to the `source` instead of cloning the instructions.

quil-rs/src/stats/mod.rs
```diff
@@ -0,0 +1,270 @@
...
+            source: program,
+            execution_graph,
+        })
+    }
+
+    // Provided because Program specifically returns a hash-set, but in general
+    // `InstructionsSource` may not provide that feature.
+    pub fn set_of_qubits_used(&self) -> &HashSet<Qubit> {
+        self.source.get_used_qubits()
+    }
```
Shadow53, line 82: If we used associated types for the trait, this would be unnecessary, since the actual type would be exposed as a `HashSet`.

quil-rs/src/stats/mod.rs
```diff
@@ -0,0 +1,270 @@
...
+    /// - gate definitions
+    pub fn body_instruction_count(&self) -> usize {
+        self.source.body_instructions().count()
+    }
+
+    /// The total number of instructions in the program.
+    ///
+    /// This includes all definitions excluded by [`Program::instruction_count`].
+    pub fn instruction_count(&self) -> usize {
+        self.source.num_instructions()
```
Shadow53, line 105: Ignoring the comments about feelings about convention, we should unify the names, IMO.

quil-rs/src/stats/mod.rs
```diff
@@ -0,0 +1,270 @@
...
+        assert_eq!(expected, count);
+    }
+
+    #[rstest]
+    #[case(QUIL_AS_TREE, false)]
+    #[case(QUIL_AS_INVERSE_TREE, false)]
+    #[case(QUIL_AS_LINEAR, false)]
+    #[case(QUIL_WITH_DIAMOND, false)]
+    #[case(KITCHEN_SINK_QUIL, false)]
+    fn has_dynamic_control_flow(#[case] input: &str, #[case] expected: bool) {
```
Shadow53, line 264: Should include cases using each kind of `JUMP` instruction

quil-rs/src/stats/mod.rs
```diff
@@ -0,0 +1,270 @@
...
+        assert_eq!(expected, fidelity);
+    }
+
+    #[rstest]
+    #[case(QUIL_AS_TREE, 0)]
+    #[case(QUIL_AS_INVERSE_TREE, 0)]
+    #[case(QUIL_AS_LINEAR, 0)]
+    #[case(QUIL_WITH_DIAMOND, 0)]
+    #[case(KITCHEN_SINK_QUIL, 0)]
+    fn topological_swap_count(#[case] input: &str, #[case] expected: usize) {
```
Shadow53, line 251: Should include cases actually using `SWAP` gates.

quil-rs/src/program/mod.rs
```diff
@@ -480,13 +480,17 @@ impl Program {
         Box::new(move |key| qubit_resolutions.get(key).copied())
     }
 
-    /// Return a copy of all of the instructions which constitute this [`Program`].
-    pub fn to_instructions(&self) -> Vec<Instruction> {
-        let capacity = self.memory_regions.len()
+    pub fn num_instructions(&self) -> usize {
```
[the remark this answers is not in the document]
BatmanAoD: I think that's fair. `len()` seems correct to me.

quil-rs/src/stats/execution_graph.rs
BatmanAoD: Here's a version where `path_fold` still takes a fallible callback, but `gate_depth` and `multi_qubit_gate_depth` use `Infallible`: https://github.com/rigetti/quil-rs/pull/336/commits/01d59cb36ecacd1516a26fbe12dd016a1d67a4b6#diff-73e235b2289330aa4f68043fbc217ba835e8adb3135e19c4a7b7e5f5a50b4479R103

I'll go ahead and just change the `path_fold` signature to require the callback not to return an error, though.

If we ever do have an execution graph visitor that can fail, we can just use `Result` as the `T` type, and then transpose from `Vec<Result<...>>` to `Result<Vec<...>, >` in the caller.

BatmanAoD: ...hm, since the callback takes `T` as an argument, that's actually not super convenient. In any case, I agree that there's probably no reason to have these things be fallible in the first place.

quil-rs/src/stats/mod.rs
BatmanAoD: I don't actually recall what this means. It doesn't look like we use `program::graph` anywhere in the `stats` module.

quil-rs/src/stats/execution_graph.rs
```diff
@@ -132,34 +133,46 @@ impl ExecutionGraph {
...
+                    } else {
+                        Ok(depth)
+                    }
+                },
+            )
+            .unwrap_or_else(|_| {
+                unreachable!(
+                    "'gate_depth' callback is infallible, so path_fold should not return an error"
+                )
+            });
```
Shadow53, line 152: Is this functionally different from `expect`? (Also below)

quil-rs/src/stats/mod.rs
```diff
@@ -10,53 +10,107 @@ use std::ops::Neg;
...
+
+impl<'a, 'b> From<&'a InstructionBlock<'b>> for InstructionBlockWithQubitSet<'a, 'b> {
+    fn from(block: &'a InstructionBlock<'b>) -> Self {
+        let qubits = block
+            .instructions
+            .iter()
+            .filter_map(|i| match i {
+                Instruction::Gate(_) | Instruction::Measurement(_) => Some(i.get_qubits()),
+                _ => None,
+            })
```
Shadow53, line 82: Why not use `get_qubits` on `Instruction` directly?

quil-rs/src/stats/mod.rs
```diff
@@ -66,25 +120,43 @@ fn make_execution_graph<S: InstructionsSource>(
...
+    pub fn from_block(block: &'a InstructionBlock<'b>) -> Self {
+        let source = block.into();
+        let execution_graph = make_execution_graph(&source);
+
+        Self {
+            source,
+            execution_graph,
+        }
     }
 }
```
Shadow53, line 143: I'm wondering if straight `From` impls would be better

quil-rs/src/stats/mod.rs
```diff
@@ -179,21 +257,59 @@ impl<S: InstructionsSource> ProgramStats<S> {
...
 
     use super::test_programs::*;
     use super::*;
 
+    #[rstest]
+    // #[case(QUIL_AS_TREE, Some(2))]
+    // #[case(QUIL_AS_INVERSE_TREE, Some(2))]
+    // #[case(QUIL_AS_LINEAR, Some(4))]
+    // #[case(QUIL_WITH_DIAMOND, Some(6))]
+    // #[case(QUIL_WITH_SWAP, Some(3))]
```
Shadow53, line 274: Uncomment or clean up

quil-rs/src/stats/execution_graph.rs
BatmanAoD: Hm, for some reason I thought it changed optimization, but no, only `unreachable_unchecked` does that, which makes sense.

quil-rs/src/stats/mod.rs
BatmanAoD: Do you mean without the `filter_map`, or as a fully-qualified `Instruction::get_qubits`?

kalzoo:
Thanks @BatmanAoD for teeing this one up after merging #339 ! Things looked good. I added some more, to include docs and copyrights, and cleaned up the failing tests.

I'd still like to, before merge:

* [ ] validate this against downstream packages (almost done with that, no issues so far)
* [ ] un-refactor BasicBlockTerminator back to its original form, which made more sense than wrapping quil instructions (a suspicion confirmed by the downstream packages)
* [ ] Add `ScheduledProgram` -> `FixedSchedule`, because while that is scope creep for this PR I'd like to make sure that it doesn't cause further breaking changes after we merge these breaking changes. While extensive, that code is already written and tested and just needs to be moved here.
  * [ ] with that added, `qpu_runtime_estimate` becomes trivial to implement.

I asked some folks on a private channel about their opinions re: fidelity; we can see what they'll say. I'd like their input before merging the current implementation; if that blocks us we can remove it and save it for later.

One note: `final_rewriting` is only relevant within the scope of `quilc` and doesn't make sense to implement on a program itself. I removed the comment about it.

kalzoo:
> Thanks @BatmanAoD for teeing this one up after merging #339 ! Things looked good. I added some more, to include docs and copyrights, and cleaned up the failing tests.
> 
> I'd still like to, before merge:
> 
> * [ ]  validate this against downstream packages (almost done with that, no issues so far)
> * [x]  un-refactor BasicBlockTerminator back to its original form, which made more sense than wrapping quil instructions (a suspicion confirmed by the downstream packages)
> * [x]  Add `ScheduledProgram` -> `FixedSchedule`, because while that is scope creep for this PR I'd like to make sure that it doesn't cause further breaking changes after we merge these breaking changes. While extensive, that code is already written and tested and just needs to be moved here.
>   
>   * [x]  with that added, `qpu_runtime_estimate` becomes trivial to implement.
> 
> I asked some folks on a private channel about their opinions re: fidelity; we can see what they'll say. I'd like their input before merging the current implementation; if that blocks us we can remove it and save it for later.
> 
> One note: `final_rewriting` is only relevant within the scope of `quilc` and doesn't make sense to implement on a program itself. I removed the comment about it.

Most of this ^ is now done. Needs only:

* [ ] A fresh look at test coverage
* [ ] completion of downstream integration to make sure nothing's missing here. That's about 30 mins work away from being done, I'd estimate

kalzoo:
Downstream validation is complete and looking good, after the addition here of `BasicBlock.instruction_index_offset`. Ready for review; may add more tests on another look tomorrow.

quil-rs/src/program/analysis/block_metrics.rs
```diff
@@ -0,0 +1,208 @@
...
+        self.instructions()
+            .iter()
+            .copied()
+            .filter_map(get_fidelity)
+            .map(|f: f64| f.ln().powi(2))
+            .sum::<f64>()
+            .sqrt()
+            .neg()
+            .exp()
+    }
```
kalzoo, line 60: Remove

quil-rs/src/program/scheduling/graph.rs
```diff
@@ -289,11 +276,11 @@ impl PreviousNodes {
     }
 }
 
-impl<'a> InstructionBlock<'a> {
+impl<'a> ScheduledBasicBlock<'a> {
+    /// Build a scheduled basic block from a basic block and a program.
+    /// TODO create a ScheduledBasicBlockBuilder so that the inputs may be optional
```
kalzoo, line 281: TODO to consider or remove

quil-rs/src/program/scheduling/graph.rs
```diff
@@ -440,107 +427,49 @@ impl<'a> InstructionBlock<'a> {
...
-            instruction_index,
-            instruction: Instruction::Label(Label {
-                target: Target::Fixed(label),
-            }),
-            variant: ScheduleErrorVariant::DuplicateLabel,
-        });
-    }
-
-    Ok(())
+    // todo needs getter
```
kalzoo, line 471: remove todo

quil-rs/src/program/scheduling/schedule.rs
```diff
@@ -0,0 +1,367 @@
...
+}
+
+impl Zero for Seconds {
+    fn zero() -> Self {
+        Self(0.0)
+    }
+}
+
+// todo: rather than Default, it should require Time::zero. For primitives those are the same,
+// but not so for Expression
```
kalzoo, line 44: Remove completed todo

quil-rs/src/program/scheduling/schedule.rs
```diff
@@ -0,0 +1,367 @@
...
+    }
+
+    pub fn duration(&self) -> &Time {
+        &self.duration
+    }
+}
+
+impl<'p> ScheduledBasicBlock<'p> {
+    // todo create more restricted subtype of instruction for ScheduledBasicBlock so we don't have
+    // to handle unreachable instructions here
```
kalzoo, line 112: Reviewer thoughts? I tried out `subenum` but found it lacking:

* I'd like for it to allow the subenum to be a ref rather than taking ownership
* It doesn't allow for easy categorization, i.e. `InSubGroup` vs `NotInSubGroup`. It only has `InSubGroup::try_from(Parent)` which means that all variants of `Parent` must still be handled when `try_from` fails.

quil-rs/src/program/scheduling/schedule.rs
```diff
@@ -0,0 +1,367 @@
...
+                duration.to_real().ok().map(Seconds)
+            }
+            Instruction::Fence(_)
+            | Instruction::SetFrequency(_)
+            | Instruction::SetPhase(_)
+            | Instruction::SetScale(_)
+            | Instruction::ShiftFrequency(_)
+            | Instruction::ShiftPhase(_)
+            | Instruction::SwapPhases(_) => Some(Seconds(0.0)),
+            // Todo should this tolerate classical instructions? how about reset?
```
kalzoo, line 140: This may depend on the way we use `subenum` if at all, but todo, either (a) confirm that classical instructions are skipped by the graph visitor and never called here or (b) report a duration of 0 for them.

In any case, there should be test cases covering this.

quil-rs/src/program/scheduling/schedule.rs
```diff
@@ -0,0 +1,367 @@
...
+                            .into_iter()
+                            .filter_map(|frame| {
+                                program
+                                    .frames
+                                    .get(frame)
+                                    .and_then(|frame_definition| {
+                                        frame_definition.get("SAMPLE-RATE")
+                                    })
+                                    .map(|sample_rate_expression| match sample_rate_expression {
+                                        AttributeValue::String(_) => todo!("handle error"),
```
kalzoo, line 174: todo

quil-rs/src/program/scheduling/schedule.rs
BatmanAoD: It seems like, rather than a set of "sub-enums", what you're looking for might be a separate enum with fewer variants, and a mapping from `Instruction` to `SchedulableInstruction`, where each variant of `SchedulableInstruction` holds references to the relevant innards of `Instruction`?

kalzoo: not quite - while that seems helpful in some ways, it makes the handling of instructions in, say, `impl<'p> From<&'p Program> for ControlFlowGraph<'p>` just as unwieldy.

quil-rs/src/program/scheduling/graph.rs
kalzoo: removed

kalzoo:
I'll give this a self-review next week, but it's feature-complete and ready for review by others. Please pay particular attention to the python bindings.

kalzoo:
added documentation to the python bindings with a usage example

quil-py/quil/program/__init__.pyi
kalzoo, line 1: reviewer: please focus on this file and whether it's the API we want to expose

quil-rs/src/program/scheduling/graph.rs
```diff
@@ -440,107 +429,48 @@ impl<'a> InstructionBlock<'a> {
...
-    Continue,
-    Halt,
+    pub fn basic_block(&self) -> &BasicBlock<'a> {
+        &self.basic_block
+    }
 }
 
 #[derive(Clone, Debug)]
 pub struct ScheduledProgram<'a> {
-    /// All blocks within the ScheduledProgram, keyed on string label.
```
kalzoo, line 490: This logic was split up into two steps:

* build CFG
* schedule blocks

in the course of the change, it got simpler and more maintainable

quil-rs/src/program/snapshots/quil_rs__program__calibration__tests__expansion@FenceVariableQubit.snap
kalzoo, line 1: This is the only snapshot with meaningful change (because it's new) - this was mishandled before, so this is a regression case

kalzoo (commented):
self-review complete. ready for team review.

quil-py/quil/program/__init__.pyi
```diff
@@ -175,6 +175,82 @@ class Program:
...
+
+        This schedule is computed by:
+
+        * Expanding each instruction within the block using the program's calibration definitions
+        * Resolving the `FixedSchedule` of the expanded instructions
+        * Mapping calibrated instructions back to the original instructions within this block, such that the
+          block's instruction is represented as a timespan encompassing all of its expanded instructions
+
+        If `include_zero_duration_instructions` is `True`, then the schedule will include instructions with zero duration
+        (such as `FENCE`). This may cause certain instructions to appear as unexpectedly long in duration.
```
BatmanAoD, line 199: Why would zero-duration instructions cause "unexpectedly long" durations?

quil-rs/src/program/scheduling/schedule.rs
```diff
@@ -0,0 +1,479 @@
...
+}
+
+impl<Time: Clone + PartialOrd + std::ops::Add<Time, Output = Time> + Zero>
+    From<Vec<ComputedScheduleItem<Time>>> for Schedule<Time>
+{
+    fn from(items: Vec<ComputedScheduleItem<Time>>) -> Self {
+        let duration = items
+            .iter()
+            .map(|item| item.time_span.start_time.clone() + item.time_span.duration.clone())
+            .fold(Time::zero(), |acc, el| if el > acc { el } else { acc });
```
BatmanAoD, line 83: Same question as on the Python doc-comment for `duration`: aren't the "items" here instructions within a block, and doesn't this need to account for possible dependencies between them?

BatmanAoD (commented):
Some questions and a few small suggestions, but I don't think anything is blocking (unless I've identified an actual bug in `duration()`, but I think I just don't understand what's happening there).

quil-py/quil/program/__init__.pyi
```diff
@@ -238,6 +314,73 @@ class CalibrationSet:
...
+class FixedSchedule:
+    def items(self) -> List[FixedScheduleItem]:
+        """
+        All the items in the schedule, in unspecified order.
+        """
+    def duration(self) -> float:
+        """
+        The duration of the schedule, in seconds.
+
+        This is the maximum of the end time of all the items.
```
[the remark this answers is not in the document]
kalzoo: I don't follow; `end_time = start_time + duration`, and the dependencies are accounted for in the staggered start times

quil-py/quil/program/__init__.pyi
```diff
@@ -238,6 +314,73 @@ class CalibrationSet:
...
         ...
 
+
+@final
+class FixedScheduleItem:
+    """
+    A single item within a fixed schedule, representing a single instruction within a basic block.
+    """
+    @property
+    def instruction_index(self) -> int:
```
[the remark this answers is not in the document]
kalzoo: that feels like more magic than is necessary

quil-rs/src/program/scheduling/schedule.rs
```diff
@@ -0,0 +1,479 @@
...
+}
+
+impl Zero for Seconds {
+    fn zero() -> Self {
+        Self(0.0)
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct Schedule<Time> {
```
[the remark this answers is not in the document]
kalzoo: yep, fair

quil-rs/src/program/scheduling/schedule.rs
```diff
@@ -0,0 +1,479 @@
...
+}
+
+impl<
+        Time: Clone
+            + PartialOrd
+            + std::ops::Add<Time, Output = Time>
+            + std::ops::Sub<Time, Output = Time>,
+    > TimeSpan<Time>
+{
+    pub(crate) fn union(self, rhs: Self) -> Self {
```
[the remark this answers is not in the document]
kalzoo: I had considered it; fortunately, it is only `pub(crate)` and so a little more flexible to get wrong

quil-rs/src/program/scheduling/schedule.rs
```diff
@@ -0,0 +1,479 @@
...
+    /// using a default built-in calculation for the duration of scheduled instructions.
+    pub fn as_fixed_schedule(&self, program: &Program) -> ComputedScheduleResult<FixedSchedule> {
+        self.as_schedule(program, Self::get_fixed_instruction_duration)
+    }
+
+    /// Compute the flattened schedule for this [`ScheduledBasicBlock`] using a user-provided
+    /// closure for computation of instruction duration.
+    ///
+    /// Return an error if the schedule cannot be computed from the information provided.
+    pub fn as_schedule<F, Time: Clone + PartialOrd + std::ops::Add<Time, Output = Time> + Zero>(
```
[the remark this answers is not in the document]
kalzoo: Yes, `Expression` is the next to be supported, and it is not `Copy`

quil-py/src/program/scheduling.rs
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    #[getter]
+    pub fn instruction_index(&self) -> usize {
+        self.0.instruction_index
+    }
+}
+
+py_wrap_type! {
+    PyFixedTimeSpan(TimeSpan<Seconds>) as "FixedTimeSpan"
```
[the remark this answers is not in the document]
kalzoo: `Fixed` maps to `Seconds`, as in `type FixedTimeSpan = TimeSpan<Seconds>`. 

I chose `Fixed` in light of the next evolution here, that of `ParametricTimeSpan = TimeSpan<Expression>`

quil-rs/src/program/scheduling/schedule.rs
kalzoo: The dependencies are handled in the scheduling of `ScheduledBasicBlock.as_fixed_schedule`.

If instruction `GATE 0` expands into 

```
PULSE
DELAY
PULSE
```

Then this just unions/covers the times of the PULSE, DELAY, PULSE and considers that combined TimeSpan as the TimeSpan of the original gate

quil-py/quil/program/__init__.pyi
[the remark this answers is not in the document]
kalzoo: It might help to glance at the Rust or Python test cases for illustration

quil-py/quil/program/__init__.pyi
kalzoo: I've been trying to think of how to illustrate this in the docs.

namely, on calibration expansion, one instruction is decomposed into many. Those decomposed instructions will now interact with other decomposed instructions in a way that "mix" the two original gates in time. A FENCE "plays" at the first clock tick following any preceding instructions on its qubits; that may not immediately precede the PULSE instruction whose timing the user actually cares about (for example).

quil-rs/src/program/scheduling/schedule.rs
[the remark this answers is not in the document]
kalzoo: fixed

quil-py/quil/program/__init__.pyi
```diff
@@ -238,6 +314,73 @@ class CalibrationSet:
...
+@final
+class ControlFlowGraph:
+    """
+    Representation of a control flow graph (CFG) for a Quil program.
+
+    The CFG is a directed graph where each node is a basic block and each edge is a control flow
+    transition between two basic blocks.
+    """
+
+    def has_dynamic_control_flow(self) -> bool:
```
mhodson-rigetti, line 344: From a DbC perspective, does `False` here imply `len(self.basic_blocks()) == 1`? If yes, that would be worth documenting.

quil-py/quil/program/__init__.pyi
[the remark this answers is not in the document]
mhodson-rigetti: This makes sense to me as-is.

quil-py/quil/program/__init__.pyi
```diff
@@ -175,6 +175,82 @@ class Program:
...
+        (such as `FENCE`). This may cause certain instructions to appear as unexpectedly long in duration.
+
+        :param program: The program containing the calibrations to be used to schedule this block. Generally,
+            this should be the program from which the block was extracted
+        :param include_zero_duration_instructions: Whether to include instructions with zero duration in the schedule
+
+        The following example demonstrates construction of such a schedule for a simple program without explicit control
+        flow (and thus with only one basic block):
+
+        .. example-code::
```
mhodson-rigetti, line 208: This doesn't seem like an appropriate example; it's for `pyquil`, but `quil-py` should be able to be demonstrated stand-alone. An alternative might be something like:

```
from quil.program import Program

program = Program.parse("""
<some example instructions>
""")

print(program.to_quil())

control_flow_graph = program.control_flow_graph()
assert control_flow_graph.has_dynamic_control_flow() == False

basic_blocks = control_flow_graph.basic_blocks()
assert len(basic_blocks) == 1

schedule = blocks[0].as_fixed_schedule(program, False)
print(f"Duration = {schedule.duration()}")

print(schedule.items())
```

quil-py/quil/program/__init__.pyi
```diff
@@ -238,6 +314,73 @@ class CalibrationSet:
         """
         ...
 
+
+@final
+class FixedScheduleItem:
```
mhodson-rigetti, line 319: These classes could really benefit from some `__repr__()` or `__str__()` since the example code (including the update I suggested) above will just output the following when the schedule `.items()` is `print()`ed:

```
[<builtins.FixedScheduleItem object at 0x7fc726269440>, <builtins.FixedScheduleItem object at 0x7fc72626bd80>, <builtins.FixedScheduleItem object at 0x7fc7262694d0>, <builtins.FixedScheduleItem object at 0x7fc726269320>, <builtins.FixedScheduleItem object at 0x7fc72626b9f0>]
```

I think in this case `pyquil` can be a very thin wrapper and the majority of the functionality can reside here? How do you see division of responsibility between the two packages?

quil-py/quil/program/__init__.pyi
```diff
@@ -238,6 +314,73 @@ class CalibrationSet:
...
+        """
+    def duration(self) -> float:
+        """
+        The duration of the schedule, in seconds.
+
+        This is the maximum of the end time of all the items.
+        """
+
+@final
+class FixedTimeSpan:
```
mhodson-rigetti, line 367: To save the user the math, I'd recommend adding an `end()` property documented as a function of the other two. However, I can also see this being more appropriate to the `pyquil` plumbing, which is yet to come.

quil-py/quil/program/__init__.pyi
```diff
@@ -238,6 +314,73 @@ class CalibrationSet:
...
+    """
+    Representation of a time span in seconds.
+    """
+    @property
+    def start(self) -> float:
```
mhodson-rigetti, line 372: I completely understand why you would use `float`, in keeping `quil-rs` hardware agnostic. However, given the typical durations will be nanosecond or microsecond, and printing floats in a typical use case results in e.g. "5.6000000000000005e-08 to 9.600000000000001e-08" being the output, where do you see the practical issue of latching to some nanosecond or microsecond hardware "tick" solved?

quil-py/quil/program/__init__.pyi
mhodson-rigetti: Update: Using some appropriate calibration and frame definitions, and a short test case:
```
ISWAP 23 30
RX(pi) 23
RX(pi) 24
DELAY 23 24 2.48e-7
RX(pi) 24
```

Then bypassing the `__repr__()` comment the following follow-on code can create some reasonable output:
```
control_flow_graph = program.control_flow_graph()
assert control_flow_graph.has_dynamic_control_flow() == False

basic_blocks = control_flow_graph.basic_blocks()
assert len(basic_blocks) == 1

basic_block = basic_blocks[0]

instructions = basic_block.instructions()

schedule = basic_block.as_fixed_schedule(program, False)

print(f"Duration = {schedule.duration()}")

for schedule_item in schedule.items():
    start = schedule_item.time_span.start
    duration = schedule_item.time_span.duration
    end = start + duration

    instruction = instructions[schedule_item.instruction_index]

    print(f"{instruction} => {start} to {end}")
```

The output of the demonstration:
```
Gate(Gate { name: "RX", parameters: [PiConstant], qubits: [Fixed(24)], modifiers: [] }) => 0.0 to 4e-08
Gate(Gate { name: "RX", parameters: [PiConstant], qubits: [Fixed(23)], modifiers: [] }) => 5.6000000000000005e-08 to 9.600000000000001e-08
Gate(Gate { name: "ISWAP", parameters: [], qubits: [Fixed(23), Fixed(30)], modifiers: [] }) => 0.0 to 5.6000000000000005e-08
Gate(Gate { name: "RX", parameters: [PiConstant], qubits: [Fixed(24)], modifiers: [] }) => 3.44e-07 to 3.84e-07
Delay(Delay { duration: Number(Complex { re: 2.48e-7, im: 0.0 }), frame_names: [], qubits: [Fixed(23), Fixed(24)] }) => 9.600000000000001e-08 to 3.44e-07
```

Observations:
- Seeing `Gate` twice in the nested representations seems unnecessary?
- Floating point precision isn't ideal for a nanosecond-scale system; comment captured elsewhere.
- A complex numeric value doesn't seem to be the right type for the delay?

mhodson-rigetti (commented):
At a high level, I'd be looking for this facility to provide the following:
- For pulses, the start and end time and possibly the associated waveform (sharing the memory);
- For delays, the inferred start time of the delay, constraining the duration to the delay time specified by the user;
- For fences, the time at which the barrier applies; this is a point-in-time operation and should have zero duration.

I think there is more to unpack here but I don't want to block an MR that already provides users with helpful debugging information. I do think breaking changes are possible in subsequent versions so I defer to you @kalzoo on the merge priority in the short term.

quil-py/quil/program/__init__.pyi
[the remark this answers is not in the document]
BatmanAoD: Ah, this does say "end time" but I was thinking of durations rather than points in time. Makes sense now, thanks.

quil-py/quil/program/__init__.pyi
BatmanAoD: So zero-duration instructions can introduce new dependencies between previously-independent instructions?

quil-py/quil/program/__init__.pyi
kalzoo: This PR doesn't really make a decision there; durations are today already specified in floating-point seconds within the waveform templates, so this new feature just accumulates them. That is, if you were to round-trip `PULSE 0 "rf" flat(duration: 5.6e-8)` on `master`, you'd encounter the same floating-point rounding error.

They are latched by the consumer of the schedule, who can also generate a `Schedule<TimeUnit>` using a `TimeUnit` of their choice.

quil-py/quil/program/__init__.pyi
kalzoo: yep, I had considered it but left it off to keep the API minimal. Will add.

quil-py/quil/program/__init__.pyi
kalzoo: > Seeing Gate twice in the nested representations seems unnecessary?

Yeah, although this is not new to this PR - it's the `Debug` output from the Rust side.

> A complex numeric value doesn't seem to be the right type for the delay?

It's because it's an expression, and all numbers in `Expression` are represented as `Complex64`

kalzoo: > This doesn't seem like an appropriate example; it's for pyquil, but quil-py should be able to be demonstrated stand-alone

Of course! Can't believe I missed that. Thanks

quil-py/quil/program/__init__.pyi
kalzoo: I added a clarifying note

quil-py/quil/program/__init__.pyi
```diff
@@ -175,6 +175,121 @@ class Program:
...
+    def control_flow_graph(self) -> "ControlFlowGraph":
+        """
+        Return the `control flow graph`_ of the program.
+
+        .. _control flow graph: https://en.wikipedia.org/wiki/Control-flow_graph
+        """
+
+@final
+class BasicBlock:
+    def as_fixed_schedule(self, program: Program, include_zero_duration_instructions: bool) -> FixedSchedule:
```
MarquessV, line 187: Does `include_zero_duration_instructions` have a reasonable default? If so, might be more pythonic to make it an optional kwarg.

quil-py/src/program/scheduling.rs
```diff
@@ -0,0 +1,86 @@
...
```
MarquessV, line 65: I think it would be nice to include the unit in the type somewhere, e.g. `PyFixedTimeSpanSeconds`, even though I know it's mentioned in the docstrings.

quil-py/src/program/analysis.rs
```diff
@@ -0,0 +1,87 @@
...
+            .as_fixed_schedule(&program.0, include_zero_duration_instructions)
+            .map(|v| v.into())
+            .map_err(|e| PyValueError::new_err(e.to_string()))
+    }
+
+    pub fn gate_depth(&self, gate_minimum_qubit_count: usize) -> PyResult<usize> {
+        let block = BasicBlock::from(&self.0);
+        QubitGraph::try_from(&block)
+            .map(|graph| graph.gate_depth(gate_minimum_qubit_count))
+            .map_err(|e| PyValueError::new_err(e.to_string()))
```
MarquessV, line 56: You can use the `wrap_error!` and `py_wrap_error!` macros from rigetti-pyo3 to wrap and create a PyResult compatible error. You associate a Python error type (e.g. PyValueError) as part of their definition, so you don't have to repeat, and potentially mix up the error type when you map to it.

quil-py/src/program/scheduling.rs
```diff
@@ -0,0 +1,86 @@
...
+
+py_wrap_type! {
+    PyScheduledBasicBlock(ScheduledBasicBlockOwned) as "ScheduledBasicBlock"
+}
+
+impl_repr!(PyScheduledBasicBlock);
+
+impl PyScheduledBasicBlock {
+    pub fn as_fixed_schedule(&self, program: &PyProgram) -> PyFixedSchedule {
+        ScheduledBasicBlock::from(&self.0)
```
MarquessV, line 17: tuple access on the wrapped py types do work, but we typically use the as/into_inner methods. Would like to keep with that pattern for consistency.

quil-py/quil/program/__init__.pyi
mhodson-rigetti: Updated example looks good.

quil-py/quil/program/__init__.pyi
```diff
@@ -175,6 +175,121 @@ class Program:
...
+        """
+        Return a list of the instructions in the block, in order of definition.
+
+        This does not include the label or terminator instructions.
+        """
+    def terminator(self) -> Optional[Instruction]:
+        """
+        Return the control flow terminator instruction of the block, if any.
+
+        If this is ``None``, the implicit behavior is to "continue" to the subsequent block.
```
mhodson-rigetti, line 286: I feel like a little more explanation is required here; it took me a moment to deconflict concepts from the Wikipedia article on CFGs to what is happening in Quil. Each basic block can have multiple output arcs- in Quil it's probably limited to 2 (or 1 and it continues to the next block) and this happens only on conditional jump instructions? It would be worth saying that here, and it would be more semantically rich if there was a Quil control flow subclass of `Instruction` that was the return type here that helped you see it was only jump instructions.

mhodson-rigetti: Also, is this not by definition `self.instructions[-1]`? Or is the terminator excluded from `instructions`? If included, I can see why you would still have this method, because the naming provides a sort of semantic richness that is valuable (ie. same argument for having the "end" method on the span; its sort of canonical to the definition).

quil-py/quil/program/__init__.pyi
mhodson-rigetti: When I do this now, I see:

```
[ComputedScheduleItem { time_span: TimeSpan { start_time: Seconds(3.44e-7), duration: Seconds(4e-8) }, instruction_index: 4 }, ComputedScheduleItem { time_span: TimeSpan { start_time: Seconds(9.600000000000001e-8), duration: Seconds(2.48e-7) }, instruction_index: 3 }, ComputedScheduleItem { time_span: TimeSpan { start_time: Seconds(5.6000000000000005e-8), duration: Seconds(4e-8) }, instruction_index: 1 }, ComputedScheduleItem { time_span: TimeSpan { start_time: Seconds(0.0), duration: Seconds(4e-8) }, instruction_index: 2 }, ComputedScheduleItem { time_span: TimeSpan { start_time: Seconds(0.0), duration: Seconds(5.6000000000000005e-8) }, instruction_index: 0 }]
```

I was expecting something close to an idiomatic output of a Python dataclass. Here I see the underlying Rust class name, which has some different member names and is missing some of the inferred members (like the end time). Can it be fixed?

quil-py/quil/program/__init__.pyi
mhodson-rigetti: Note is good!

quil-py/quil/program/__init__.pyi
mhodson-rigetti: Looks good.

quil-py/quil/program/__init__.pyi
mhodson-rigetti: OK cool. Doing something like latching to a ns and getting an integer could be interesting. Future feature?

mhodson-rigetti (commented):
@kalzoo I finished a second, more thorough review of the API.

quil-py/quil/program/__init__.pyi
```diff
@@ -175,6 +175,121 @@ class Program:
...
+                    FENCE 0
+                    PULSE 0 1 "a" flat(duration: 1.0)
+
+                A 0
+                B 0
+
+        If ``include_zero_duration_instructions`` is ``True``, ``B 0`` will be scheduled from time 0 to time 2, because it includes the time
+        of the FENCE.
+
+        If not, it will be from time 1 to time 2, considering only the non-zero duration instructions (here, ``PULSE``) within the calibration.
```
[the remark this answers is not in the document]
kalzoo: > So it would have a duration of 1.

I don't agree with this - nothing plays in between the FENCE and its paired PULSE, but it "ends" at the same time it starts, 0, and that's how Quil scheduling interprets it as well (for instructions which depend on it). By definition it has no duration - it's just a marker between other elements on the timeline.

> I don't think the case of users seeing the "unusually long gate times" could make any useful debugging sense

I agree, this may not, but someone (can't recall if it was you) recently specifically asked to "see when their PULSEs, CAPTUREs, and FENCEs were scheduled"), and this flag _also_ affects top-level FENCEs within the pre-expansion program. If those are missing, I'll probably hear about it.

I agree that this is a coarse implementation though. As far as seeing the "member" instructions, yeah,we could (in another PR) expose the source mapping used to actually compute the gate-model schedule.

quil-py/quil/program/__init__.pyi
```diff
@@ -175,6 +175,121 @@ class Program:
...
+                    PULSE 1 "a" flat(duration: 1.0)
+
+                # The FENCE "plays" immediately upon program start at time 0
+                # The `PULSE` is blocked by the `PULSE` of `A 0` on intersecting frame 1 "a", and does not play until time 1
+                DEFCAL B 0:
+                    FENCE 0
+                    PULSE 0 1 "a" flat(duration: 1.0)
+
+                A 0
+                B 0
```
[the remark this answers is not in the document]
kalzoo: true, good points

quil-py/quil/program/__init__.pyi
kalzoo: good point; at @mhodson-rigetti 's recommendation though I'll just end up removing it

quil-py/src/program/scheduling.rs
kalzoo: True, I should do the same on the Rust side.

quil-py/quil/program/__init__.pyi
kalzoo: >  Can it be fixed?

Yes, but/and:

* ...but it's not new to this PR; that's how `impl_repr!` implements it for all of our Python-bound types across all relevant Rigetti packages
* ...and that could make it quick & easy to fix if that makes a difference

quil-py/quil/program/__init__.pyi
```diff
@@ -175,6 +175,121 @@ class Program:
...
+
+        This does not include the label or terminator instructions.
+        """
+    def terminator(self) -> Optional[Instruction]:
+        """
+        Return the control flow terminator instruction of the block, if any.
+
+        If this is ``None``, the implicit behavior is to "continue" to the subsequent block.
+        """
+    def topological_swap_count(self) -> int:
```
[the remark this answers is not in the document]
kalzoo: removed

quil-py/src/program/scheduling.rs
```diff
@@ -0,0 +1,98 @@
...
+
+wrap_error!(RustComputedScheduleError(ComputedScheduleError));
+py_wrap_error!(
+    quil,
+    RustComputedScheduleError,
+    PyComputedScheduleError,
+    PyValueError
+);
+
+impl PyScheduledBasicBlock {
```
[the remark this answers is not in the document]
kalzoo: Even better - it's not fully implemented and not practically useful on the Python side so I'll just remove it.

quil-py/quil/program/__init__.pyi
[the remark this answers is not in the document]
kalzoo: @mhodson-rigetti I'm confused about something in this thread after re-reading your original review on this PR:

1. it was you who had said that one of your success criteria for this was specifically seeing the FENCE timings:

> For fences, the time at which the barrier applies

2. Then, you had also called out that FENCE is zero-duration:

> this is a point-in-time operation and should have zero duration

----

I think this can be resolved by, as you say, expanding calibrations and returning the source mapping so that the user can trace back "expanded instruction time" to "original unexpanded instruction" for all instructions, and then decide how to handle that information themselves. I'll save that extension for the next PR.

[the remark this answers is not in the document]
mhodson-rigetti: True, I wasn't being very consistent. I'm imagining what I want to visualize with this data; for a FENCE, I would draw a barrier - a vertical line on a timeline - at the latest time of any frame involved in the fence. This is "the time of the fence" if you wanted to assign it a single time. However, from a frame tracking perspective, a given frame affected by a fence goes from whatever time it is at (e.g. the time the previous pulse finished) up to this barrier time. This is basically a delay, and it might be handy to have captured that explicitly. However the data structures might not support this, and giving a FENCE a duration like this only came up when I was tracking how I would explain your new example to someone- I'd be saying that the FENCE caused "b" frame to wait from 0 to 1. I will need to deconflict this viewpoint when I write the needs statement.

quil-py/quil/program/__init__.pyi
mhodson-rigetti: Great; a distinction worth documenting?

quil-py/quil/program/__init__.pyi
mhodson-rigetti: Understood. I feel its confusing so definitely a nice-to-have but outside the scope of this feature (or library!).

quil-py/quil/program/__init__.pyi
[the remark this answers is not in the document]
mhodson-rigetti: Updated example noted. All good.

quil-py/quil/program/__init__.pyi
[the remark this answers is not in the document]
mhodson-rigetti: I like the tweaks you made to the example and documentation- simple and effective.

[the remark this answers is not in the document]
kalzoo: @mhodson-rigetti after some more thought on this, I'd like to omit the flag (as you suggested and as it is now) but instead turn the default behavior to include _all_ instructions (even zero-duration) in the rolled-up timespan, because:

* While both ways can be surprising, treating all instructions equally is probably less surprising
* It suits the common case and thus will be less _frequently_ surprising as well. In most cases, FENCE shouldn't lead or lag the calibration body as in the pathological example here
* It means that pre-expansion FENCEs will show up in the overall schedule
* It means that SET- and SHIFT- instructions will as well, which lets us treat this data structure as an IR of sorts
* I intend to give the user more control over the details (as in this comment thread), so they'll have the ability to decide on a per-instruction basis what gets "rolled up" soon anyway
* Finally, what swayed me: it means that this will work as you might expect for an already-expanded pulse program; every instruction in the program will get a time span. if we skip zero-duration, some will just be missing, to no benefit.

I'd leave the illustration here to help users identify this behavior when it does happen.

Does that check out with you?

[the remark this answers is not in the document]
mhodson-rigetti: Yeah I'm totally comfortable with that. Good call.

PR #339, opened 2024-02-09
Extend ProgramStats PR

kalzoo:
Extends #336 with:

* refactoring to better organize files under the `program` module. Note that all the shiny new stuff is now at `program::analysis`. I'd like to re-export structs from that level rather than make submodules private, for user benefit. A user shouldn't have to burrow more than 2 levels deep here, anything deeper is an implementation detail.
* `ControlFlowGraph` and `BasicBlock` which are neatly constructed from (and which borrow) a `Program`
  * then, a refactor of `ScheduledProgram` with `InstructionBlock`s to `ScheduledControlFlowGraph` with `ScheduledBasicBlock`s
  * This change is obvious in hindsight. The ScheduledProgram was overloaded with two functions: control flow and Quil pulse program vector clock construction. These are distinct. Combining them into one pass means that programs which contain gates and other non-pulse-level instructions but which benefit from control flow analysis
  * Differences from the previous InstructionBlock implementation:
    * Unlabeled blocks now don't have labels auto-generated but rather have a label of `None`. Blocks are stored in a Vec rather than an index map. Blocks which do not end with a terminator now `Continue` into the block of succeeding index rather than `Jump`ing there. This is more intuitive behavior.
    * ControlFlowGraph construction is _much_ simpler and easier to follow logically than `ScheduledProgram::from_program` was
    * It now supports Label::Placeholder, which makes sense; this leaves it to downstream consumers to error on those placeholders when not allowed. This isn't of much practical value because in most use cases today the program is serialized between construction in `quil-rs` and the next consumer...but in principle, this is the right thing to do.
    * Small changes in how the dot format graph is labeled and serialized.
 * Implements `From<BasicBlock> for ProgramStats` to follow the existing pattern in #336 
   * Note: this feels less than ergonomic. See below for my proposal for improvement. 

# What's not done here (and can be done if merged into #336):

* Correct the `pub` interface and correctly mark PR as breaking
* Documentation

# Proposal for way forward on stats

* first, `metrics` or `analysis` sounds more correct than `stats` to me; I hear the standard statistics like `mean`, `median`, etc when I hear `stats`

The following metrics make sense for their respective targets:

### Program via ControlFlowGraph

* width
* uses_dynamic_control_flow: walk the blocks and look at their terminators. if any conditional jumps, then `true`, else `false`
* when no dynamic control flow: then the BasicBlock metrics make sense, because they can be folded into a single value because the execution flow is known.

## BasicBlock

* width
* swap_count
* gate_volume
* qubits_used -> I don't think this needs to be cached, that adds complexity and unclear value. Caching on `Program` was key because of how and when it was called; I'd expect this to be called exactly once in the lifetime of `ProgramStats`

### ExecutionGraph wrapping BasicBlock

_I might rename this to something more specific like `QubitGraph`_

* depth
* fidelity_estimate

## ScheduledBasicBlock

* duration -> end time of flattened pulse program (TBD in a different, hefty PR)

### None of the above

* instruction_count seems unnecessary given how easy `body_instructions().len()` is

# Summary

So, given the above proposal, I don't know that traits or wrapper structs are worth the complexity (except for `ExecutionGraph`). Each respective type can just implement its methods directly because even the same method might have different return types (e.g. `Program#depth` vs `BasicBlock#depth` will have different failure modes and thus errors). To get a program's depth, the user might call `ExecutionGraph::from(self.get_only_basic_block()?).depth()`. I don't yet have a neat picture for how the user could efficiently calculate `program.depth(); program.expected_fidelity()` because that needs, in the case of multiple blocks without DCF, to construct an `ExecutionGraph` around each `BasicBlock` and persist that between calls.


quil-rs/src/program/analysis/control_flow_graph.rs
kalzoo, line 1: This is a key new file

quil-rs/src/program/analysis/stats/mod.rs
```diff
@@ -66,15 +68,41 @@ impl InstructionsSource for Program {
     }
 }
 
+impl<'p> InstructionsSource for BasicBlock<'p> {
```
kalzoo, line 71: This is to fit the existing pattern, although would not be how I'd do it from scratch; see description

quil-rs/src/program/scheduling/graph.rs
```diff
@@ -550,160 +474,18 @@ impl<'a> ScheduledProgram<'a> {
...
-            &mut working_instructions,
-            &mut blocks,
-            &mut working_label,
-            program,
-            None,
-            custom_handler,
-        )?;
-
-        Ok(ScheduledProgram { blocks })
+        let control_flow_graph = program.as_control_flow_graph();
```
kalzoo, line 477: Note how much cleaner the implementation in `as_control_flow_graph` is than this one.

Also, that new implementation tolerates/elides more instructions than this previous one did. Some of those end up in errors within `ScheduledBasicBlock` because of lack of calibrations, but `DEFCAL` and friends are allowed now where they were once errors.

quil-rs/src/program/scheduling/graphviz_dot.rs
```diff
@@ -100,91 +140,182 @@ impl<'a> ScheduledProgram<'a> {
...
-                                        condition.to_quil_or_debug(),
-                                        equality_operators.1
-                                    )
-                                    .as_str(),
+                                    format!("if {} == 0", condition.to_quil_or_debug(),).as_str(),
                                 );
                         };
                     }
-                    BlockTerminator::Unconditional { target } => {
+                    // BasicBlockTerminator::Conditional {
```
kalzoo, line 245: todo: I want to come back to this and rework `BasicBlockTerminator::JumpWhen` and `::JumpUnless` into a single enum similar to this previous `Conditional`.

quil-rs/src/program/analysis/control_flow_graph.rs
```diff
@@ -0,0 +1,153 @@
...
+            };
+            graph.blocks.push(block);
+        }
+
+        graph
+    }
+
+    pub fn get_first_basic_block(&self) -> Option<BasicBlock> {
+        self.as_control_flow_graph().blocks.into_iter().next()
+    }
```
BatmanAoD, line 139: This seems like sort of a surprising thing to do: construct the entire graph, then discard most of it. Do you think this functionality will be needed outside of tests?

quil-rs/src/program/analysis/control_flow_graph.rs
```diff
@@ -0,0 +1,153 @@
...
+        if blocks.len() == 1 {
+            Ok(blocks.into_iter().next().unwrap())
+        } else {
+            Err(ProgramContainsMultipleBasicBlocks)
+        }
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+#[error("Program contains multiple basic blocks")]
```
BatmanAoD, line 152:
```suggestion
#[error("Program is empty or contains multiple basic blocks")]
```

quil-rs/src/program/analysis/stats/mod.rs
BatmanAoD: It seems like `BasicBlock` is effectively what I was trying to capture with `InstructionsSource` anyway.

quil-rs/src/program/analysis/stats/mod.rs
```diff
@@ -66,15 +68,41 @@ impl InstructionsSource for Program {
...
+
 struct InstructionBlockWithQubitSet<'a, 'b> {
-    block: &'a InstructionBlock<'b>,
+    block: &'a BasicBlock<'b>,
     qubits: HashSet<&'a Qubit>,
 }
 
-impl<'a, 'b> From<&'a InstructionBlock<'b>> for InstructionBlockWithQubitSet<'a, 'b> {
-    fn from(block: &'a InstructionBlock<'b>) -> Self {
+impl<'a, 'b> From<&'a ScheduledBasicBlock<'b>> for InstructionBlockWithQubitSet<'a, 'b> {
```
BatmanAoD, line 96: Oh, `InstructionBlockWithQubitSet` was just a hack to let `InstructionBlock` work as an instruction-source.

quil-rs/src/program/scheduling/graphviz_dot.rs
BatmanAoD: Should that be considered a future `todo` or something to be done as part of #334?

BatmanAoD (approved):
Seems reasonable.

What consumers are there that will need to be updated for the breaking change?

kalzoo:
> Seems reasonable.
> 
> What consumers are there that will need to be updated for the breaking change?

Only internal ones, to my knowledge. I can validate that before we merge #334

quil-rs/src/program/analysis/control_flow_graph.rs
kalzoo: yeah this is not something we want to ship, but I do think it's quite useful as a convenience function to let the user treat the program like a block and get the metrics they want from it. Your new `TryFrom` implementation is just as good though

PR #342, opened 2024-02-27
Revert "fix(quil-rs): match exactly one qubit for DELAYs without frame specifier"

kalzoo:
Reverts rigetti/quil-rs#300

The existing implementation appears to have been correct:

```
DELAY 0 1 matches frame 0 "rf" -> no

DELAY 0 matches frame 0 "rf" -> yes

DELAY 0 1 matches frame 0 1 "rf" -> yes

DELAY 0 1 matches frame 0 "rf" -> no
```

Quil spec:

> If frame names are specified, then the delay instruction affects those frames on those qubits. If no frame names are specified, all frames on precisely those qubits are affected.

```
match := {frame_qubits} == {delay_qubits}
```

### DIFF
--- a/quil-rs/src/program/mod.rs
+++ b/quil-rs/src/program/mod.rs
@@ -827,24 +827,18 @@ DECLARE ec BIT
 
     #[test]
     fn frame_blocking() {
-        let input = r#"DEFFRAME 0 "a":
-    HARDWARE-OBJECT: "hardware"
+        let input = "DEFFRAME 0 \"a\":
+\tHARDWARE-OBJECT: \"hardware\"
 
-DEFFRAME 0 "b":
-    HARDWARE-OBJECT: "hardware"
+DEFFRAME 0 \"b\":
+\tHARDWARE-OBJECT: \"hardware\"
 
-DEFFRAME 1 "c":
-    HARDWARE-OBJECT: "hardware"
+DEFFRAME 1 \"c\":
+\tHARDWARE-OBJECT: \"hardware\"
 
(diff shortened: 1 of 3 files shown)


quil-rs/src/program/mod.rs
```diff
@@ -827,24 +827,18 @@ DECLARE ec BIT
...
-        let input = r#"DEFFRAME 0 "a":
-    HARDWARE-OBJECT: "hardware"
-
-DEFFRAME 0 "b":
-    HARDWARE-OBJECT: "hardware"
-
-DEFFRAME 1 "c":
-    HARDWARE-OBJECT: "hardware"
+        let input = "DEFFRAME 0 \"a\":
+\tHARDWARE-OBJECT: \"hardware\"
```
mhodson-rigetti, line 831: This looks like hard tabs were injected? Is this intentional?

mhodson-rigetti (commented):
The key change in `frame.rs` looks right to me. I assume the test case updates are by a revert so they should be right (though I have trouble reading them). Not sure why clippy is not happy in CI?

quil-rs/src/program/mod.rs
MarquessV: I believe so, the commit this is reverting originally changed them from tabs to spaces: https://github.com/rigetti/quil-rs/commit/18081daa34acd090b1daaabf3c5a8c46d13f6993#diff-02ccf5616a656b995be52b5d351335deded8c166cf246b234a0bfa1f33162c1eL723

PR #348, opened 2024-03-26
feat(quil-cli): Add CLI for interacting with quil-rs

### LINKED ISSUE #349: Create a CLI for quil-rs
quil-rs is currently only accessible as a Rust package or Python library. This is great for its primary use case, but can be cumbersome if you just want to see how quil-rs parses something.

Parsing programs and expressions is a good starting point, as it's something I want to test often. However, we can add in things like program analysis, calibration expansion, expression simplification, and more as needed.

MarquessV:
I often find myself wanting to quickly test the behavior of the parser. Right now that requires setting up some boilerplate in either pyQuil or quil-rs to go through the parser. This CLI exposes the functionality more directly, allowing me to bisect issues quicker.

In addition to using the parser, I see this as a useful entry point for the recent program analysis feature. It also enables easier benchmarking with tools like hyperfine. Not that the existing benchmarks aren't sufficient, but it makes benching new programs or expressions a bit more accessible and faster in the iteration phase.

closes #349

### DIFF
--- a/quil-cli/Cargo.toml
+++ b/quil-cli/Cargo.toml
@@ -2,10 +2,11 @@
 name = "cli"
 version = "0.1.0"
 edition = "2021"
+license = "Apache-2.0"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-quil-rs = {path = "../quil-rs"}
+quil-rs = { path = "../quil-rs", version = "0.23.0" }
 clap =  {version = "4.5.4", features = ["derive"]}
 anyhow = "1.0.81"
--- a/quil-cli/Cargo.toml
+++ b/quil-cli/Cargo.toml
@@ -1,0 +1,11 @@
+[package]
+name = "cli"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+quil-rs = {path = "../quil-rs"}
+clap =  {version = "4.5.4", features = ["derive"]}
+anyhow = "1.0.81"
--- a/quil-cli/Cargo.toml
+++ b/quil-cli/Cargo.toml
@@ -1,5 +1,5 @@
 [package]
-name = "cli"
+name = "quil-cli"
 version = "0.1.0"
 edition = "2021"
 license = "Apache-2.0"
--- a/quil-cli/src/main.rs
+++ b/quil-cli/src/main.rs
@@ -11,6 +11,7 @@ struct Cli {
 
 #[derive(Subcommand, Clone, Debug)]
 pub enum Command {
+    /// Parse a Quil program or expression
     Parse {
         #[arg(short = 't', long = "type")]
         input_type: Option<InputType>,
@@ -20,8 +21,10 @@ pub enum Command {
 
 #[derive(ValueEnum, Clone, Debug, Default)]
 pub enum InputType {
+    /// Parse a Quil program (default)
     #[default]
     Program,
+    /// Parse a Quil expression
     Expression,
 }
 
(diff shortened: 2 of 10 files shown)


kalzoo:
lgtm, pending CI 👍 seems useful.

don't forget linked issues

quil-cli/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "cli"
```
jselig-rigetti, line 2: Will this allow the package to be `cargo installed` as a binary?

Shadow53: Also curious what name this would get installed under. Regardless, if we ever publish this to crates.io, we'll want this to be `quil-cli`, `quil-parser-cli`, or similar

quil-cli/src/main.rs
```diff
@@ -0,0 +1,55 @@
...
+    Ok(())
+}
+
+fn handle_parse(input_type: InputType, input: String) -> anyhow::Result<()> {
+    let parsed = match input_type {
+        InputType::Program => {
+            Program::from_str(&input).context("Failed to parse program from input string.")?.to_quil().expect("Parsed Program from valid Quil string, should be able to convert it back to valid Quil")
+        }
+        InputType::Expression => Expression::from_str(&input)
+            .context("Failed to parse expression from input string.")?.to_quil().expect("Parsed Expression from valid Quil expression, should be able to convert it back to valid Expression"),
```
Shadow53, line 49: I know we shouldn't expect to run into the `expect` calls, but since we're already using `anyhow`, I think it makes sense to just make them errors. That way, if something unexpected happens, the user experience is consistent.

quil-rs/src/reserved.rs
```diff
@@ -13,6 +13,7 @@ pub enum ReservedToken {
 }
 
 #[derive(Clone, Debug)]
+#[allow(dead_code)] // clippy 0.1.77 incorrectly identifies this NewType as dead code.
 pub struct NotReservedToken(String);
```
Shadow53, line 17: More specifically, it's that the inner (private) `String` is never accessed.

Since this is used as an error, it should implement `Error`:

```suggestion
#[derive(Clone, Debug, thiserror::Error)]
#[error("{0} is not a reserved token")]
pub struct NotReservedToken(String);
```

Shadow53: If the above suggestion is applied, also need to change the `Err` case below to `Err(NotReservedToken(s.to_string()))`

Shadow53 (changes requested):

quil-cli/Cargo.toml
MarquessV: I went ahead and named it `quil-cli`, since I could see things like program analysis being added to the CLI.

quil-cli/src/main.rs
```diff
@@ -0,0 +1,62 @@
...
+    #[clap(subcommand)]
+    command: Command,
+}
+
+#[derive(Subcommand, Clone, Debug)]
+pub enum Command {
+    /// Parse a Quil program or expression
+    Parse {
+        #[arg(short = 't', long = "type")]
+        input_type: Option<InputType>,
```
Shadow53, line 17: Same concept, but more direct, imo.

```suggestion
        #[arg(short = 't', long = "type", default_value_t)]
        input_type: InputType,
```

quil-cli/src/main.rs
```diff
@@ -0,0 +1,62 @@
...
+    /// Parse a Quil expression
+    Expression,
+}
+
+fn main() -> anyhow::Result<()> {
+    let cli = Cli::parse();
+
+    match cli.command {
+        Command::Parse { input_type, input } => {
+            handle_parse(input_type.unwrap_or_default(), input)?
```
Shadow53, line 36: With above

```suggestion
            handle_parse(input_type, input)?
```

Shadow53 (approved):
One suggestion (split into two), but looks good otherwise.

PR #352, opened 2024-04-08
fix!: CalibrationSet's and Program's will be considered equal if they contain the same set of calibrations, regardless of order.

### LINKED ISSUE #351: Program equality is sensitive to calibration ordering
When checking if two programs are equal, calibrations should be compared as orderless sets. Calibrations are global state, so the order of them should not matter.

`quil-rs` current `Program` equality logic respects the order of calibrations (and possibly waveforms), which is incorrect.

MarquessV:
I've reworked `CalibrationSet` to actually be a set, where existence in the set is keyed on the calibrations "signature" (e.g. the fields determine the instructions they expand into). In addition, the set maintains insertion order, but disregards order during equality checks. This is to keep program serialization deterministic, while recognizing that the order of calibrations doesn't matter in the semantic meaning of the program.

This required a custom data structure for a few reasons:

1. We care about calibration equality on two different levels: "signature" and equality of contents. This makes using something like an `IndexSet` impossible. To prevent duplicate signatures in a set, we'd have to defined `Hash` and thus `Eq` in terms of the signature. However, when comparing the equality of two calibration sets, we care both about the signature and the contents of the calibration, which would conflict with the definition of `Eq` needed to support using a set.
2. We can't use a `HashMap`, because program serialization wouldn't be deterministic. `BTreeMap` isn't supported by calibrations because there is no clear way to define `Ord` for them, and `IndexMap` defines equality on the order of contents in the map, which leaves us with the same problem we are trying to solve.

Closes #351

### DIFF
--- a/quil-rs/src/instruction/calibration.rs
+++ b/quil-rs/src/instruction/calibration.rs
@@ -10,7 +10,9 @@ use crate::{
 use super::write_qubit_parameters;
 
 pub trait CalibrationSignature {
-    type Signature<'a>: where Self: 'a;
+    type Signature<'a>
+    where
+        Self: 'a;
 
     fn signature(&self) -> Self::Signature<'_>;
     fn has_signature(&self, signature: &Self::Signature<'_>) -> bool;
@@ -80,9 +82,7 @@ impl CalibrationSignature for Calibration {
 
     fn has_signature(&self, signature: &Self::Signature<'_>) -> bool {
         let (name, parameters, qubits) = signature;
-        self.name == *name
-            && self.parameters == *parameters
-            && self.qubits == *qubits
+        self.name == *name && self.parameters == *parameters && self.qubits == *qubits
     }
 }
 
--- a/quil-rs/src/instruction/calibration.rs
+++ b/quil-rs/src/instruction/calibration.rs
@@ -9,6 +9,12 @@ use crate::{
 
 use super::write_qubit_parameters;
 
+pub trait CalibrationSignature {
+    type Signature: PartialEq;
+
+    fn signature(&self) -> Self::Signature;
+}
+
 #[derive(Clone, Debug, Default, PartialEq)]
 pub struct Calibration {
     pub instructions: Vec<Instruction>,
@@ -60,6 +66,18 @@ impl Quil for Calibration {
     }
 }
 
+impl CalibrationSignature for Calibration {
+    type Signature = (String, Vec<Expression>, Vec<Qubit>);
+
+    fn signature(&self) -> Self::Signature {
+        (
+            self.name.clone(),
+            self.parameters.clone(),
+            self.qubits.clone(),
+        )
+    }
+}
+
 #[derive(Clone, Debug, PartialEq)]
 pub struct MeasureCalibrationDefinition {
     pub qubit: Option<Qubit>,
@@ -74,6 +92,14 @@ impl MeasureCalibrationDefinition {
             parameter,
             instructions,
         }
+    }
+}
+
+impl CalibrationSignature for MeasureCalibrationDefinition {
+    type Signature = (Option<Qubit>, String);
+
+    fn signature(&self) -> Self::Signature {
+        (self.qubit.clone(), self.parameter.clone())
     }
 }
 
--- a/quil-rs/src/instruction/calibration.rs
+++ b/quil-rs/src/instruction/calibration.rs
@@ -10,9 +10,10 @@ use crate::{
 use super::write_qubit_parameters;
 
 pub trait CalibrationSignature {
-    type Signature: PartialEq;
+    type Signature<'a>: where Self: 'a;
 
-    fn signature(&self) -> Self::Signature;
+    fn signature(&self) -> Self::Signature<'_>;
+    fn has_signature(&self, signature: &Self::Signature<'_>) -> bool;
 }
 
 #[derive(Clone, Debug, Default, PartialEq)]
@@ -67,14 +68,21 @@ impl Quil for Calibration {
 }
 
 impl CalibrationSignature for Calibration {
-    type Signature = (String, Vec<Expression>, Vec<Qubit>);
+    type Signature<'a> = (&'a str, &'a [Expression], &'a [Qubit]);
 
-    fn signature(&self) -> Self::Signature {
+    fn signature(&self) -> Self::Signature<'_> {
         (
-            self.name.clone(),
-            self.parameters.clone(),
-            self.qubits.clone(),
+            self.name.as_str(),
+            self.parameters.as_slice(),
+            self.qubits.as_slice(),
         )
+    }
+
+    fn has_signature(&self, signature: &Self::Signature<'_>) -> bool {
+        let (name, parameters, qubits) = signature;
+        self.name == *name
+            && self.parameters == *parameters
+            && self.qubits == *qubits
     }
 }
 
@@ -96,10 +104,15 @@ impl MeasureCalibrationDefinition {
 }
 
 impl CalibrationSignature for MeasureCalibrationDefinition {
-    type Signature = (Option<Qubit>, String);
+    type Signature<'a> = (Option<&'a Qubit>, &'a str);
 
-    fn signature(&self) -> Self::Signature {
-        (self.qubit.clone(), self.parameter.clone())
+    fn signature(&self) -> Self::Signature<'_> {
+        (self.qubit.as_ref(), self.parameter.as_str())
+    }
+
+    fn has_signature(&self, signature: &Self::Signature<'_>) -> bool {
+        let (qubit, parameter) = signature;
+        self.qubit.as_ref() == *qubit && self.parameter == *parameter
     }
 }
 
--- a/quil-rs/src/program/calibration.rs
+++ b/quil-rs/src/program/calibration.rs
@@ -14,6 +14,7 @@
 
 use std::collections::HashMap;
 
+use indexmap::IndexMap;
 use itertools::FoldWhile::{Continue, Done};
 use itertools::Itertools;
 
@@ -33,8 +34,8 @@ use super::ProgramError;
 /// A collection of Quil calibrations (`DEFCAL` instructions) with utility methods.
 #[derive(Clone, Debug, Default, PartialEq)]
 pub struct CalibrationSet {
-    pub calibrations: HashMap<String, Calibration>,
-    pub measure_calibrations: HashMap<(String, Option<Qubit>), MeasureCalibrationDefinition>,
+    pub calibrations: IndexMap<(String, Vec<Qubit>), Calibration>,
+    pub measure_calibrations: IndexMap<(String, Option<Qubit>), MeasureCalibrationDefinition>,
 }
 
 struct MatchedCalibration<'a> {
@@ -386,8 +387,10 @@ impl CalibrationSet {
     /// If a calibration with the same name already exists in the set, it will be replaced, and the
     /// old Calibration is returned. Otherwise, None is returned.
     pub fn insert_calibration(&mut self, calibration: Calibration) -> Option<Calibration> {
-        self.calibrations
-            .insert(calibration.name.clone(), calibration)
+        self.calibrations.insert(
+            (calibration.name.clone(), calibration.qubits.clone()),
+            calibration,
+        )
     }
 
     /// Insert a [`MeasureCalibration`] into the set.
--- a/quil-rs/src/program/calibration.rs
+++ b/quil-rs/src/program/calibration.rs
@@ -27,13 +27,13 @@ use crate::{
     },
 };
 
-use super::{ProgramCalibrationSet, ProgramError};
+use super::{CalibrationSet, ProgramError};
 
 /// A collection of Quil calibrations (`DEFCAL` instructions) with utility methods.
 #[derive(Clone, Debug, Default, PartialEq)]
-pub struct CalibrationSet {
-    pub calibrations: ProgramCalibrationSet<Calibration>,
-    pub measure_calibrations: ProgramCalibrationSet<MeasureCalibrationDefinition>,
+pub struct Calibrations {
+    pub calibrations: CalibrationSet<Calibration>,
+    pub measure_calibrations: CalibrationSet<MeasureCalibrationDefinition>,
 }
 
 struct MatchedCalibration<'a> {
@@ -57,7 +57,7 @@ impl<'a> MatchedCalibration<'a> {
     }
 }
 
-impl CalibrationSet {
+impl Calibrations {
     /// Return a vector containing a reference to all [`Calibration`]s in the set.
     pub fn calibrations(&self) -> Vec<&Calibration> {
         self.calibrations.iter().collect()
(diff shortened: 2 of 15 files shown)


quil-rs/src/program/calibration.rs
```diff
@@ -33,8 +33,8 @@ use super::ProgramError;
 /// A collection of Quil calibrations (`DEFCAL` instructions) with utility methods.
 #[derive(Clone, Debug, Default, PartialEq)]
 pub struct CalibrationSet {
-    pub calibrations: Vec<Calibration>,
-    pub measure_calibrations: Vec<MeasureCalibrationDefinition>,
+    pub calibrations: HashMap<String, Calibration>,
+    pub measure_calibrations: HashMap<(String, Option<Qubit>), MeasureCalibrationDefinition>,
```
Shadow53, line 37: According to [the spec](https://quil-lang.github.io/#12-6Defining-Calibrations), a calibration "key" should be a combination of name, parameters, and qubits. Essentially, it seems that the entire string after `DEFCAL` is important for disambiguating calibrations.

Shadow53: Should also note that any tests using `Debug` output, e.g. with `insta`, may have unexpected test failures due to the random ordering of `HashMap`. (This can be worked around with `pretty_assertions_sorted`, but worth noting).

Possibly a better solution is to keep it as a `Vec`, but make the fields private and use methods to ensure that the items in the `Vec` are always sorted. Using `Vec::sort_by(_cached)_key` makes the most sense to me.

MarquessV:
Remaining CI checks will be fixed when #348 can be merged.

quil-rs/src/program/calibration.rs
MarquessV: Turns out existing maps and sets don't really handle the particular needs of how we want these calibration sets to work w/ a program, so I created a custom set, backed by a Vec.

MarquessV:
cc: @mhodson-rigetti does this seem sound to you? Does it meet your expectations of how calibrations should work in a program?

mhodson-rigetti:
> cc: @mhodson-rigetti does this seem sound to you? Does it meet your expectations of how calibrations should work in a program?

Yes, this describes the expected behavior perfectly.

quil-rs/src/program/calibration.rs
```diff
@@ -58,15 +57,26 @@ impl<'a> MatchedCalibration<'a> {
...
 }
 
-impl CalibrationSet {
-    /// Return all [`Calibration`]s in the set
-    pub fn calibrations(&self) -> &[Calibration] {
-        self.calibrations.as_ref()
+impl Calibrations {
+    /// Return a vector containing a reference to all [`Calibration`]s in the set.
+    pub fn calibrations(&self) -> Vec<&Calibration> {
+        self.calibrations.iter().collect()
```
Shadow53, line 63:
```suggestion
    /// Return a vector containing a reference to all [`Calibration`]s in the set.
    pub fn calibrations(&self) -> Vec<&Calibration> {
        self.iter_calibrations().collect()
```

quil-rs/src/program/calibration.rs
```diff
@@ -58,15 +57,26 @@ impl<'a> MatchedCalibration<'a> {
...
     }
 
-    /// Return all [`MeasureCalibrationDefinition`]s in the set
-    pub fn measure_calibrations(&self) -> &[MeasureCalibrationDefinition] {
-        self.measure_calibrations.as_ref()
+    /// Return a vector containing a reference to all [`MeasureCalibrationDefinition`]s
+    /// in the set.
+    pub fn measure_calibrations(&self) -> Vec<&MeasureCalibrationDefinition> {
+        self.measure_calibrations.iter().collect()
+    }
```
Shadow53, line 70:
```suggestion
    pub fn measure_calibrations(&self) -> Vec<&MeasureCalibrationDefinition> {
        self.iter_measure_calibrations().collect()
    }
```

Shadow53 (commented):
Left a couple small suggestions for consistency, but are fine to ignore.

I'm curious about the implementation of `CalibrationSet`. Maintaining insertion order is nice, but is there a specific use case beyond stable ordering?

I was imagining this would be implemented by sorting the calibrations in the `Vec`.

1. Searching for an item (`get`, `remove`) could be done with a binary search instead of linear (i.e. [`Vec::binary_search_by_key`](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.binary_search_by_key)
2. Equality could be derived, as the same calibrations would be in the same order, regardless of insertion order.
3. Insertion would be less performant due to sorting on insertion.

MarquessV:
> Left a couple small suggestions for consistency, but are fine to ignore.
> 
> I'm curious about the implementation of `CalibrationSet`. Maintaining insertion order is nice, but is there a specific use case beyond stable ordering?
> 
> I was imagining this would be implemented by sorting the calibrations in the `Vec`.
> 
>     1. Searching for an item (`get`, `remove`) could be done with a binary search instead of linear (i.e. [`Vec::binary_search_by_key`](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.binary_search_by_key)
> 
>     2. Equality could be derived, as the same calibrations would be in the same order, regardless of insertion order.
> 
>     3. Insertion would be less performant due to sorting on insertion.

Maintaining insertion has two intents:

1. User preference: Re-ordering the user's instructions seems to be a source of confusion. However, I think the minimum ask is that serialization deterministic.
2. My assumption is that insertion performance would be preferred to equality and serialization performance. Is that not true?

quil-rs/src/instruction/calibration.rs
```diff
@@ -9,6 +9,12 @@ use crate::{
 
 use super::write_qubit_parameters;
 
+pub trait CalibrationSignature {
+    type Signature: PartialEq;
+
+    fn signature(&self) -> Self::Signature;
+}
+
```
Shadow53, line 17: If you do something like this, we could return references and wouldn't need to clone in the implementations (not tested):

```suggestion
pub trait CalibrationSignature {
    type Signature<'a>: PartialEq;

    fn signature<'a>(&'a self) -> Self::Signature<'a>;
}
```

Shadow53:
> > Left a couple small suggestions for consistency, but are fine to ignore.
> > I'm curious about the implementation of `CalibrationSet`. Maintaining insertion order is nice, but is there a specific use case beyond stable ordering?
> > I was imagining this would be implemented by sorting the calibrations in the `Vec`.
> > ```
> > 1. Searching for an item (`get`, `remove`) could be done with a binary search instead of linear (i.e. [`Vec::binary_search_by_key`](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.binary_search_by_key)
> > 
> > 2. Equality could be derived, as the same calibrations would be in the same order, regardless of insertion order.
> > 
> > 3. Insertion would be less performant due to sorting on insertion.
> > ```
> 
> Maintaining insertion has two intents:
> 
> 1. User preference: Re-ordering the user's instructions seems to be a source of confusion. However, I think the minimum ask is that serialization deterministic.
> 2. My assumption is that parsing and insertion performance would be preferred to equality  performance. Is that not true?

I took a closer look at the implementation, because you are right that insertion performance makes sense to prefer over equality. I'm not sure it's actually currently faster than it would be when sorted, though.

Insertion uses the `replace` function, which is:

```rust
pub fn replace(&mut self, value: T) -> Option<T> {
    if let Some(index) = self.signature_position(value.signature()) {
        let replaced = std::mem::replace(&mut self.data[index], value);
        Some(replaced)
    } else {
        self.data.push(value);
        None
    }
}
```

And `signature_position` is:

```rust
fn signature_position(
    &self,
    signature: <T as CalibrationSignature>::Signature,
) -> Option<usize> {
    self.data
        .iter()
        .position(|element| element.signature() == signature)
}
```

So on every insert, the list is traversed linearly, giving us O(n) performance. `signature` can also clone lists (though somewhat cheaply due to the types involved), so there's a potential performance hit as well (that's a separate issue, which I'll comment on in the code).

If the list were sorted, `signature_position` becomes O(log n) due to binary search. If an entry exists, the performance remains O(log n) due to `mem::replace`. If it does not exist, we'd use `Vec::insert` to insert in the correct location, possibly shifting items, which brings us back to O(n).

Maybe I've analyzed this wrong? I agree that insertion performance is more important, so whatever we merge should be faster at insertion.

> Re-ordering the user's instructions seems to be a source of confusion

I can see that in the case of `HashSet`/`HashMap`, but that's a random re-ordering. I feel like users would be less confused if the reordering was in alphabetical order, or similar.

MarquessV:
> So on every insert, the list is traversed linearly, giving us O(n) performance...
> 
> If the list were sorted, `signature_position` becomes O(log n) due to binary search. If an entry exists, the performance remains O(log n) due to `mem::replace`. If it does not exist, we'd use `Vec::insert` to insert in the correct location, possibly shifting items, which brings us back to O(n).
> 
> Maybe I've analyzed this wrong? I agree that insertion performance is more important, so whatever we merge should be faster at insertion.

With the assumption that inserting a new calibration, rather than one with an existing signature is a much more common case (and is definitely the more useful case): Inserting a new calibration into the vector requires shifting any element in the vector that should be after it, which I think is where the performance would suffer the most with the sort approach. They both have the same time complexity, but appending to a Vec is significantly less work than shifting items in a Vec.

We're getting into the weeds with this, and these kinds of differences need benchmarks to really evaluate. I'd be open to testing that, but I don't want to slow down delivery of a useful feature if there isn't a clear reason to churn on the implementation details.

> > Re-ordering the user's instructions seems to be a source of confusion
> 
> I can see that in the case of `HashSet`/`HashMap`, but that's a random re-ordering. I feel like users would be less confused if the reordering was in alphabetical order, or similar.

I hear you, but that's not the impression I've gathered from users.

quil-rs/src/instruction/calibration.rs
MarquessV: Nice, I recall trying this without any luck. Will give it another shot.

MarquessV: Okay, I found the issue again. The provided snippet isn't valid, as Rust doesn't recognize the lifetime `'a` without defining it as a generic parameter to the trait. When you do provide it, it complains that the `'a` in `Signature` shadows the parameter.

I can get further with this definition:

```rs
pub trait CalibrationSignature<'a> {
    type Signature: PartialEq + 'a;

    fn signature(&'a self) -> Self::Signature;
}
```

But I can't figure out how to convince Rust that the lifetimes in the replace method is okay:

```rs
impl<'a, T> CalibrationSet<T>
where
    T: CalibrationSignature<'a>,
{
    // ...
    pub fn replace(&mut self, value: T) -> Option<T> {
        if let Some(index) = self.signature_position(value.signature()) {
            let replaced = std::mem::replace(&mut self.data[index], value);
            Some(replaced)
        } else {
            self.data.push(value);
            None
        }
    }
}
```

Errors with:

```sh
error[E0597]: `value` does not live long enough
   --> quil-rs/src/program/calibration_set.rs:109:54
    |
58  | impl<'a, T> CalibrationSet<T>
    |      -- lifetime `'a` defined here
...
105 |     pub fn replace<'b>(&'b mut self, value: T) -> Option<T>
    |                                      ----- binding `value` declared here
...
109 |         if let Some(index) = self.signature_position(value.signature()) {
    |                                                      ^^^^^------------
    |                                                      |
    |                                                      borrowed value does not live long enough
    |                                                      argument requires that `value` is borrowed for `'a`
...
116 |     }
    |     - `value` dropped here while still borrowed

error[E0505]: cannot move out of `value` because it is borrowed
   --> quil-rs/src/program/calibration_set.rs:110:69
    |
58  | impl<'a, T> CalibrationSet<T>
    |      -- lifetime `'a` defined here
...
105 |     pub fn replace<'b>(&'b mut self, value: T) -> Option<T>
    |                                      ----- binding `value` declared here
...
109 |         if let Some(index) = self.signature_position(value.signature()) {
    |                                                      -----------------
    |                                                      |
    |                                                      borrow of `value` occurs here
    |                                                      argument requires that `value` is borrowed for `'a`
110 |             let replaced = std::mem::replace(&mut self.data[index], value);
    |                                                                     ^^^^^ move out of `value` occurs here

error[E0505]: cannot move out of `value` because it is borrowed
   --> quil-rs/src/program/calibration_set.rs:113:28
    |
58  | impl<'a, T> CalibrationSet<T>
    |      -- lifetime `'a` defined here
...
105 |     pub fn replace<'b>(&'b mut self, value: T) -> Option<T>
    |                                      ----- binding `value` declared here
...
109 |         if let Some(index) = self.signature_position(value.signature()) {
    |                                                      -----------------
    |                                                      |
    |                                                      borrow of `value` occurs here
    |                                                      argument requires that `value` is borrowed for `'a`
...
113 |             self.data.push(value);
```

There are similar lifetime issues in signature_position(), and some of the trait implementations, but I suspect if we can fix this, we can fix the others.

Shadow53:
Fair enough. I'm happy to punt this discussion to a later time.

mhodson-rigetti:
> > > Re-ordering the user's instructions seems to be a source of confusion
> > 
> > I can see that in the case of `HashSet`/`HashMap`, but that's a random re-ordering. I feel like users would be less confused if the reordering was in alphabetical order, or similar.
> 
> I hear you, but that's not the impression I've gathered from users.

Enforced lexical order in the Quil-T components of the program would be contrary to the [Quil specification](https://quil-lang.github.io/#12-6Defining-Calibrations) in at least one case: "Multiple calibration definitions can be defined for different parameter and qubit values. When a gate is translated into control instructions the calibration definitions are enumerated in reverse order of definition and the first match will be taken." User insertion order must be preserved to obtain the intended translation result.

I think `quil-rs` takes some convenient interpretive decisions such as understanding all frame definitions are global and independent, and so places them up-front in the output in some (currently unspecified) order. This makes no difference to the translation result (at least as far as the specification is concerned) but also backs us in to a place where semantics based on the order cannot be proposed (practically) in the specification without also changing this implementation.

In Quil, I can write pretty much any instruction type in any order, and it is valid. Ideally, the `quil-rs` program would preserve this ordering. Non-deterministic ordering is the key issue here, because on any test failure of program comparison, the very next thing is to debug by comparing the textual Quil inputs and outputs. These cannot be reliably compared if output is non-deterministic. These can be compared if it is deterministic, but then the user must also pre-prepare the input using `quil-rs` to assure the same order- ie. Quil in textual form with a DEFFRAME at the end of the program will never be produced by this library.

A fully general solution would be to use a linked list for all instructions of all types (allowing a DEFFRAME at the end if that is what the user wrote, even though it has global scope in translation), and have the Program additionally index the instructions with (internally unordered) sets or maps keyed by the identity of e.g. DEFFRAME, DEFCAL, DEFGATE, etc to support O(log N) lookup. A linked list of instruction would be needed here as it permits a single O(log N) set update on insertion or deletion of a relevant type of instruction- a vector insert would need to reindex everything at O(N.log N) (which is worse than the insert itself).

At some scale of system you could imagine needing a syntactic ordering constraint- ie. that all objects referred to by an instruction are defined prior to that instruction, so the entire program is not needed first in memory and all indices built, but that's possibly for the future and would need tightening of the specification first.

kalzoo:
To follow up here, equality should _not_ be independent of ordering. Ordering of calibrations can change the way that they are applied within the program and thus are inherently not equal to one another.

Consider:

```
DEFCAL A(%x, pi):
   ...

DEFCAL A(pi, %x):
   ...

A(pi, pi) # expands into one or the other based on ordering
```

These two calibrations are of equivalent precedence and thus ordering makes the difference in which is applied to the gate. They should not be considered equal.

PR #355, opened 2024-04-15
feat!: Program instruction iteration and serialization is deterministic.

### LINKED ISSUE #353: Make program serialization deterministic
Due to the use of `HashMap`s in instruction fields, program serialization is not deterministic. As a user, this is confusing, because you can see different results between serialization attempts.

We can swap out `HashMap`s with `BTreeMap`s to make serialization deterministic while still being insensitive to ordering. One thing to consider is the hit to insertion performance `BTreeMap`s can theoretically have. This should be put to the test by benchmarks.

MarquessV:
Closes #353

### DIFF
--- a/quil-rs/src/instruction/frame.rs
+++ b/quil-rs/src/instruction/frame.rs
@@ -1,5 +1,6 @@
-use std::{collections::BTreeMap, str::FromStr};
+use std::str::FromStr;
 
+use indexmap::IndexMap;
 use nom_locate::LocatedSpan;
 
 use super::{MemoryReference, Qubit, QuotedString, WaveformInvocation};
@@ -30,7 +31,7 @@ impl Quil for AttributeValue {
     }
 }
 
-pub type FrameAttributes = BTreeMap<String, AttributeValue>;
+pub type FrameAttributes = IndexMap<String, AttributeValue>;
 
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct FrameDefinition {
--- a/quil-rs/src/instruction/frame.rs
+++ b/quil-rs/src/instruction/frame.rs
@@ -1,4 +1,4 @@
-use std::{collections::HashMap, str::FromStr};
+use std::{collections::BTreeMap, str::FromStr};
 
 use nom_locate::LocatedSpan;
 
@@ -30,12 +30,12 @@ impl Quil for AttributeValue {
     }
 }
 
-pub type FrameAttributes = HashMap<String, AttributeValue>;
+pub type FrameAttributes = BTreeMap<String, AttributeValue>;
 
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct FrameDefinition {
     pub identifier: FrameIdentifier,
-    pub attributes: HashMap<String, AttributeValue>,
+    pub attributes: FrameAttributes,
 }
 
 impl FrameDefinition {
--- a/quil-rs/src/instruction/waveform.rs
+++ b/quil-rs/src/instruction/waveform.rs
@@ -1,4 +1,4 @@
-use std::collections::HashMap;
+use std::collections::BTreeMap;
 
 use crate::{
     expression::Expression,
@@ -82,11 +82,11 @@ mod test_waveform_definition {
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct WaveformInvocation {
     pub name: String,
-    pub parameters: HashMap<String, Expression>,
+    pub parameters: BTreeMap<String, Expression>,
 }
 
 impl WaveformInvocation {
-    pub fn new(name: String, parameters: HashMap<String, Expression>) -> Self {
+    pub fn new(name: String, parameters: BTreeMap<String, Expression>) -> Self {
         Self { name, parameters }
     }
 }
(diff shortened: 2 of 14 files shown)


Shadow53 (approved):
Looks simple enough. My only question is one of performance. Has this been benchmarked, and how much does it affect performance? If it's worse, have you also checked against e.g. `IndexMap`?

kalzoo:
Using `index_map` would also help solve the problem of duplicate calibrations.

quil-rs/src/instruction/frame.rs
```diff
@@ -30,12 +30,12 @@ impl Quil for AttributeValue {
...
 
-pub type FrameAttributes = HashMap<String, AttributeValue>;
+pub type FrameAttributes = BTreeMap<String, AttributeValue>;
 
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct FrameDefinition {
     pub identifier: FrameIdentifier,
-    pub attributes: HashMap<String, AttributeValue>,
+    pub attributes: FrameAttributes,
 }
```
kalzoo, line 39: This is a breaking change.

quil-rs/src/instruction/waveform.rs
```diff
@@ -82,11 +82,11 @@ mod test_waveform_definition {
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct WaveformInvocation {
     pub name: String,
-    pub parameters: HashMap<String, Expression>,
+    pub parameters: BTreeMap<String, Expression>,
```
kalzoo, line 85: Likewise breaking.

quil-rs/src/program/mod.rs
```diff
@@ -1418,4 +1418,83 @@ CNOT 2 3";
...
+DEFFRAME 0 \"xy\":
+\tDIRECTION: \"rx\"
+\tCENTER-FREQUENCY: 1000
+\tHARDWARE-OBJECT: \"some object\"
+\tINITIAL-FREQUENCY: 2000
+\tSAMPLE-RATE: 3000";
+        let program = Program::from_str(input).unwrap();
+        let quil = program.to_quil().unwrap();
+
+        // Asserts that serialization doesn't change on reperated attempts.
```
kalzoo, line 1492:
```suggestion
        // Asserts that serialization doesn't change on repeated attempts.
```

kalzoo (changes requested):
if breaking, the commit should indicate as much.

as long as you're breaking, should move to index map where it's needed so as to break only once.

edit to clarify: BTreeMap makes sense where you've added it here; IndexMap is useful for Calibrations where ordering has semantic meaning. Both changes can be made at once though to reduce 2 breaking changes into 1

quil-rs/src/instruction/frame.rs
MarquessV: Added the necessary ! to the PR title to indicate a breaking change. Will make sure it's there when squashed to.

MarquessV:
Opened rigetti/rigetti-pyo3#43 to add support for IndexMap.

MarquessV:
@Shadow53 @kalzoo - Is there a good reason not to replace all instances of `BTreeMap`/`HashMap` in the public API with `IndexMap`? Theoretically, if offers better insert performance, and it arguably provides better UX since users can expect instructions to be ordered the same way they inserted them (with respect to how `Program` groups them, of course).

quil-rs/src/program/mod.rs
```diff
@@ -1427,4 +1427,83 @@ CNOT 2 3";
...
+        let program = Program::from_str(input).unwrap();
+        let quil = program.to_quil().unwrap();
+
+        // Asserts that serialization doesn't change on repeated attempts.
+        // 100 is chosen because it should be more than sufficient to reveal an
+        //     issue and it has a negligible impact on execution speed on the test suite.
+        let iterations = 100;
+        for _ in 0..iterations {
+            assert_eq!(program.to_quil().unwrap(), quil);
+        }
```
Shadow53, line 1507: Would it also be worth testing this (i.e. that programs parsed from the same quil serialize to the same quil)?

```
let left = input.parse::<Program>().unwrap().to_quil().unwrap();
let right = input.parse::<Program>().unwrap().to_quil().unwrap();
assert_eq!(left, right);
```

Shadow53 (approved):
Looks good! One suggestion is for another test, if it makes sense to add.

kalzoo:
> @Shadow53 @kalzoo - Is there a good reason not to replace all instances of `BTreeMap`/`HashMap` in the public API with `IndexMap`? Theoretically, if offers better insert performance, and it arguably provides better UX since users can expect instructions to be ordered the same way they inserted them (with respect to how `Program` groups them, of course).
> 
> EDIT: Updated this PR w/ those changes.

Our criteria should be:

1. equality is correct (e.g. ordering does not matter for frames, does matter for defcals)
2. duplication behavior is correct (i.e. all of them act as a set and duplicates are not stored)
3. best performance as used; I'd hazard a guess that iteration speed is the most important, followed by insertion and then cloning.

So if indexmap checks those boxes, I'm good with it.

quil-rs/src/program/mod.rs
MarquessV: I wasn't sure, so I went ahead and updated the test so that it parses a new program and asserts the quil is still the same.

PR #359, opened 2024-04-17
feat: Support getting and constructing BasicBlocks.

### LINKED ISSUE #358: Allow control flow graph and related types to be constructed and subclassed
It would be helpful if the control flow graph types could be subclassed and had some way of constructing them. This allows downstream dependencies to more easily extend them or modify their behavior.

MarquessV:
closes #358 

supports rigetti/pyquil#1770

### DIFF
--- a/quil-py/quil/program/__init__.pyi
+++ b/quil-py/quil/program/__init__.pyi
@@ -52,24 +52,20 @@ class Program:
     @property
     def declarations(self) -> Dict[str, Declaration]: ...
     def dagger(self) -> "Program":
-        """
-        Creates a new conjugate transpose of the ``Program`` by reversing the order of gate
-        instructions and applying the DAGGER modifier to each.
+        """Creates a new conjugate transpose of the ``Program`` by reversing the order of gate instructions and applying the DAGGER modifier to each.
 
         Raises a ``GateError`` if any of the instructions in the program are not a ``Gate`
         """
         ...
     def expand_calibrations(self) -> "Program":
-        """
-        Expand any instructions in the program which have a matching calibration, leaving the others
-        unchanged. Recurses though each instruction while ensuring there is no cycle in the expansion
-        graph (i.e. no calibration expands directly or indirectly into itself)
+        """Expand any instructions in the program which have a matching calibration, leaving the others unchanged.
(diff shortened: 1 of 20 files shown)


quil-rs/src/program/analysis/control_flow_graph.rs
```diff
@@ -269,6 +269,37 @@ pub struct BasicBlockOwned {
...
+    }
+
+    pub fn instruction_index_offset(&self) -> usize {
+        self.instruction_index_offset
+    }
+
+    pub fn terminator(&self) -> &BasicBlockTerminatorOwned {
+        &self.terminator
+    }
+}
```
kalzoo, line 301: Are these needed? the `*Owned` structs here are explicitly just as a holding pen for the inner data because the python wrappers can't just hold a ref. AFAIK they should never be directly constructed or queried but only be converted to/from `BasicBlock`

quil-py/quil/program/__init__.pyi
```diff
@@ -330,7 +329,9 @@ class CalibrationSet:
         function returns the previous calibration. Otherwise, None is returned.
         """
         ...
-    def insert_measurement_calibration(self, calibration: MeasureCalibrationDefinition) -> Optional[MeasureCalibrationDefinition]:
+    def insert_measurement_calibration(
+        self, calibration: MeasureCalibrationDefinition
+    ) -> Optional[MeasureCalibrationDefinition]:
```
kalzoo, line 334: Would it be quick to cover these stubs with a CI lint so no one merges misformatted code like I did here?

quil-rs/src/program/analysis/control_flow_graph.rs
MarquessV: Nah, I thought I had those changes discarded. Removed.

quil-py/quil/program/__init__.pyi
MarquessV: I've been wanting to move to `ruff` for all our projects, so I went ahead and did that, addressed the new issues it flagged, and added lints and formatting to the default checks.

kalzoo (approved):

Issue #367, opened 2024-04-22
proptest_assert_eq may consider NaN == NaN to be false


erichulburd:
See the results of this job in MSRV task: https://github.com/rigetti/quil-rs/actions/runs/8790199931/job/24121695036

```
Test failed: assertion failed: `(left == right)` 
  left: `Number(Complex { re: NaN, im: 0.0 })`, 
 right: `Number(Complex { re: NaN, im: 0.0 })`: Simplified expressions should be equal:
parenthesized aa[0]/sin(pi) ("aa[0]/sin(pi)") extracted from (aa[0]/(sin(pi))) simplified to NaN
vs original aa[0]/sin(pi) ("aa[0]/sin(pi)") simplified to NaN at quil-rs/src/expression/mod.rs:993.
minimal failing input: e = Infix(
    InfixExpression {
        left: Address(
            MemoryReference {
                name: "aa",
                index: 0,
            },
        ),
        operator: Slash,
        right: FunctionCall(
            FunctionCallExpression {
                function: Sine,
                expression: PiConstant,
            },
        ),
    },
)
```

I've added this test case locally to `specific_round_trip_tests`, but couldn't reproduce the failure. That being the case, I suppose we can wrap that particular test case to check for the NaN case.

PR #368, opened 2024-04-22
test: handle nan in proptest cases

erichulburd:
Fixes https://github.com/rigetti/quil-rs/issues/367

### DIFF
--- a/quil-rs/src/expression/mod.rs
+++ b/quil-rs/src/expression/mod.rs
@@ -810,6 +810,13 @@ mod tests {
         ((-tau..tau), (-tau..tau)).prop_map(|(re, im)| Complex64 { re, im })
     }
 
+    /// Filter an Expression to not be constantly zero.
+    fn nonzero(strat: impl Strategy<Value = Expression>) -> impl Strategy<Value = Expression> {
+        strat.prop_filter("Exclude constantly-zero expressions", |expr| {
+            expr.clone().into_simplified() != Expression::Number(Complex64::new(0.0, 0.0))
+        })
+    }
+
     /// Generate an arbitrary Expression for a property test.
     /// See https://docs.rs/proptest/1.0.0/proptest/prelude/trait.Strategy.html#method.prop_recursive
     fn arb_expr() -> impl Strategy<Value = Expression> {
@@ -820,11 +827,12 @@ mod tests {
             Just(PiConstant),
             arb_name().prop_map(Variable),
         ];
-        (leaf).prop_recursive(
+        leaf.prop_recursive(
             4,  // No more than 4 branch levels deep
             64, // Target around 64 total nodes
             16, // Each "collection" is up to 16 elements
             |expr| {
+                let inner = expr.clone();
                 prop_oneof![
                     (any::<ExpressionFunction>(), expr.clone()).prop_map(|(function, e)| {
                         Expression::FunctionCall(FunctionCallExpression {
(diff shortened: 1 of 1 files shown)


quil-rs/src/expression/mod.rs
```diff
@@ -848,6 +848,14 @@ mod tests {
         )
     }
 
+    macro_rules! prop_assert_eq_or_nan {
+        ($left:expr, $right:expr) => {
+            if !is_nan($left) && !is_nan($right) {
+                prop_assert_eq!($left, $right);
+            }
+        };
```
Shadow53, line 856: This does nothing in the case where only one is `NaN`, which is incorrect.

```suggestion
        ($left:expr, $right:expr) => {
            if !is_nan($left) || !is_nan($right) {
                prop_assert_eq!($left, $right);
            }
        };
```

quil-rs/src/expression/mod.rs
```diff
@@ -990,27 +1003,38 @@ mod tests {
...
+            }
         }
 
     }
 
+    fn is_nan(expression: &Expression) -> bool {
+        match expression {
+            Expression::Number(value) => value.is_nan(),
+            _ => false,
+        }
```
Shadow53, line 1027: What about the case where an `Expression` has `NaN` nested somewhere inside it?

Shadow53 (commented):
I don't think this properly resolves the issue. I left a couple comments indicating what I think are logic errors.

I think the proper solution would be one of:

1. Rewrite `arb_expr` to disallow `NaN` and avoid these test cases from being generated
2. Replace our usage of `f64` with `ordered_float::OrderedFloat`, which impls `Ord` by violating the floating-point spec and saying `Nan == NaN`.

quil-rs/src/expression/mod.rs
erichulburd: If only one is NaN, my assumption is `prop_assert_eq!($left, $right);` would return false, which is what we want, right?

quil-rs/src/expression/mod.rs
erichulburd: Uff, good point. Seems like we should re-write `impl PartialEq for Expression`. Specifically,

```rs
(Self::Number(left), Self::Number(right)) => left == right
```

should explicitly check for NaN?

Shadow53: That seems fine to me

quil-rs/src/expression/mod.rs
```diff
@@ -125,6 +127,10 @@ impl PartialEq for Expression {
     }
 }
 
+fn are_equal_or_nan(left: f64, right: f64) -> bool {
+    (left.is_nan() && right.is_nan()) || (left == right)
```
MarquessV, line 131: In principle, I would prefer that `NaN != NaN` (or rather `NaN != x` for any expression `x`. That aligns with the IEEE specification of `NaN` for floats that most languages implement. As a Quil user, I would be surprised to see that `NaN` evaluated as equivalent to anything else.

Is there a practical reason I'm missing as to why we would want this?

MarquessV (changes requested):

quil-rs/src/expression/mod.rs
erichulburd: The only reason I made this change was because proptest generated test input that yielded NaN; the test simplified the expression to NaN, and there was an assertion that the two should be equal.

I guess a better tact here, one that would not impact non-test code, would be to simply filter out proptest input that generates NaN.

quil-rs/src/expression/mod.rs
erichulburd: I just filtered `/ 0` out of any prop test generated expressions. This won't affect any non-test code.

antalsz:
This looks good to me – seems like we might want to have a "nonzero" generator since we use the same code in a few places, but I'd merge this.  (Full disclosure: I just ran into this bug.)

kalzoo (approved):

PR #369, opened 2024-04-24
feat: add waveform templates

kalzoo:
Adds waveform templates described by the [Quil spec](https://quil-lang.github.io/). (Section 12.2).

This moves out internal/closed-source code into the open source arena to help users better understand how their programs are executed.

TODO:

- [x] Tests (there were tests relying on private infrastructure that could not be copied out to GItHub)
- [x] Stylistically, should we make the `-> Vec<Complex64>` methods on their parameter structs rather than free functions?


quil-rs/src/waveform/templates.rs
```diff
@@ -0,0 +1,264 @@
...
+}
+
+/// A custom `ceil()` method that includes a machine-epsilon sized region above each integer in the
+/// set of values that are mapped to that integer. In other words:
+/// ceil_eps(x) = n, for all x and integers n s.t. n <= x < (n + n*epsilon)
+///
+/// To handle accumulated floating point errors in sweeps above typical floating point imprecision
+/// we make epsilon 10x larger than floating point epsilon.
+pub(super) fn ceiling_with_epsilon(value: f64) -> f64 {
+    let truncated = value - (value * 10.0 * std::f64::EPSILON);
```
MarquessV, line 33: Should the adjusted value be the value _plus_ 10*epsilon?

quil-rs/src/waveform/templates.rs
```diff
@@ -0,0 +1,264 @@
...
+    /// Scale to apply to waveform envelope
+    pub scale: f64,
+    /// Phase shift for entire waveform
+    pub phase: f64,
+    /// Explicit detuning to bake into iq values
+    pub detuning: f64,
+}
+
+/// Creates a waveform with a flat top and edges that are error functions (erfs).
+pub fn erf_square(parameters: ErfSquareParameters) -> Vec<Complex64> {
```
MarquessV, line 79: I do wonder if these should be methods on the struct, rather than top level methods. Would be less import to worry about for users.

MarquessV (commented):
A few comments, but looking good!

quil-rs/src/waveform/templates.rs
kalzoo: yep, that was what I meant in the MR description. Moved them and added a unifying trait

quil-rs/src/waveform/templates.rs
kalzoo: I don't think so.

kalzoo:
Added visual snapshots to help understand template behavior: 
image

MarquessV (approved):
Love the visual snapshots!

PR #370, opened 2024-05-04
feat: add source mapping for calibration expansion

### LINKED ISSUE #366: Expand calibrations with source mapping
Currently, calibration expansion happens somewhat opaquely: a `Vec<Instruction>` goes in, and another `Vec<Instruction>` comes out. In complex programs, it can be hard for a user to discern which post-calibration instructions originated from which pre-calibration instructions.

This can be fixed with a source map. Perhaps it can be structured like this:

```
struct SourceMap<SourceIndex, TargetIndex> {
    entries: Vec<SourceMapEntry<SourceIndex, TargetIndex>>
}

struct SourceMapEntry<Label, SourceIndex, TargetIndex> {
    /// A description of the mapping, such the calibration used to perform the expansion
    label: Label,

    source_location: Vec<SourceIndex>,
    target_location: Vec<TargetIndex>,
}
```

these are generic over `*Index` in order to be re-used for other mappings, such as in parsing a program (mapping `nom::locate` information such as line and column span). `Label`, likewise, would be a different type in these situations:

* For calibration expansion, may be the calibration used to perform the expansion. Users like @bramathon have often wondered about which DEFCAL was actually being used, and this could help make that explicit. We'd have to figure out how this works for nested expansions.
* For parsing, might just be `()`, I don't see a need for a label

This type would be made available for other libraries to use as well, such as compilers which consume Quil.

Finally, `SourceMap`s should be able to be folded, such that a program can undergo multiple transformations in series and emit a single mapping from the original text down to the final artifact.

kalzoo:
This PR allows the (Rust package) user to do the following:

* Expand a single instruction in a self-descriptive, recursive way. That is, expand an instruction using its calibrations, and alongside the newly expanded instructions, receive an object describing the calibration decisions made and which of the new instructions are a result of which calibration decisions.
* Expand an entire program in the same way, receiving not only the expanded program but a source mapping explaining how every instruction in the new program was derived from the first one: either it was expanded (and it will show the expanded instruction as well as the DEFCAL [MEASURE]s used to expand it) or it was not (and it will yield the instruction index in both source and target program which contains it)

Reviewer:

* Please consider the two comments suggesting a breaking change to the `MeasureCalibrationDefinition` and `Calibration` structs. This would make them somewhat more similar to DEFFRAME in that they would have identifiers that could be used to index them without respect to their instruction bodies. We could also take the chance to rename `Calibration` to `CalibrationDefinition` for consistency. 
  * **I have made this change, but in a way that should not break some python consumers.**
  * All of the original `@property`s are there, so this will only break users who construct and modify calibrations and measure calibrations. I could add additional setters to help migrate, but there's only one `__new__` and it doesn't feel right to leave those behind.
* Please take a look at the new tests and consider the SourceMap format. Do you see any shortcomings?
* Please look at the `CalibrationSet#_expand` method and those nearby.
  * I'd like to have this only construct a `CalibrationExpansion` when that is returned to the user; I first tried an `expansion: Option<&mut CalibrationExpansion>` argument, but that doesn't work well here because that doesn't have a sensible default.
  * So instead I went with `build_source_map` which skips _most_ of the work involved in calibration expansion source mapping while still technically constructing one which is then discarded prior to return from `CalibrationSet.expand`.
  * Can you suggest a more elegant solution? I want to share logic between expand-with-source-map and expand-without-source-map while not doing any source mapping work if one isn't going to be returned.

TODOs:

- [x] Fix the bug in program calibration expansion source mapping
- [x] Restore the "fast path" to instruction calibration expansion when no source map is desired. In the current implementation, it's built in all cases and then thrown away in some in order to maintain the existing API.
- [x] (maybe, at request) An ergonomic reverse-mapping to allow looking up the source instruction that resulted in a given target instruction. That can always be built by the user, but it might be nice to have one here since the source mapping structure takes a moment to grok even for an experienced Quil dev.
- [ ] (maybe) implement source mapping for parsing. This could be done on an experimental branch just to validate that this PR has the right API to support that.

Closes #366


mhodson-rigetti:
As much as I can read Rust, this is making sense to me. On the first point about a lurking bug, I wonder if `impl SourceMapRange for CalibrationExpansion` `contains()` is incomplete as containment is only by the index being valid, not that it also came from the same calibration? On the third point about reverse-mapping help, if I could see the pattern for this implemented using only the Python bindings as a recipe I think I'd be fine, but if doing that means you might as well codify on the Rust side, happy for that also!

kalzoo:
> As much as I can read Rust, this is making sense to me. On the first point about a lurking bug, I wonder if `impl SourceMapRange for CalibrationExpansion` `contains()` is incomplete as containment is only by the index being valid, not that it also came from the same calibration? On the third point about reverse-mapping help, if I could see the pattern for this implemented using only the Python bindings as a recipe I think I'd be fine, but if doing that means you might as well codify on the Rust side, happy for that also!

Hey, thanks @mhodson-rigetti . I've added python bindings and docs but no example (yet) - I'll have to think about how best to expose those direct mappings from source to target. Right now (following our chat a week ago) I made the structure recursive to model the recursive nature of calibration expansion, and that's the least intuitive part in practical use, because each level of expansion has its own index basis. There's an example describing that problem in the python docstrings now.

You were right about `SourceMapRange` for exactly that reason - two ranges may not share the same basis. I just removed `SourceMapRange` altogether in this iteration, I don't think this functionality needs to be so general-purpose just yet.

The lingering bug though was actually fixed in [this commit](https://github.com/rigetti/quil-rs/pull/370/commits/bf246b3cce0e9ea5da107638461688d7a756b641) and was something prompted by our chat a week ago - certain instructions being yoinked out of the instruction body and thus (necessarily) out the source map.

quil-rs/src/instruction/calibration.rs
```diff
@@ -20,46 +20,42 @@ pub trait CalibrationSignature {
...
             instructions,
-            modifiers,
-            name: name.to_string(),
-            parameters,
-            qubits,
         })
     }
 }
 
+impl CalibrationSignature for Calibration {
```
kalzoo, line 40: From @MarquessV : this should not be needed anymore if the `CalibrationIdentifier` can serve as the signature

quil-rs/src/instruction/calibration.rs
```diff
@@ -20,46 +20,42 @@ pub trait CalibrationSignature {
 
 #[derive(Clone, Debug, Default, PartialEq)]
 pub struct Calibration {
```
kalzoo, line 22: @MarquessV is open to renaming and breaking this, as long as this PR is already breaking

quil-rs/src/program/mod.rs
```diff
@@ -261,11 +276,76 @@ impl Program {
...
+                let end_length = self.instructions.len();
+
+                // If the instruction was not added to the program body, remove its target index from the source map
+                // so that the map stays correct.
+                if start_length == end_length {
+                    let relative_target_index =
+                        start_length - previous_program_instruction_body_length;
+                    expansion_output
+                        .detail
+                        .remove_target_index(relative_target_index);
```
kalzoo, line 331: This feels very clumsy and likely inefficient, but no more clever way occurred to me. Suggestions welcome.

Basically, we determine whether an instruction is a "header" or "body" instruction by whether or not it made the instruction body longer, and then based on that we excise it from the source mapping so that the `target_index` ranges stay accurate and don't map to a missing instruction

quil-py/src/program/source_map.rs
```diff
@@ -0,0 +1,147 @@
...
+
+impl_repr!(PyCalibrationExpansion);
+
+#[pymethods]
+impl PyCalibrationExpansion {
+    pub fn calibration_used(&self) -> PyCalibrationSource {
+        self.as_inner().calibration_used().into()
+    }
+
+    // Reviewer: is there a better return type?
```
MarquessV, line 29: I'm not against this implementation, but an option is to return a Python range. That might be slightly more Pythonic:

```rs
pub fn range(&self) -> (usize, usize) {
    let range = py.import("builtins")?.get("range")?;
    let args = PyTuple::new(py, &[self.as_inner().range().start, self.as_inner().range().end]);
    range.call1(args)
}
```
Return type aside, I wonder if it's possible to only consume the inner range a single time, rather that once for the start and once for the end.

quil-py/src/program/source_map.rs
```diff
@@ -0,0 +1,147 @@
...
+        (self.as_inner().range().start, self.as_inner().range().end)
+    }
+
+    pub fn expansions(&self) -> PyCalibrationExpansionSourceMap {
+        self.as_inner().expansions().into()
+    }
+}
+
+py_wrap_type! {
+    #[derive(Debug, PartialEq)]
```
MarquessV, line 40: Any Python type that can derive PartialEq should also use the `impl_eq!` macro so that Python can leverage it as well.

quil-rs/src/program/calibration.rs
```diff
@@ -298,22 +502,23 @@ impl Calibrations {
...
         for calibration in self.iter_calibrations() {
             // Filter out non-matching calibrations: check rules 1-4
-            if calibration.name != gate.name
-                || calibration.modifiers != gate.modifiers
-                || calibration.parameters.len() != gate.parameters.len()
-                || calibration.qubits.len() != gate.qubits.len()
+            if calibration.identifier.name != gate.name
+                || calibration.identifier.modifiers != gate.modifiers
+                || calibration.identifier.parameters.len() != gate.parameters.len()
+                || calibration.identifier.qubits.len() != gate.qubits.len()
```
MarquessV, line 508: Would it make sense to throw this in an `is_match(&self, gate: &Gate)` method on CalibrationIdentifier?

quil-rs/src/program/mod.rs
MarquessV: I don't necessarily have a better idea for this, but an observation that the distinction between "header" and "body" instruction comes up enough that maybe we should have a more concrete definition of it in the library? Like a method in `Instruction` maybe? I know it's not a part of the spec, but it is a practical one that this library makes decisions around.

antalsz (changes requested):
Looks overall good to me, but I think some details need to be hashed out before we merge this

quil-rs/src/program/calibration.rs
kalzoo: yep, good thought, done

quil-py/src/program/source_map.rs
kalzoo: sure - TIL this trick with pyo3. It's compiling but I don't think I did it right; will fix it up and test on the next pass.

quil-py/src/program/source_map.rs
kalzoo: good call - added some, will sweep for rest on next pass

quil-rs/src/program/mod.rs
kalzoo: > if the calibration expansion can check if something is a "header" instruction, then it can refrain from adding it to the expanded source, and then we don't need to do this fixup.

I don't see how this is possible. When the instruction is expanded, it does not have the scope of a program, and so it is just expanded into a `Vec<Instruction>`. However, when we add that Vec to a program here, then some of its members are "hoisted" into data structures like memory region declarations and calibration declarations and such. Expanding the calibration in such a way that "predicted" how the program would handle it, without the use of an actual program, is something that could fall out of sync and requires extra care and maintenance.

quil-rs/src/program/source_map.rs
```diff
@@ -0,0 +1,47 @@
...
+/// mapped in either direction.
+///
+/// The behavior of such mappings depends on the implementations of the generic `Index` types,
+/// but this may be a many-to-many mapping, where one element of the source is mapped (contributed)
+/// to zero or many elements of the target, and vice versa.
+///
+/// This is also intended to be mergeable in a chain, such that the combined result of a series
+/// of transformations can be expressed within a single source mapping.
+#[derive(Clone, Debug, PartialEq)]
+pub struct SourceMap<SourceIndex, TargetIndex> {
```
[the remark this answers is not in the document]
kalzoo: yeah, I've had that in mind from the start, but the first attempt was clumsy given the generics involved. I've added one now, though - good interface, although perhaps less efficient than it could be.

> (maybe, at request) An ergonomic reverse-mapping to allow looking up the source instruction that resulted in a given target instruction. That can always be built by the user, but it might be nice to have one here since the source mapping structure takes a moment to grok even for an experienced Quil dev.

quil-py/src/program/source_map.rs
kalzoo: got this fixed up and under test - looking good. please lmk what you think

quil-rs/src/program/calibration.rs
```diff
@@ -57,6 +61,118 @@ impl<'a> MatchedCalibration<'a> {
...
+        self.expansions.entries.retain_mut(
+            |entry: &mut SourceMapEntry<usize, CalibrationExpansion>| {
+                if let Some(target_with_offset) = target_index.checked_sub(target_index_offset) {
+                    eprintln!("Target with offset: {}", target_with_offset);
+                    entry
+                        .target_location
+                        .remove_target_index(target_with_offset);
+                }
+
+                !entry.target_location.range.is_empty()
```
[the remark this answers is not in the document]
kalzoo: how would we maintain that as an invariant, if one expands to something that is all headers?

kalzoo (commented):
Thanks for the review, @antalsz

quil-rs/src/instruction/calibration.rs
MarquessV: That is true, I don't know of a good way to make the generics work with just a `CalibrationIdentifier` struct. I agree that it makes sense to formalize the concept of a calibration identifier w/ a struct and defer to it to implement the signature logic. They serve different purposes, so it's not entirely redundant, and the core logic doesn't need to be duplicated between either, so I'm good with this.

quil-py/src/program/source_map.rs
antalsz: Looks good to me!

quil-rs/src/instruction/calibration.rs
```diff
@@ -153,7 +153,7 @@ impl CalibrationIdentifier {
                         (calib, gate) => calib == gate,
                     }
                 });
-        !fixed_parameters_match
+        fixed_parameters_match
```
antalsz, line 156: This completely inverts the meaning of the function but nothing else seems to have changed?

quil-rs/src/program/calibration.rs
```diff
@@ -87,34 +87,33 @@ pub struct CalibrationExpansion {
...
     /// Remove the given target index from all entries, recursively.
     ///
-    /// This is to be used when the named index is removed from the target program
-    /// in the process of calibration expansion.
+    /// This is to be used when the given index is removed from the target program
+    /// in the process of calibration expansion (for example, a `DECLARE`).
     pub(crate) fn remove_target_index(&mut self, target_index: usize) {
-        eprintln!("Removing target index {} from\n{self:#?}", target_index);
-
+        // Adjust the start of the range if the target index is within the range
```
antalsz, line 93:
```suggestion
        // Adjust the start of the range if the target index is before the range
```

quil-rs/src/instruction/calibration.rs
```diff
@@ -69,7 +65,45 @@ impl Quil for Calibration {
...
+#[derive(Clone, Debug, Default, PartialEq)]
+pub struct CalibrationIdentifier {
+    /// The modifiers applied to the gate
+    pub modifiers: Vec<GateModifier>,
+
+    /// The name of the gate
+    pub name: String,
+
+    /// The parameters of the gate - these are the variables in the calibration definition
+    pub parameters: Vec<Expression>,
```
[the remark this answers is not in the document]
antalsz: Ahh, that makes sense – I was preemptively normalizing.

quil-rs/src/program/calibration.rs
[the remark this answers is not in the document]
antalsz: Since I did this review I have come to understand that headers move 👍

quil-rs/src/program/calibration.rs
```diff
@@ -219,25 +371,77 @@ impl Calibrations {
...
+                                .new_instructions
+                                .push(instruction);
                         }
-                        None => recursively_expanded_instructions.push(instruction),
                     };
                 }
+
+                if build_source_map {
+                    // While this appears to be duplicated information at this point, it's useful when multiple
+                    // source mappings are merged together.
```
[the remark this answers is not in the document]
antalsz: I have read the code again and figured out what I was missing, which is that sometimes the ranges aren't created at index `0`.  Sorry for the miss

quil-rs/src/program/mod.rs
antalsz: It's statically knowable which instructions get hoisted and which ones don't: `Instruction::CalibrationDefinition` does, `Instruction::Gate` doesn't, etc.  Expansion could return *two* vectors: `Vec<HoistedInstruction>` and `Vec<ComputationInstruction>` (new types optional, but perhaps clearer).  Then anything working on the results of expansion wouldn't have to do its own classification.

antalsz (commented):
These changes look good, thanks!  I have a few open threads that weren't addressed, but none I'd consider blockers.  However, the discussion around header/body instruction distinction should wrap up before we merge this.

quil-rs/src/instruction/calibration.rs
kalzoo: Yeah, it was just fixing a typo made within this MR when I moved this from its [previous home](https://github.com/rigetti/quil-rs/pull/370/files#diff-1816364e525c509ac2bb6d523144a0f2c0fc33697abf580e045bbb6e25afa74bL354)

antalsz: Whoops, I should've caught that during my last review, sorry!

antalsz (approved):
As I said in my last review – now that I understand the negation change, I would not consider any of my remaining comments blocking.  Approved!

quil-py/quil/program/__init__.pyi
```diff
@@ -1,3 +1,70 @@
...
+        DEFCAL X 0:
+            Y 0
+
+        DEFCAL Y 0:
+            Z 0
+
+        Z 0 # This instruction is index 0
+        Z 0 # This instruction is index 1
+        \"\"\"
+    )
```
antalsz, line 41: I think this ought to have an assert.  That raises questions of the precise whitespace, I know.  Could we parse them both to Quil and assert the parsed programs are equal?

quil-py/quil/program/__init__.pyi
```diff
@@ -1,3 +1,70 @@
...
+
+    # In order to discover _which_ calibration led to each Z in the resulting program, we
+    # can interrogate the expansion source mapping:
```
antalsz, line 44: I like this section a lot, but I found the specific comments a bit confusing.  Perhaps we could revise this first comment to something like

"…can interrogate the expansion source mapping.  For instance, The X at index 0 should have been replaced with a Z at index 0.  Here's how we could confirm that: First, we…".

As it stands, the comment on line 46 made me think that we were also going to check the Y at index 1.  (I think adding that would probably be redundant.)

Another possibility would be to drop the ellipses on lines 51 and 63; they're not following on from a previous sentence.

quil-py/quil/program/__init__.pyi
```diff
@@ -1,3 +1,70 @@
+"""
+The `quil.program` module contains classes for constructing and representing a Quil program.
+
+# Examples
+
+## Source Mapping for Calibration Expansion
+
+```py
+    import inspect
```
antalsz, line 9: Markdown code blocks with ` ``` ` don't need to have their contents indented, so we could dedent what follows.

quil-py/quil/program/__init__.pyi
```diff
@@ -1,3 +1,70 @@
+"""
+The `quil.program` module contains classes for constructing and representing a Quil program.
+
+# Examples
+
+## Source Mapping for Calibration Expansion
+
+```py
```
antalsz, line 8: Is this snippet run and tested?  It would be nice if it were but I don't know enough about Python to tell if that's happening/how easy it would be to have that happen.

antalsz (commented):
Reviewed 5d31908ca31f86e87c0baf1e45e57b21b2c0e822; a couple minor thoughts about the explanation, but the overall concept is great.

quil-py/quil/program/__init__.pyi
kalzoo: yes, it's a test case

quil-py/quil/program/__init__.pyi
kalzoo: ah, of course. will fix

kalzoo: fixed

quil-py/quil/program/__init__.pyi
kalzoo: the test did, I left it out here for readability but have restored it now

quil-py/quil/program/__init__.pyi
antalsz: Can we add a pointer to the test case it corresponds to?

quil-py/quil/program/__init__.pyi
kalzoo: yep, fixed

bramathon:
Just wanted to check on the status of this - it's important for pulse plotting to work.

bramathon:
I'm a little surprised by the extent of the changes in this PR. I was expecting something more along the lines of this:

```python
from quil.program import Program
from typing import Tuple, List

def expand_with_source_mapping(program: Program) -> Tuple[Program, List[int]]:
    """
    Expand any instructions in the program which have a matching calibration, leaving the others
    unchanged. 

    :param program: A quil.Program.
    :return: A quil.Program with the instructions expanded and a source map.
    The source map is a list with the length of the expanded instructions, and indicates the index
    of the logical instruction in the original program from which the expanded instruction originated.
    """
    instructions = program.body_instructions
    calibrations = program.calibrations
    expanded_program = program.clone_without_body_instructions()

    source_map = []
    expanded_instructions = []
    for logical_index, inst in enumerate(instructions):
        expanded_instruction = calibrations.expand(inst, [])
        source_map += [logical_index]*len(expanded_instruction)
        expanded_instructions += expanded_instruction
    expanded_program.add_instructions(expanded_instructions)
    return expanded_program, source_map

expanded_program, source_map = expand_with_source_mapping(quil_program)
default_expanded_program = quil_program.expand_calibrations()
assert expanded_program == default_expanded_program
```

kalzoo:
> I'm a little surprised by the extent of the changes in this PR. I was expecting something more along the lines of this:
> 
> ```python
> from quil.program import Program
> from typing import Tuple, List
> 
> def expand_with_source_mapping(program: Program) -> Tuple[Program, List[int]]:
>     """
>     Expand any instructions in the program which have a matching calibration, leaving the others
>     unchanged. 
> 
>     :param program: A quil.Program.
>     :return: A quil.Program with the instructions expanded and a source map.
>     The source map is a list with the length of the expanded instructions, and indicates the index
>     of the logical instruction in the original program from which the expanded instruction originated.
>     """
>     instructions = program.body_instructions
>     calibrations = program.calibrations
>     expanded_program = program.clone_without_body_instructions()
> 
>     source_map = []
>     expanded_instructions = []
>     for logical_index, inst in enumerate(instructions):
>         expanded_instruction = calibrations.expand(inst, [])
>         source_map += [logical_index]*len(expanded_instruction)
>         expanded_instructions += expanded_instruction
>     expanded_program.add_instructions(expanded_instructions)
>     return expanded_program, source_map
> 
> expanded_program, source_map = expand_with_source_mapping(quil_program)
> default_expanded_program = quil_program.expand_calibrations()
> assert expanded_program == default_expanded_program
> ```

@bramathon that snippet doesn't:

* break out recursive calibrations
* capture which calibration was used to do the expansion

Some of the diff, also, improves how calibrations are handled and matched

bramathon:
Yup, that's true. I don't have any need for those things for my use case.

quil-rs/src/program/mod.rs
```diff
@@ -693,14 +773,38 @@ impl ops::AddAssign<Program> for Program {
     }
 }
 
+type InstructionIndex = usize;
```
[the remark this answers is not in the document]
kalzoo: I realize I had missed this comment on the last round, but it's a good idea. Changed that now.

kalzoo:
Over to you now @bramathon

PR #374, opened 2024-06-05
feat: make all instructions and contents Eq and Hash

kalzoo:
Make `Instruction` and all of its sub-structs both `Hash` and `Eq`:

- Handle f64 equality as `NaN == NaN` which is defensible in the case of equating instructions
- Handle `IndexMap` equality by relying on iteration order. This is suboptimal in that ordering shouldn't matter for equality checks, but sorting them would be expensive. This is where `BTreeMap` may actually make more sense, but that doesn't need to be addressed here.

TODO

- right now, we just `impl Eq for TypeX` when it contains `f64`, which in effect will just cause a panic when one of those `f64` is `NaN`. this should be better handled and possibly include a `PartialEq` override.

### DIFF
--- a/quil-rs/src/instruction/calibration.rs
+++ b/quil-rs/src/instruction/calibration.rs
@@ -18,7 +18,7 @@ pub trait CalibrationSignature {
     fn has_signature(&self, signature: &Self::Signature<'_>) -> bool;
 }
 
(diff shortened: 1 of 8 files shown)


antalsz:
@asaites Am I correct that this has been superseded by the latest work and is closable?

PR #376, opened 2024-06-06
feat!: change Rust representation of classical instructions

### LINKED ISSUE #375: The representations of classical instruction types should be consistent
As it stands, in [quil_rs::instruction::classical](https://github.com/rigetti/quil-rs/blob/main/quil-rs/src/instruction/classical.rs), we have

```
pub struct Arithmetic {
    pub operator: ArithmeticOperator,
    pub destination: ArithmeticOperand,
    pub source: ArithmeticOperand,
}

pub type BinaryOperands = (MemoryReference, BinaryOperand);

pub struct BinaryLogic {
    pub operator: BinaryOperator,
    pub operands: BinaryOperands,
}

pub struct Comparison {
    pub operator: ComparisonOperator,
    pub operands: (MemoryReference, MemoryReference, ComparisonOperand),
}
```

This is inconsistent, and we're not gaining anything by the inconsistency.  All three of these structs should contain explicit fields for their operands, à la `Arithmetic`, as there's no need to pass around the tuples at any point; additionally, the `Arithmetic` struct ought to take its `destination` as a `MemoryReference`, as the other two do.  Thus:

```
pub struct Arithmetic {
    pub operator: ArithmeticOperator,
    pub destination: MemoryReference,
    pub source: ArithmeticOperand,
}

pub struct BinaryLogic {
    pub operator: BinaryOperator,
    pub destination: MemoryReference,
    pub source: BinaryOperand,
}

pub struct Comparison {
    pub operator: ComparisonOperator,
    pub destination: MemoryReference,
    pub lhs: MemoryReference,
    pub rhs: ComparisonOperand,
}
```

antalsz:
This makes the implementations of `quil_rs::instruction::classical::{Arithmetic, BinaryLogic, Comparison}` consistent, removing some imprecision from the definition of `Arithmetic` along the way.

Closes #375.


antalsz:
The test failures here are due to the problem fixed in #368, and should go away when that's merged

quil-rs/src/instruction/classical.rs
```diff
@@ -5,14 +5,14 @@ use super::MemoryReference;
 #[derive(Clone, Debug, Hash, PartialEq)]
 pub struct Arithmetic {
     pub operator: ArithmeticOperator,
-    pub destination: ArithmeticOperand,
+    pub destination: MemoryReference,
```
MarquessV, line 8: Nice! This change alone simplifies `Arithmetic` significantly.

MarquessV (approved):

quil-rs/src/instruction/classical.rs
antalsz: Unsurprisingly, this change was the specific motivation for doing the rest! :-)

PR #378, opened 2024-06-13
chore: remove deprecated `f64` constants to satisfy Clippy

antalsz:
The MSRV Clippy check was failing; looks like `std::f64::EPSILON` is now `f64::EPSILON`.


BatmanAoD:
I'm surprised by the deprecation of `std::f64` but I guess it makes sense. Docs: https://doc.rust-lang.org/std/f64/constant.EPSILON.html

BatmanAoD (approved):

PR #382, opened 2024-07-25
feat: Initialize `Instruction`s from a Quil string. Python `Instruction`s support the `pickle` module.

MarquessV:
This is to support [pyQuil#1791](https://github.com/rigetti/pyquil/issues/1791)

todo:

- [x] Apply `impl_pickle_for_instruction` everywhere.


quil-rs/src/instruction/mod.rs
```diff
@@ -763,6 +762,29 @@ impl Instruction {
...
+        let input = LocatedSpan::new(s);
+        let lexed = lex(input).map_err(|e| ParseInstructionError::Parse(e.to_string()))?;
+        let instructions =
+            parse_instructions(&lexed).map_err(|e| ParseInstructionError::Parse(e.to_string()))?;
+        if instructions.1.len() != 1 {
+            return Err(ParseInstructionError::ZeroOrMany(instructions.1.len()));
+        }
+        Ok(instructions.1[0].to_owned())
+    }
+}
```
MarquessV, line 786: Are we good with this being part of the public API?

quil-py/src/lib.rs
```diff
@@ -45,6 +45,63 @@ macro_rules! impl_to_quil {
...
+            // __reduce__ must return a tuple containing the necessary components to successfully
+            // construct a class instance.
+            //
+            // In this case, we initialize a new instance using _fromstate with the state
+            // generateed by __getstate__.
+            //
+            // See [Python's __reduce__ documentation](https://docs.python.org/3/library/pickle.html#object.__reduce__)
+            fn __reduce__<'py>(&'py self, py: Python<'py>) -> pyo3::PyResult<&'py pyo3::PyAny> {
+                use pyo3::IntoPy;
+                let callable = py.get_type::<Self>().getattr("_from_state")?;
```
MarquessV, line 80: This took a bit to figure out how to do correctly, but I'm happy with it now.

quil-py/src/lib.rs
```diff
@@ -45,6 +45,66 @@ macro_rules! impl_to_quil {
     };
 }
 
+/// Implements pickling for an instruction by implementing __getstate__ and __reduce__.
```
BatmanAoD, line 48: I know you link to the `__reduce__` documentation below, but it would probably be better to link to that here as well, or even instead of doing it on `__reduce__` itself.

BatmanAoD (approved):
I'm not familiar with this method of pickling, and I don't have an answer for your question about the public API (I don't know what harm it could do), but otherwise this looks good.

jselig-rigetti:
This is great! I see there's already a test for `Program` pickling at https://github.com/rigetti/quil-rs/blob/main/quil-py/test/program/test_program.py#L11-L23 does that use this now or is that doing something else?

quil-py/src/lib.rs
MarquessV: Updated the docstring per your comment.

MarquessV:
> This is great! I see there's already a test for `Program` pickling at https://github.com/rigetti/quil-rs/blob/main/quil-py/test/program/test_program.py#L11-L23 does that use this now or is that doing something else?

Yeah, good eye. We already had a handwritten solution for `Program` previous to this. It hasn't changed with this PR, since the new macro is really built for the Instruction types. The implementation in the macro wouldn't be valid for Program.

Issue #251, opened 2023-07-12
Support an immutable program


MarquessV:
Downstream, immutable programs could be a helpful tool. For example, a template program that can be used as a base for similar programs, but the template itself shouldn't change. Maybe this is scope creep, but supporting immutability is a something a Rust SDK is well suited to solve. In Python it's hard to guarantee the immutability of something, but in a Rust SDK w/ Python bindings it's relatively trivial to lock things down.

I think it could be implemented as a new type wrapper around a private `Program`:

```rust
pub struct ImmutableProgram(Program)
```

with methods to go from a `Program` -> `ImmutableProgram` and back. The API could be as simple as that, but we could also port read only methods to the new type wrapper.

MarquessV:
An additional observation from @kalzoo: Since guaranteeing immutability in Rust is a simple as not using `mut`, it may make sense to only implement this new type wrapper as a Python binding. It could be accessed idiomatically with a `freeze` method and provide a read-only subset of the `Program` API. This way we don't clutter the Rust API with a type that isn't necessarily needed.

PR #384, opened 2024-07-31
feat(quil-py): Add an immutable FrozenProgram class

MarquessV:
This is a proposal for #251 

The API is very simple, and only supports serialization at the moment.

### DIFF
--- a/quil-py/src/program/mod.rs
+++ b/quil-py/src/program/mod.rs
@@ -5,6 +5,7 @@ use std::{
 
 use indexmap::IndexMap;
 use numpy::{PyArray2, ToPyArray};
+use pyo3::types::PyTuple;
 use quil_rs::{
     instruction::{Instruction, QubitPlaceholder, TargetPlaceholder, Waveform},
     program::{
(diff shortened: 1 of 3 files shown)


quil-py/src/program/mod.rs
```diff
@@ -373,6 +378,58 @@ impl PyProgram {
     }
 }
 
+#[pyclass]
+#[derive(Debug, Clone, PartialEq)]
+pub struct FrozenProgram {
+    program: PyProgram,
+}
+impl_eq!(FrozenProgram);
```
MarquessV, line 386: Question: The `__richcmp__` implementation this macro generates will only compare two `FrozenProgram`s with equivalent inner `Program`s as equal.

What about the case where a `FrozenProgram` is compared to a `Program`. If the inner program of a `FrozenProgram` instance is equivalent to an instance of `Program`, should they evaluate as equivalent?

mhodson-rigetti: I'm flexible. If converting a Program to a FrozenProgram is a light operation (it encapsulates the Program and exposes a limited immutable interface but does not deeply copy the program) then at the level of `quil-py` I would adhere to its Rust heritage and force the user to write "if frozen_program == FrozenProgram(program):".

How this is net exposed in `pyquil` is a separate question.

mhodson-rigetti: Strike that ill-informed comment of mine.

I can see it takes the clone. Of course it needs to as the user has a mutable reference to it already.

Let me think about the performance implications of typical workflows here and get back to you.

PR #390, opened 2024-08-21
feat!: Use 4 spaces for indentation.

### LINKED ISSUE #385: The parser is not robust to whitespace
The parser considers 4 spaces as an indentation token, which means that 8 spaces should be considered 2 indentations. The parser is "flexible" in that it allows for extra spaces after the initial 4, but that has less precedence than recognizing an additional indentation, so 4-7 spaces becomes the valid range for a single indentation. This doesn't necessarily seem incorrect to me, but it is brittle, and not really made clear to the user.

Since quil-rs serializes program to Quil using tabs for whitespace, and many editors and terminals default to rendering 8 spaces for tabs, this leads to a surprising error when attempting to roundtrip a serialized program copied from one's editor or terminal.

Ideas:

* Allow configuration of whitespace when serializing a program. I can see this as an enum with a method that returns the desired whitespace. The serialization routine would then inject that whitespace instead of hardcoding '\t'
* Implement Python-esque handling of whitespace, where indentation must be consistent within a block, but the composition of said whitespace can vary.

MarquessV:
closes #385 

TODO
- [x] Apply changes based on [this discussion](https://github.com/rigetti/quil-rs/pull/390#discussion_r1725429383). Until then, CI is not expected to pass.

### DIFF
--- a/quil-rs/src/parser/lexer/mod.rs
+++ b/quil-rs/src/parser/lexer/mod.rs
@@ -18,9 +18,9 @@ mod wrapped_parsers;
 
 use nom::{
     bytes::complete::{is_a, take_till, take_while, take_while1},
-    character::complete::{digit1, one_of},
+    character::complete::{digit1, one_of, space0},
     combinator::{all_consuming, map, recognize, value},
-    multi::many0,
+    multi::{many0, many_m_n},
     number::complete::double,
     sequence::{pair, preceded, terminated, tuple},
     Finish, IResult,
@@ -135,23 +135,25 @@ pub(crate) fn lex(input: LexInput) -> Result<Vec<TokenWithLocation>, LexError> {
         .map_err(LexError::from)
 }
 
-/// TODO: Discuss:
-///     * Should we make indentation exactly 4 spaces, rather than 4-7?
-///     * Should we allow for multiple levels of indentation? The Quil spec implies that only a
-///       single level indentation is possible, as each indentation must be preceded by a newline
-///     * What are the implications of using spaces instead of tabs for program serialization?
-///       It is technically 4x the bytes for indentations in serialized program. Seems minor, but
-///       maybe tabs were chosen for a reason?
 fn _lex(input: LexInput) -> InternalLexResult<Vec<TokenWithLocation>> {
     terminated(
         many0(alt(
             "indentation or a token preceded by whitespace",
-            (
-                token_with_location(value(Token::Indentation, tag("    "))),
-                preceded(many0(tag(" ")), lex_token),
-            ),
+            (lex_indent, preceded(many0(tag(" ")), lex_token)),
         )),
         many0(one_of("\n\t ")),
+    )(input)
+}
+
+/// The Quil spec defines an indent as exactly 4 spaces. However, the lexer recognizes tabs as well
+/// to allow for more flexible formatting.
+fn lex_indent(input: LexInput) -> InternalLexResult<TokenWithLocation> {
+    alt(
+        "indentation",
+        (
+            token_with_location(value(Token::Indentation, tag("    "))),
+            token_with_location(value(Token::Indentation, tag("\t"))),
+        ),
     )(input)
 }
 
--- a/quil-rs/src/parser/lexer/mod.rs
+++ b/quil-rs/src/parser/lexer/mod.rs
@@ -135,6 +135,13 @@ pub(crate) fn lex(input: LexInput) -> Result<Vec<TokenWithLocation>, LexError> {
         .map_err(LexError::from)
 }
 
+/// TODO: Discuss:
+///     * Should we make indentation exactly 4 spaces, rather than 4-7?
+///     * Should we allow for multiple levels of indentation? The Quil spec implies that only a
+///       single level indentation is possible, as each indentation must be preceded by a newline
+///     * What are the implications of using spaces instead of tabs for program serialization?
+///       It is technically 4x the bytes for indentations in serialized program. Seems minor, but
+///       maybe tabs were chosen for a reason?
 fn _lex(input: LexInput) -> InternalLexResult<Vec<TokenWithLocation>> {
     terminated(
         many0(alt(
--- a/quil-rs/src/parser/lexer/mod.rs
+++ b/quil-rs/src/parser/lexer/mod.rs
@@ -18,9 +18,9 @@ mod wrapped_parsers;
 
 use nom::{
     bytes::complete::{is_a, take_till, take_while, take_while1},
-    character::complete::{digit1, one_of, space0},
+    character::complete::{digit1, one_of},
     combinator::{all_consuming, map, recognize, value},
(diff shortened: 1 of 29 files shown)


quil-rs/src/parser/lexer/mod.rs
```diff
@@ -135,6 +135,13 @@ pub(crate) fn lex(input: LexInput) -> Result<Vec<TokenWithLocation>, LexError> {
         .map_err(LexError::from)
 }
 
+/// TODO: Discuss:
+///     * Should we make indentation exactly 4 spaces, rather than 4-7?
+///     * Should we allow for multiple levels of indentation? The Quil spec implies that only a
+///       single level indentation is possible, as each indentation must be preceded by a newline
+///     * What are the implications of using spaces instead of tabs for program serialization?
+///       It is technically 4x the bytes for indentations in serialized program. Seems minor, but
+///       maybe tabs were chosen for a reason?
```
MarquessV, line 144: I am in favor of using exactly 4 spaces for indentation, as that most closely aligns with the Quil spec.

Not allowing for multiple levels of indentation would also be in line with the Quil spec, but Program serialization currently allows for that (e.g by adding a DEFCIRCUIT to another DEFCIRCUIT), these cases would fail to roundtrip. That is a potential UX issue, but could probably be solved by checking for nested indentation in any instruction that indents a body of instructions.

Shadow53: 1. I am also in favor of enforcing exactly 4 spaces
2. I'd expect nested indentation not to be used in practice anywhere, so I don't _think_ disallowing nested indentation would break much. That said, we _do_ recursively expand calibrations, so it's possible people are using nested definitions and having it work. I think it's worth maybe discussing further but not making a change here.
3. I tested this locally using a large program from [quilc](https://github.com/quil-lang/quilc) and a Rigetti calibrations file:

```
== quilc/benchmarking/quil-available/0008q-0019430-grover8bit.quil ==
Tabs:         4916568
Tabs (gz):      38937
Spaces:       4946520
Spaces (gz):    39868
Size increase:  0.061%
Size inc. (gz): 0.23%

== Ankaa-2 Calibrations ==
Tabs:          460487
Tabs (gz):      50748
Spaces:        480647
Spaces (gz):    51077
Size increase:  4.38%
Size inc. (gz): 0.65%
```

I'm not sure that 4% is worth worrying about and, if it is, IMO we should rely on compression instead of tabs vs spaces.

quil-rs/src/program/mod.rs
```diff
@@ -702,7 +702,7 @@ mod tests {
             Gate, Instruction, Jump, JumpUnless, JumpWhen, Label, Matrix, MemoryReference, Qubit,
             QubitPlaceholder, Target, TargetPlaceholder,
         },
-        quil::Quil,
+        quil::{Quil, INDENTATION as INDENT},
```
Shadow53, line 705: Why the rename? Why not just call it `INDENT` everywhere?

MarquessV: Fair, I did that last minute while trying to fix the test. I renamed the constant to INDENT.

quil-rs/src/parser/lexer/mod.rs
MarquessV: Per these discussions, and some offline, I've decided to use 4 spaces for serialization, but continue to support tabs in parsing to avoid breaking downstream users more than necessary. I also created [an issue](https://github.com/rigetti/quil-rs/issues/391) to handle multiple level of indentations.

quil-rs/src/parser/lexer/mod.rs
```diff
@@ -139,15 +139,24 @@ fn _lex(input: LexInput) -> InternalLexResult<Vec<TokenWithLocation>> {
     terminated(
         many0(alt(
             "indentation or a token preceded by whitespace",
-            (
-                token_with_location(value(Token::Indentation, tag("    "))),
-                preceded(many0(tag(" ")), lex_token),
-            ),
+            (lex_indent, preceded(many0(tag(" ")), lex_token)),
```
Shadow53, line 142: Thinking we should remove the `preceded(many0(tag(" "))` part, since we're being more strict about indentation being four spaces.

```suggestion
            (lex_indent, lex_token),
```

quil-rs/src/parser/lexer/mod.rs
```diff
@@ -139,15 +139,24 @@ fn _lex(input: LexInput) -> InternalLexResult<Vec<TokenWithLocation>> {
...
+/// to allow for more flexible formatting.
+fn lex_indent(input: LexInput) -> InternalLexResult<TokenWithLocation> {
+    alt(
+        "indentation",
+        (
+            token_with_location(value(Token::Indentation, tag("    "))),
+            token_with_location(value(Token::Indentation, tag("\t"))),
+        ),
+    )(input)
+}
```
Shadow53, line 158: Last I recalled, the plan was to only support spaces and disallow tabs. Did I misunderstand?

MarquessV: Yeah, after talking with @kalzoo we decided to keep support for `\t`. Removing it would break a lot of code here, in magneto, and surely downstream. We decided it was better to be more conservative with this change.

quil-rs/src/parser/lexer/mod.rs
MarquessV: Yeah, I was expecting to remove that too, but it becomes clear pretty quickly that `preceded` is actually necessary to catch spaces between instructions.

Shadow53: Do you have a concrete example of this? Instructions should not be separated by spaces afaik

MarquessV: Oh yeah, instructions was the wrong word there, I meant `tokens`. Having an issue with my environment, but I'll share an example of how removing the preceded parser causes an error.

Shadow53: Oh, that makes more sense. In that case, it seems like `separated_list0` or `separated_list1` makes more sense

MarquessV: Hmm, I could be missing my shot in nom golf, but I can't figure out how to make that combinator work for this case. It doesn't satisfy the trait bounds of the other wrapping combinators.

Shadow53: I believe my intent was something like this, which does change the signature of the right-hand parser from `TokenWithLocation` to `Vec<TokenWithLocation>`, but also seems more correct?

```suggestion
            (lex_indent, separated_list1(space1, lex_token)),
```

Shadow53: I suppose, in order to make this "more correct", newline parsing would have to be handled by `lex_indent` (since an indent is specifically a newline followed by four spaces). That seems like it would cause more code changes, since it seems like newlines are currently handled by `lex_token`?

PR #394, opened 2024-08-29
feat(quil-py): support extern call instructions

erichulburd:
#### High-level overview

This introduces support for `CALL` and `PRAGMA EXTERN` instructions. The former is supported directly as an `Instruction` and I've introduced a `ReservedPragma` instruction to accommodate the latter (see source code comment below for discussion on this particular choice).

There are three main aspects to this support:

* Parsing of call in `crate::parser::command::parse_call` and parsing of `ExternSignature` in `crate::parser::pragma_extern`.
* Rust representation of these instructions in `crate::instruction::extern_call`.
* Resolution of memory accesses in `crate::instruction::extern_call` and `crate::program::Program::get_memory_accesses`.

This functionality was also ported to Python in `quil-py/src/instruction/extern_call.rs`.

#### Public API Changes

There are two breaking public API changes from adding EXTERN / CALL support:

1. ~~Two~~ One new enum variant on `Instruction`: `Call` ~~and `ReservedPragma` (see source code comment below for the choice of `ReservedPragma`)~~.
2. `Instruction::get_memory_accesses` is fallible now. This reflects the fact that a `CALL` instruction cannot know its memory accesses until it has been resolved to an `ExternSignature` (ie it has to know the mutability of its different arguments.


quil-rs/src/instruction/pragma.rs
```diff
@@ -77,3 +77,26 @@ impl Include {
         Self { filename }
     }
 }
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub enum ReservedPragma {
```
erichulburd, line 82: The main alternative considered here was to convert `Pragma` into an `enum Pragma { Free(...), Reserved(enum ReservedPragma()) }`. This feels somewhat more semantically clear, but:

1. This alternative still requires the `enum ReservedPragma`.
2. We create an additional layer of enums, rather than hoisting it up to the pre-existing `enum Instruction` level. Note, there is precedence for this: `Instruction::Arithmetic` actually maps to several different Quil commands (eg `ADD`, `SUB`, etc).

Generally, the overall justification for the approach is: adding enum variants is an easier change to handle, rather than converting a `struct` to an `enum`.

I don't feel incredibly strong about this, however.

quil-rs/src/instruction/extern_call.rs
```diff
@@ -0,0 +1,2060 @@
...
+    Name(#[from] IdentifierValidationError),
+}
+
+/// A call instruction with a name and arguments.
+#[derive(Clone, Debug, PartialEq, Hash, Eq)]
+pub struct Call {
+    /// The name of the call instruction. This must be a valid user identifier.
+    pub name: String,
+    /// The arguments of the call instruction.
+    pub arguments: CallArguments,
```
erichulburd, line 649: I considered three alternatives here:

1. Just have a `Vec<CallArgument>` where each `CallArgument` has a `resolution: Option<Resolution>` attribute.
2. Do not mutate the `CallArgument`s and instead return a struct that represents the resolution, roughly of the structure ("name", instruction_index) => Vec<ResolvedCallArguments> (we need the `index` to refer to index of the `CALL` instruction within the program).
3. Add a type parameter to the `CALL` instructions and then to the program itself. Resolution would then return a program of a different type (eg `into_call_resolved_program(self) -> Result<Program<ResolveCallArgument>, ProgramError>`).

I landed on the de facto implementation because:

1. Within a given instruction, call arguments are all resolved at the same time. Either all arguments are resolved or none are.
2. There are existing patterns within Quil that mutate the program, such as `resolve_placeholders`.
3. Tracking instruction indices seems fairly unergonomic and brittle.
4. I did not want to add type complexity to the `Program` struct which is pretty easy to use.
5. I wanted to avoid the user resolving the program more than once for efficiency's sake.

I definitely feel like this resolution functionality belongs in quil-rs and not separately in downstream compilers. After going back and forth, I think this is the right implementation, but am open to input here.

quil-py/quil/instructions/__init__.pyi
```diff
@@ -1175,6 +1189,303 @@ class FrameIdentifier:
...
+@final
+class CallArgument:
+    """An argument to a `Call` instruction.
+
+    This may be expressed as an identifier, a memory reference, or an immediate value. Memory references and identifiers require a corresponding memory region declaration by the time of
+    compilation (at the time of call argument resolution and memory graph construction to be more precise).
+
+    Additionally, an argument's resolved type must match the expected type of the corresponding `ExternParameter`
+    in the `Externsignature`.
+    """
```
antalsz, line 1201:
```suggestion
    """An argument to a `Call` instruction.

    This may be expressed as an identifier, a memory reference, or an immediate value. Memory references and identifiers require a corresponding memory region declaration by the time of
    compilation (at the time of call argument resolution and memory graph construction, to be more precise).

    Additionally, an argument's resolved type must match the expected type of the corresponding `ExternParameter`
    in the `ExternSignature`.
    """
```

quil-py/quil/program/__init__.pyi
```diff
@@ -141,6 +141,21 @@ class Program:
...
+        This does two things:
+
+        1. It validates all ``EXTERN`` instructions in the program.
+        2. It resolves the ``CALL`` instruction data types based on the declared memory types
+            and ensures those match one and only one ``EXTERN`` instruction.
+
+        This method is only useful for validating program structure before it is compiled. Note,
+        this mutates the inner representation of the program.
+
+        :raises ProgramError: If any of the above conditions are not met.
```
antalsz, line 156: I'm not sure which parts of the above are "conditions"; it would be good to separate those out in the documentation.

quil-rs/src/instruction/extern_call.rs
```diff
@@ -0,0 +1,2060 @@
...
+    pub fn new(return_type: Option<ScalarType>, parameters: Vec<ExternParameter>) -> Self {
+        Self {
+            return_type,
+            parameters,
+        }
+    }
+
+    fn has_return_or_parameters(&self) -> bool {
+        self.return_type.is_some() || !self.parameters.is_empty()
+    }
```
antalsz, line 106: Maybe better to make this required in the (fallible) constructor and make the fields private?

quil-rs/src/instruction/extern_call.rs
```diff
@@ -0,0 +1,2060 @@
...
+}
+
+/// An error that can occur when validating an extern definition.
+#[derive(Debug, thiserror::Error, PartialEq)]
+pub enum ExternValidationError {
+    /// The specified name is not a valid user identifier.
+    #[error(transparent)]
+    Name(#[from] IdentifierValidationError),
+    /// There are more than one extern definitions with the same name.
```
antalsz, line 216:
```suggestion
    /// There is more than one extern definition with the same name.
```

quil-rs/src/instruction/extern_call.rs
```diff
@@ -0,0 +1,2060 @@
...
+    #[error("duplicate extern definition {0}")]
+    Duplicate(String),
+    /// The extern definition has a signature but it lacks a return or parameters.
+    #[error("extern definition {0} has a signature but it lacks a return or parameters")]
```
antalsz, line 220: What signature can it have, then?

quil-rs/src/instruction/extern_call.rs
```diff
@@ -0,0 +1,2060 @@
...
+/// [`crate::Program`] memory graph.
+#[derive(Clone, Debug, PartialEq)]
+pub enum UnresolvedCallArgument {
+    /// A reference to a declared memory location. Note, this may be resolved to either
+    /// a scalar or vector. In the former case, the assumed index is 0.
+    Identifier(String),
+    /// A reference to a memory location. This may be resolved to a scalar.
+    MemoryReference(MemoryReference),
+    /// An immediate value. This may be resolved to a scalar.
+    Immediate(Complex64),
```
antalsz, line 288:
```suggestion
    /// A reference to a memory location. This can only be resolved to a scalar.
    MemoryReference(MemoryReference),
    /// An immediate value. This can only be resolved to a scalar.
    Immediate(Complex64),
```

quil-rs/src/instruction/extern_call.rs
```diff
@@ -0,0 +1,2060 @@
...
+}
+
+impl Eq for UnresolvedCallArgument {}
```
antalsz, line 291: Is this okay, despite the `Complex64`?

quil-rs/src/instruction/extern_call.rs
```diff
@@ -0,0 +1,2060 @@
...
+        }
+    }
+
+    /// Check if the argument is compatible with the return type of the [`ExternSignature`]. If so,
+    /// return the appropriate [`ResolvedCallArgument`]. If not, return an error.
+    fn resolve_return(
+        &self,
+        memory_regions: &IndexMap<String, MemoryRegion>,
+        return_type: ScalarType,
+    ) -> Result<ResolvedCallArgument, CallArgumentResolutionError> {
```
antalsz, line 415: The return type of this function is misleading – it's not a call argument.  Can we rename the type, and maybe share code with the above function?

quil-rs/src/instruction/extern_call.rs
```diff
@@ -0,0 +1,2060 @@
...
+            ResolvedCallArgument::MemoryReference {
+                memory_reference, ..
+            } => Some(memory_reference.name.clone()),
+            ResolvedCallArgument::Immediate { .. } => None,
+        }
+    }
+}
+
+/// A list of arguments for a call instruction. These may be resolved or unresolved.
+/// To resolve a [`Call`] instruction, use [`crate::Program::resolve_call_instructions`].
```
antalsz, line 592:
```suggestion
/// A list of arguments for a call instruction. These may be either all resolved
/// or all unresolved; resolved and unresolved arguments will never be
/// intermingled in a single list.  To resolve a [`Call`] instruction, use
/// [`crate::Program::resolve_call_instructions`].
```

quil-rs/src/instruction/extern_call.rs
```diff
@@ -0,0 +1,2060 @@
...
+                    return argument
+                        .resolve_return(memory_regions, return_type)
+                        .map_err(CallArgumentError::Return);
+                }
+            }
+            let parameter_index = if signature.return_type.is_some() {
+                i - 1
+            } else {
+                i
+            };
```
antalsz, line 731: I find this indexing logic confusing.  Perhaps we could reflect it better either in the type or in another function that bundles up this logic?

quil-rs/src/instruction/extern_call.rs
```diff
@@ -0,0 +1,2060 @@
...
+                                mutable: true,
+                                data_type: ExternParameterType::FixedLengthVector(Vector {
+                                    data_type: ScalarType::Bit,
+                                    length: 2,
+                                }),
+                            },
+                        ],
+                    }),
+                },
+                expected: "foo \"INTEGER (bar : INTEGER, baz : mut BIT[2])\"",
```
antalsz, line 882: Raw quotes might be easier to read, here and below:
```suggestion
                expected: r#"foo "INTEGER (bar : INTEGER, baz : mut BIT[2])""#,
```

quil-rs/src/instruction/pragma.rs
antalsz: I think I'd prefer

```
enum Pragma {
    Extern { … },
    Unknown { // Or `Custom`, or `Generic`, or …
        pub name: String,
        pub arguments: Vec<PragmaArgument>,
        pub data: Option<String>,
    },
}
```

as it captures pragmas fairly cleanly, but I think this may relate to the question of splitting the parsed and resolved ASTs

quil-rs/src/parser/command.rs
```diff
@@ -120,6 +123,61 @@ pub(crate) fn parse_declare<'a>(input: ParserInput<'a>) -> InternalParserResult<
     ))
 }
 
+/// Parse the contents of a `CALL` instruction.
+///
+/// Note, the `CALL` instruction here is unresolved; it may only be resolved within the
+/// full context of a program. See [`crate::Program::resolve_call_instructions`].
```
antalsz, line 129:
```suggestion
/// Note, the `CALL` instruction here is unresolved; it can only be resolved within the
/// full context of a program. See [`crate::Program::resolve_call_instructions`].
```

quil-rs/src/parser/command.rs
```diff
@@ -120,6 +123,61 @@ pub(crate) fn parse_declare<'a>(input: ParserInput<'a>) -> InternalParserResult<
...
+///
+/// Call instructions are of the form:
+///     `CALL @ms{Identifier} @rep[:min 1]{@group{@ms{Identifier} @alt @ms{Memory Reference} @alt @ms{Complex}}}`
+///
+/// For additional detail, see the [Quil specification "Call"](https://github.com/quil-lang/quil/blob/7f532c7cdde9f51eae6abe7408cc868fba9f91f6/specgen/spec/sec-other.s).
```
antalsz, line 134:
```suggestion
/// For additional detail, see the ["Call" in the Quil specification](https://github.com/quil-lang/quil/blob/7f532c7cdde9f51eae6abe7408cc868fba9f91f6/specgen/spec/sec-other.s).
```

quil-rs/src/program/mod.rs
```diff
@@ -1507,4 +1562,81 @@ DEFFRAME 0 \"xy\":
             assert_eq!(new_program.to_quil().unwrap(), quil);
         }
     }
+
+    /// Test that a program with a `CALL` instruction can be parsed and properly resolved to
+    /// the corresponding `EXTERN` instruction. Finally, test that the memory accesses are
+    /// correctly calculated with the resolved `CALL` instruction.
```
antalsz, line 1568:
```suggestion
    /// Test that a program with a `CALL` instruction can be parsed and properly resolved to
    /// the corresponding `EXTERN` instruction. Additionally, test that the memory accesses are
    /// correctly calculated with the resolved `CALL` instruction.
```

quil-rs/src/instruction/extern_call.rs
antalsz: This is indeed a thorny question!  I agree that alternatives \#2 and (to a lesser extent) \#1 are inferior, but I think there's a lot to be said for \#3.  I've talked to you about this offline, but to expand here:

I think this raises the question of if we've hit the point where we want to separate the parsed AST from the "typechecked"/"resolved" AST.  I think there's a lot of merit to doing so, as it allows for capturing invariants much more cleanly.  But I'm not sure if this PR is the place to do it.

I think I might favor a design where we parameterize all our types by the stage of compilation, passing that down to where we need to make a decision, and default that type parameter to the stage that corresponds to the existing situation.  Something like the following:

```
pub trait Stage {
    type CallArgument;
}

pub enum Parsed;

impl Stage for Parsed {
    type CallArgument = UnresolvedCallArgument;
}

pub enum Resolved;

impl Stage for Resolved {
    type CallArgument = ResolvedCallArgument;
}

pub struct Program<S: Stage = Resolved> {
    // …
    instructions: Instruction<S>,
    // …
}

pub enum Instruction<S: Stage = Resolved> {
    // …
    Call(Call<S>),
    // …
}

pub struct Call<S: Stage = Resolved> {
    pub name: String,
    pub arguments: Vec<S::CallArgument>,
}
```

The upside to this is that we can bundle all the types we need to parameterize by together; the downside is the extra trait.  I think the upside is likely to be worth it, but it's not obvious.

Another limitation of this approach is that it forces the ASTs to be almost identical.  This can be good or bad.  Another approach would be to have

```
pub trait Stage {
    type AST;
}
```

even if `Parsed::AST = Program<α>` and `Resolved::AST = Program<β>` for now.

We can also hide some of the complexity by having the parser return a `Program<Parsed>`, a function `fn resolve(parsed: Program<Parsed>) -> Result<Program<Resolved>, ResolutionError>`, and then exposing at the top level a function that simply combines the two, returning a `Result<Program, …>`, so the user is not confronted with this new API.

That said: this adds extra complexity!  I think that may be worth it, but it's not obvious.  This mutate-and-resolve approach is not a bad one, and it might even be the implementation behind the more complex version I outline above.

erichulburd:
> then exposing at the top level a function that simply combines the two, returning a Result<Program, …>, so the user is not confronted with this new API.

The one caveat here is program mutation (either mutating a parsed program or just some `Program::new()`). If the user deletes an `EXTERN` pragma, or adds another call, then our resolution is invalid. So we'll need to either expose the resolution method to the user, or re-resolve after every relevant mutation (less efficient and more complex, but also could help us guarantee correctness opaquely). My knee-jerk preference is the former option.

quil-rs/src/instruction/extern_call.rs
erichulburd: Just spoke with Kalan on this. We landed somewhere around here:

* `Pragma` doesn't need to be an `enum`. We can add something like `extern_definitions: HashMap<String, ExternDefinition>` to `Program`.
* `struct Call` will only have `arguments: Vec<UnresolvedCallArgument>`. `Call.resolve(...)` returns `Result<Vec<ResolvedCallArgument>, ...>`. This will be a public function for the purposes of translating.

There's a bit of inefficiency here WRT resolving in `get_memory_accesses` (still fallible) and then resolving for translation. I may think a bit more about that, but otherwise, this seems tenable to me.

antalsz: So is the idea here that resolution doesn't need to produce a new `Program`, just store the `extern_definitions`?  And then any processing that needs to consume a resolved `Call` will just call `.resolve` in situ when necessary?  I think this seems fine, if a bit of kicking the can down the road wrt a new AST – but as I said above, this PR is like not the right place for that anyway, so I don't think that's an issue.

erichulburd: Yup, you got it. Here I'm just going for consistency with the existing implementation but I'm onboard with your general vision for generating a separate "validated and resolved" AST. We'll keep the conversation going.

quil-rs/src/instruction/extern_call.rs
erichulburd: This is an odd part of the spec, which basically requires that the `CALL` has at least one argument. I get that the intent is to require that statefulness is maintained in the user memory space, however, this seems a futile requirement to try to enforce because a user could easily workaround it:

```
DECLARE useless_bit BIT[1]
PRAGMA EXTERN foo "(ignored : BIT[1])"
CALL foo useless_bit[0] # the bit is ignored and the function mutates state
```

I'm partial to dropping this requirement in the implementation as well as the spec.

quil-rs/src/instruction/extern_call.rs
erichulburd: Yes on the fallible constructor. Private fields present a couple issues:

* The Python package uses this `py_wrap_data_struct!` macro that requires the fields to be public.
* Most other instructions in quil-rs have fully public fields.

There's not really precedent to create an intermediate in `quil-py` with public fields that then converts to a `quil-rs` `Call`, so I hesitate to introduce one here.

antalsz (commented):
I overall really like this representation, and I think it's a big improvement.  I have some questions around exactly where we perform resolution, and I think you may be able to punt some (or most?) of them away with "we'll figure this out when we figure out what broader program checking looks like".  I also have various specific comments.

One general note: I wonder if we should provide a way to trim out unused `PRAGMA EXTERN`s?  I believe `quil-rs` provides that for unused calibrations etc. from the CLI; we should add removing unused `PRAGMA EXTERN`s to that, I think.

quil-rs/src/parser/command.rs
```diff
@@ -120,6 +122,58 @@ pub(crate) fn parse_declare<'a>(input: ParserInput<'a>) -> InternalParserResult<
...
+        map(
+            parse_memory_reference_with_brackets,
+            UnresolvedCallArgument::MemoryReference,
+        ),
+        map(token!(Identifier(v)), UnresolvedCallArgument::Identifier),
+        map(parse_immediate_value, UnresolvedCallArgument::Immediate),
+    ))(input)
+}
+
+fn parse_immediate_value(input: ParserInput) -> InternalParserResult<Complex64> {
```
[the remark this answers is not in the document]
erichulburd: All of them, actually. All numbers are represented as `Complex64` within Quil `Expression`s, the other place where immediate values are parsed. I suppose the alternative here would be to parse something like an enum with integer, float, and complex variants. That's without precedent and I'm not 100% convinced it'd be worth it.

quil-rs/src/instruction/extern_call.rs
erichulburd: Ok there is `rigetti_pyo3::py_wrap_type!` to handle this case. It's a bit more verbose, but otherwise fine to support this.

quil-rs/src/instruction/extern_call.rs
```diff
@@ -59,16 +65,16 @@ impl Quil for ExternParameterType {
...
+    pub(crate) mutable: bool,
     /// The data type of the parameter.
-    pub data_type: ExternParameterType,
+    pub(crate) data_type: ExternParameterType,
 }
 
 impl ExternParameter {
-    /// Create a new extern parameter. This validates the parameter name as a user
-    /// identifier.
+    /// Create a new extern parameter. This will faile if the parameter name
```
antalsz, line 76:
```suggestion
    /// Create a new extern parameter. This will fail if the parameter name
```

quil-rs/src/parser/command.rs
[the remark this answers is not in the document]
antalsz: Thanks for that pointer – I found the existing parsing code now, at https://github.com/erichulburd/quil-rs/blob/393-extern_call/quil-rs/src/parser/expression.rs#L78-L102 .  I'd like to share code between these two implementations, since it's fiddly – that was the comment I was going to write but then I couldn't find the other parser! :-)

antalsz (approved):
Thanks for the detailed response, Eric, this looks great – I have a couple of tiny comments (one comment typo and some desire for code deduplication), but other than that I think it looks ready to go!

PR #395, opened 2024-09-04
feat: allow using instruction handler when simplifying program

Shadow53:
Users of `quil-rs` can use `InstructionHandler` to override how instructions are associated with frames, but this is not considered by `Program::into_simplified`. This PR adds a method, `InstructionHandler::simplify_program`, that uses the frame matching of `InstructionHandler` to determine which frames to retain in the program.


kalzoo (approved):

kalzoo:
This is a breaking change, right? New trait method

Issue #393, opened 2024-08-23
Support parsing of EXTERN / CALL instructions


erichulburd:
https://github.com/quil-lang/quil/pull/73 introduced `EXTERN` / `CALL` instructions to the Quil spec. We should support parsing and handling of these instructions.

erichulburd:
Closed by https://github.com/rigetti/quil-rs/pull/394

PR #404, opened 2024-09-20
393 extern call refactor

erichulburd:
Alternative implementation for #393 . May supersede https://github.com/rigetti/quil-rs/pull/394.


erichulburd:
Actually, let's keep the conversation in https://github.com/rigetti/quil-rs/pull/394

PR #406, opened 2024-09-20
fix!: correctly compute duration for `erfsquare` waveform templates

### LINKED ISSUE #405: The computed duration of `erfsquare` waveform templates is incorrect
`ScheduledBasicBlock::get_waveform_duration_seconds` assumes that the duration of all built-in waveforms is equal to their `duration` parameter.  This isn't true for `erfsquare`, though; the duration is just that of the main pulse, and the `padleft` and `padright` parameters specify silent padding on either side, which contributes to the total duration.

To see this, consider the following program:

```
DEFFRAME 0 "a":
    SAMPLE-RATE: 1e9
PULSE 0 "a" erfsquare(duration: 1.0, padleft: 0.2, padright: 0.3)
PULSE 0 "a" erfsquare(duration: 0.1, padleft: 0.7, padright: 0.7)
PULSE 0 "a" erfsquare(duration: 0.5, padleft: 0.6, padright: 0.4)
FENCE
```

If provided as a test case to `program::schedule::tests::schedule_seconds`, it will produce instruction timings of `[0.0, 1.0, 1.1, 1.6]`, when it should instead produce `[0.0, 1.5, 3.0, 4.5]`.

antalsz:
`ScheduledBasicBlock::get_waveform_duration_seconds` previously assumed that the duration of all built-in waveforms is equal to their duration parameter.  This isn't true for `erfsquare`, though; the `duration` parameter just specifies the duration of the active pulse, and the `padleft` and `padright` parameters specify silent padding on either side which contribute to the total duration.

This PR updates the computation of the duration to include the padding.  This is a breaking change, because now scheduling a program with an `erfsquare` waveform requires that the `erfsquare` waveform have these extra parameters and that they be constants.  Indeed, some of our very own tests were broken by this.

There is also some inconsistency in names, which this PR papers over by accepting both:

* The Quil spec calls the waveform `erfsquare`, but our test cases call it `erf_square`
* The Quil spec calls the padding parameters `padleft` and `padright`, but our `ErfSquare` type calls them `pad_left` and `pad_right`.

As implemented, any mix of the underscoreless and underscoreful names can be used.

Closes #405.

### DIFF
--- a/quil-py/tests_py/program/test_program.py
+++ b/quil-py/tests_py/program/test_program.py
@@ -134,8 +134,8 @@ DEFCAL CZ q0 q1:
     FENCE q0 q1
     SET-PHASE q0 "flux_tx_cz" 0.0
     SET-PHASE q1 "flux_tx_iswap" 0.0
-    NONBLOCKING PULSE q0 "flux_tx_cz" erf_square(duration: 6.000000000000001e-08)
-    NONBLOCKING PULSE q1 "flux_tx_iswap" erf_square(duration: 6.000000000000001e-08)
+    NONBLOCKING PULSE q0 "flux_tx_cz" erf_square(duration: 6.000000000000001e-08, pad_left: 0.1e-08, pad_right: 0.1e-08)
+    NONBLOCKING PULSE q1 "flux_tx_iswap" erf_square(duration: 6.000000000000001e-08, pad_left: 0.1e-08, pad_right: 0.1e-08)
     SHIFT-PHASE q0 "flux_tx_cz" 1.0
     SHIFT-PHASE q1 "flux_tx_iswap" 1.0
     FENCE q0 q1
--- a/quil-rs/src/program/analysis/control_flow_graph.rs
+++ b/quil-rs/src/program/analysis/control_flow_graph.rs
@@ -745,8 +745,8 @@ DEFCAL CZ q0 q1:
     FENCE q0 q1
     SET-PHASE q0 "flux_tx_cz" 0.0
     SET-PHASE q1 "flux_tx_iswap" 0.0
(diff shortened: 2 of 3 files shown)


MarquessV (approved):

kalzoo:
Good start, but I'd propose two changes:

1. The Quil spec can be "wrong." In fact, the only public-facing implementation is ours, and was written by some of the same people writing the spec. If `erf_square` and `pad_*` are how the template has been invoked for 4 years now, then we should just correct the spec, it was likely an oversight.
2. In my view it is tech debt to special-case the `erf_square` waveform like this. Just look for `pad_left` and `pad_right` on all waveforms and add those to the `duration`. Using `duration` naively and assuming it's in seconds and does not include the padding is already a short cut; IMO we don't make the shortcut any worse by making assumptions about duration-adjacent parameters. Further, this could be added to the Quil spec such that `pad_*` are supported for all of today's templates.

The "most correct" fix, I think, is to actually compute every waveform's IQs and then count samples. However, that is more computationally expensive and thus not _necessarily_ superior to this approach. So let me propose a middle ground: unit tests which assert, for some varied parameters in all supported templates, that this `duration + pad_*` heuristic is identical to the actual length of the samples. So users can rely on that without having to perform the computation themselves.

antalsz:
@kalzoo: I agree with your changes, but/and I think that the unit tests belong in a separate PR; to write them correctly, we have to handle parsing `WaveformTemplate`s from `WaveformInvocation`s, machinery that doesn't exist in `quil-rs` today.  I've made the changes to commit to `_` in the `pad_*` names and to always check for `pad_left` + `pad_right`.

antalsz:
@kalzoo Sorry, I claimed to have pushed changes yesterday but it looks like they didn't actually go through.  They went through now

kalzoo (approved):

PR #412, opened 2024-10-07
Add quil-plotting module

### LINKED ISSUE #411: Create a schedule plotting package
Visualizing the schedule of a quil program is an important feature for users. Program timing is often important to understanding program behaviour and the effects of noise.

To support plotting, we propose a new package called "quil-plotting", which would live in the following location:

```
    quil-rs
    ├── quil-cli                               
    ├── quil-py           
    ├── quil-plotting
    │    ├── quil_plotting       # The Python library package root.
    │    │     └── pulse.py
    │    │     └── __init__.py
    │    ├── pyproject.toml
    │    ├── README.md
    │    └── test                    # Plotting tests
    ├── quil-rs
    ├── scripts                              
    ├── .gitignore
    └── ...                           
```

The standalone package would be tested and released along with `quil-py` and `quil-rs`. The package will be fairly small, containing additional python dependencies for plotting, and functions to create a plot. An example of such a plot is attached.

gate-view

We envision that the package may grow to include other visualizations of quil programs, including multi-block visualizations, graph visualizations or ZX-diagrams. However, for the time being, we restrict the scope to a pulse plot.

## Open questions

1. `quil-py` exposes a rust library in python, and thus has a number of project structures to do that. It uses the maturin build system and has a src/ directory with rust code. This is unnecessary for quil-plotting, but it's desirable to maintain consistency. Should quil-plotting use this build system and have a skeleton src/ directory?

bramathon:


quil-plotting/quil_plotting/dataframe.py
```diff
@@ -0,0 +1,346 @@
+"""dataframe module.
+-----------------
+
+This module produces a plottable dataframe with IQ values and metadata.
+
+Todo:
+- Handle runner ordering
+- Use quil waveform templates
```
MarquessV, line 8: A canonical example would be great to have in the documentation here.

quil-plotting/quil_plotting/dataframe.py
```diff
@@ -0,0 +1,346 @@
...
+
+    :param hex_color: The hex color code.
+    :param opacity: The opacity for the rgba color.
+    """
+    hex_color = hex_color.strip("#")
+    r, g, b = tuple(int(hex_color[i : i + 2], 16) for i in (0, 2, 4))
+    return f"rgba({r},{g},{b},{opacity})"
+
+
+def rgba_to_hex(rgba: Union[np.ndarray, List[int]]) -> str:
```
MarquessV, line 48: It looks like all of these functions are public. Is that necessary? Can some of them be made private such that the public API is easier to grok and maintain?

quil-plotting/quil_plotting/dataframe.py
```diff
@@ -0,0 +1,346 @@
...
+    """
+    if len(rgba) == 3:
+        return f"#{rgba[0]:02x}{rgba[1]:02x}{rgba[2]:02x}"
+    elif len(rgba) == 4:
+        return f"#{rgba[0]:02x}{rgba[1]:02x}{rgba[2]:02x}{rgba[3]:02x}"
+    else:
+        raise ValueError("RGBA array must have 3 or 4 elements")
+
+
+def expand_with_source_mapping(program: Program, include_pragmas: bool = False) -> Tuple[Program, List[int]]:
```
MarquessV, line 62: We should probably use the recently added [expand_calibrations_with_source_map](https://github.com/rigetti/quil-rs/blob/main/quil-py/quil/program/__init__.pyi#L140) instead of a manual implementation here.

quil-plotting/quil_plotting/dataframe.py
```diff
@@ -0,0 +1,346 @@
...
+    return expanded_program, source_map
+
+
+def add_plot_metadata(
+    df: pd.DataFrame,
+    runners: str = "Qubit",
+    color_by: str = "Channel Type",
+    runner_order: str = "Time (s)",
+    exclude_readout: bool = True,
+    normalize_by: Optional[str] = "Channel Type",
```
MarquessV, line 95: Instead of plain strings, can we make "categoricals" and the like enums?

quil-plotting/quil_plotting/dataframe.py
```diff
@@ -0,0 +1,346 @@
...
+            "Label": "category",
+            "Runner": "category",
+        }
+    )
+
+    return df
+
+
+def deoxidize(num) -> complex:
+    """Remove Rust from a Number or PrefixExpression."""
```
MarquessV, line 192: This mentions the `Number` variant, but doesn't handle it. Also, the `evaluate` method is available on any `Expression`, and raises an error if it cannot evaluate the expression to a complex number. So this could be simplified to a single line while keeping the same behavior. Since it is a single line, this function can likely be removed.

```python
num.evaluate({}, {})
```

quil-plotting/quil_plotting/dataframe.py
```diff
@@ -0,0 +1,346 @@
...
+
+
+def get_iqs(pulse: Pulse, waveforms: Dict, frame):
+    waveform_reference = pulse.waveform
+    name = waveform_reference.name
+    parameters = pulse.waveform.parameters
+    sample_rate = frame["SAMPLE-RATE"].inner().to_real()
+    match name:
+        case "flat":
+            waveform_envelope = FlatWaveform(**parameters)
```
MarquessV, line 211: Each of the template waveforms have an `into_iq_values` method, any reason that isn't used here?

quil-plotting/quil_plotting/dataframe.py
```diff
@@ -0,0 +1,346 @@
...
+
+    :param program:
+    """
+    # Get the instructions and the schedule
+    if hasattr(program, "_program"):
+        quil_program = program._program
+    elif isinstance(program, Program):
+        quil_program = program
+    else:
+        raise ValueError("Program must be a pyquil.Program or a quil.Program")
```
MarquessV, line 245: Hmm, should this package be aware of pyQuil's internal API? I think it would be better to remove this special case, and add a method to pyQuil's Program that can call into this package instead. That way, implementation details about pyQuil stay in pyQuil.

quil-plotting/quil_plotting/dataframe.py
```diff
@@ -0,0 +1,346 @@
...
+
+    calibrations = quil_program.calibrations
+    expanded_quil_program, source_map = expand_with_source_mapping(quil_program, include_pragmas=False)
+    blocks = expanded_quil_program.control_flow_graph().basic_blocks()
+    assert len(blocks) == 1
```
MarquessV, line 250: `ruff` should raise a lint for this. In short, `assert` is removed from optimized Python code, and it's generally better to explicitly raise an error. [Reference](https://docs.astral.sh/ruff/rules/assert/)

MarquessV (changes requested):
Looks like a good start! I wonder how much can be simplified by leveraging quil-py's TemplateWaveforms and expression evaluation. I would also like to see documented examples, and clarification on what needs to be part of the public API here.

bramathon:
> Looks like a good start! I wonder how much can be simplified by leveraging quil-py's TemplateWaveforms and expression evaluation. I would also like to see documented examples, and clarification on what needs to be part of the public API here.

The intention is to use the TemplateWaveforms (I believe this is mentioned in the TODO list and it's actually necessary to avoid a circular dependency with pyquil). At the time of writing, I don't think this feature was released so that's why I made due with the pyquil ones.

As for documentation, what do you have in mind? Should this integrate with the quil-rs documentation?

quil-plotting/quil_plotting/dataframe.py
```diff
@@ -0,0 +1,346 @@
...
+
+
+# color
+class Color(str, Enum):
+    GREEN = "#00b5ad"  # Teal
+    YELLOW = "#ffc504"  # Yellow
+    RED = "#ef476f"  # Magneta
+    BLUE = "#3d47d9"  # Palatinate Blue
+    GRAY = "#8a8b92"  # Gray
+    NAVY = "#0d0d36"  # Cetacean Blue
```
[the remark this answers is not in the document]
bramathon: It's the Rigetti colors. They aren't easily configurable. Any suggestion how to make it so?

quil-plotting/tests/test_plots/.ipynb_checkpoints/randomized_circuit_1-checkpoint.svg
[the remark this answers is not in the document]
bramathon: This is a mistake, will remove.

quil-plotting/pyproject.toml
```diff
@@ -0,0 +1,120 @@
...
+classifiers = [
+  "Development Status :: 3 - Alpha",
+  "License :: OSI Approved :: Apache Software License",
+  "Programming Language :: Python :: 3.8",
+  "Programming Language :: Python :: 3.9",
+  "Programming Language :: Python :: 3.10",
+  "Operating System :: OS Independent",
+]
+dependencies = [
+    "numpy>=1.2.1",
```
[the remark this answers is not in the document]
bramathon: Yup

quil-plotting/quil_plotting/dataframe.py
bramathon: Yes, however I'm not sure if using enums with a pandas dataframe is entirely conventional. I'm not actually sure what best practice would be here.

quil-plotting/quil_plotting/dataframe.py
bramathon: This doesn't seem to work on PrefixExpressions. Will revisit it though.

quil-plotting/quil_plotting/dataframe.py
bramathon: No, removing this is in the TODO.

PR #428, opened 2025-01-09
fix!: tokenize keywords and identifiers correctly

### LINKED ISSUE #427: Identifiers cannot begin with a keyword
When parsing keywords, `quil-rs` does not correctly tokenize its input; the input `BITS` will be parsed as the keyword `BIT` followed by the identifier `S`, and so it can't be used.  For example, `MOVE BITS 0` fails with `"expected Identifier, found DATATYPE(BIT)"`

```
> cargo run -q -- parse "MOVE BITS 0"                                                            round-trip-fails-on-mut|?5⚑2
Error: Failed to parse program from input string.

Caused by:
    0: error while parsing: at line 1, column 1 (COMMAND(MOVE)): failed to parse arguments for MOVE
    1: error while parsing: at line 1, column 1 (COMMAND(MOVE)): failed to parse arguments for MOVE
    2: error while parsing: at line 1, column 1 (COMMAND(MOVE)): failed to parse arguments for MOVE
    3: at line 1, column 1 (COMMAND(MOVE)): failed to parse arguments for MOVE
    4: at line 1, column 6 (DATATYPE(BIT)): expected Identifier, found DATATYPE(BIT)
```

antalsz:
Fix the tokenization of identifiers and keywords so that keywords are not parsed within word boundaries.  Previously, `BITS` would be parsed as the data type `BIT` followed by the identifier `S`; now, it is correctly parsed as the single identifier `BITS`.

This also fixes the tokenization of `NaN`, `Inf`, and `Infinity`; previously, these were case-insensitively parsed as floats (with the same incorrect prefix behavior, so `nana` would parse as the float `NaN` followed by the identifier `a`), but now they are correctly parsed as normal identifiers.

Finally, this PR makes the "mut" keyword lowercase-only.

Closes #427

### DIFF
--- a/quil-rs/src/parser/lexer/mod.rs
+++ b/quil-rs/src/parser/lexer/mod.rs
@@ -170,9 +170,12 @@ fn lex_token(input: LexInput) -> InternalLexResult<TokenWithLocation> {
             token_with_location(lex_string),
             // Operator must come before number (or it may be parsed as a prefix)
             token_with_location(lex_operator),
-            token_with_location(lex_number),
             token_with_location(lex_variable),
+            // Identifiers must come before numbers so that `NaN`, `Inf`, and `Infinity` aren't
+            // parsed as floats; Nom, as of version 7.1.1, will parse those strings,
+            // case-insensitively, as floats
             token_with_location(lex_keyword_or_identifier),
+            token_with_location(lex_number),
         ),
     )(input)
 }
--- a/quil-rs/src/parser/lexer/mod.rs
+++ b/quil-rs/src/parser/lexer/mod.rs
@@ -399,7 +399,7 @@ mod tests {
     use nom_locate::LocatedSpan;
     use rstest::*;
 
-    use crate::parser::common::tests::KITCHEN_SINK_QUIL;
+    use crate::parser::{common::tests::KITCHEN_SINK_QUIL, DataType};
 
     use super::{lex, Command, Operator, Token};
 
@@ -568,7 +568,13 @@ mod tests {
                 Token::Identifier("a-2".to_string()),
                 Token::Operator(Operator::Minus),
                 Token::Variable("var".to_string())
-        ])
+        ]),
+        case("BIT", vec![Token::DataType(DataType::Bit)]),
+        case("BITS", vec![Token::Identifier("BITS".to_string())]),
+        case("NaN", vec![Token::Identifier("NaN".to_string())]),
+        case("nan", vec![Token::Identifier("nan".to_string())]),
+        case("NaNa", vec![Token::Identifier("NaNa".to_string())]),
+        case("nana", vec![Token::Identifier("nana".to_string())]),
     )]
     fn it_lexes_identifier(input: &str, expected: Vec<Token>) {
         let input = LocatedSpan::new(input);
--- a/quil-rs/src/parser/lexer/mod.rs
+++ b/quil-rs/src/parser/lexer/mod.rs
@@ -16,6 +16,8 @@ mod error;
 mod quoted_strings;
 mod wrapped_parsers;
 
+use std::str::FromStr;
+
 use nom::{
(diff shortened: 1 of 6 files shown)


quil-rs/src/parser/lexer/mod.rs
```diff
@@ -568,7 +489,13 @@ mod tests {
...
                 Token::Operator(Operator::Minus),
                 Token::Variable("var".to_string())
-        ])
+        ]),
+        case("BIT", vec![Token::DataType(DataType::Bit)]),
+        case("BITS", vec![Token::Identifier("BITS".to_string())]),
+        case("NaN", vec![Token::Identifier("NaN".to_string())]),
+        case("nan", vec![Token::Identifier("nan".to_string())]),
+        case("NaNa", vec![Token::Identifier("NaNa".to_string())]),
+        case("nana", vec![Token::Identifier("nana".to_string())]),
```
jselig-rigetti, line 498: would be worth adding cases for `-Infinity, Infinity, -Inf, Inf` to validate the comment from https://github.com/rigetti/quil-rs/pull/428/files#diff-a5bcc7a39fd5fe75b07e825fc96eeaa32b6cbed091f5936b09d97f9e9112769eR174

quil-rs/src/parser/token.rs
```diff
@@ -67,6 +67,78 @@ where
     }
 }
 
+/// The subset of [`Token`]s which (a) do not contain more specific data and (b) are keywords.  Used
```
jselig-rigetti, line 70: what does `more specific data` mean here? Like arguments?

quil-rs/src/parser/token.rs
```diff
@@ -67,6 +67,78 @@ where
...
+        }
+    }
+}
+
+impl TryFrom<Token> for KeywordToken {
+    type Error = ();
+
+    fn try_from(token: Token) -> Result<Self, Self::Error> {
+        // This match is explicit so that if you add a new [`Token`] constructor you have to decide
+        // if it's a keyword.  Please do not add a top-level wildcard match here.
```
jselig-rigetti, line 107: Would adding `#![deny(clippy::wildcard_enum_match_arm)]` work?

quil-rs/src/reserved.rs
```diff
@@ -20,79 +25,36 @@ impl FromStr for ReservedToken {
...
-        if let Ok(keyword) = ReservedKeyword::from_str(s) {
-            Ok(Self::Keyword(keyword))
-        } else if let Ok(gate) = ReservedGate::from_str(s) {
-            Ok(Self::Gate(gate))
-        } else if let Ok(constant) = ReservedConstant::from_str(s) {
-            Ok(Self::Constant(constant))
-        } else {
-            Err(NotReservedToken(s.to_string()))
+        fn parse<T: FromStr>(
+            reserved: impl Fn(T) -> ReservedToken,
```
jselig-rigetti, line 29: Can the `Self` type be used here? Not saying it would be better, just curious the scoping

```suggestion
            reserved: impl Fn(T) -> Self,
```

jselig-rigetti (approved):
Really clean changes! LGTM pending CI errors

quil-rs/src/parser/token.rs
antalsz: Yes, that’s a better way to say it, thanks!

quil-rs/src/parser/token.rs
antalsz: Oh perfect, I didn’t know about that!

quil-rs/src/reserved.rs
antalsz: It actually can’t – I wrote it that way first, but you get “error[E0401]: can't use `Self` from outer item”.

quil-rs/src/parser/lexer/mod.rs
antalsz: Done, and they correctly pass – thanks!

Issue #420, opened 2024-11-12
quil-py: Calibration type stubs differ from implementation


MarquessV:
In 0.13.2, some of the accessor methods on `Calibration` and `MeasureCalibrationDefinition` changed from being properties, to being methods that need to be called. However, the type stubs have not been updated to remove the @property decorators from the type stubs. This misrepresents the API, and raise unnecessary errors in downstream projects.

jselig-rigetti:
From https://github.com/rigetti/pyquil/pull/1809#discussion_r1880875006 looks like the type stubs were upgraded? At least in the repo, not sure they made it out fully in a release though

PR #435, opened 2025-01-22
fix(py)!: make Calibration and MeasureCalibrationDefinition getter methods into properties

jselig-rigetti:
See https://github.com/rigetti/quil-rs/issues/420

It was always intended for these to be getters, this fixes the implementation and types.

### DIFF
--- a/quil-py/pyproject.toml
+++ b/quil-py/pyproject.toml
@@ -27,7 +27,7 @@ sdist-include = ["README.md"]
 dev = [
     "ruff>=0.3.7",
     "maturin>=1.2.3",
-    "mypy>=1.13.0",
+    "mypy==1.13.0",
     "pytest>=7.2.2",
     "pdoc>=14.1.0",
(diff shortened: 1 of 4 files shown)


Shadow53 (approved):

quil-py/pyproject.toml
```diff
@@ -27,7 +27,7 @@ sdist-include = ["README.md"]
 dev = [
     "ruff>=0.3.7",
     "maturin>=1.2.3",
-    "mypy>=1.13.0",
```
kalzoo, line 30: Is this necessary? Is there a regression?

kalzoo (approved):
Thanks for fixing this up, @jselig-rigetti

quil-py/pyproject.toml
jselig-rigetti: Not sure why this happened, I think `uv` made this change on my behalf. Reverted.

PR #440, opened 2025-02-18
feat(quil-py): expose pragma extern map iter

### LINKED ISSUE #439: Requirement: expose Program.pragma_extern_map to Python for copying
Downstream, pyQuil `Program.copy_everything_except_instructions` manually initializes a program with all Quil instructions that are not included in the quil-rs `Program.body_instructions`. The `pragma_extern_map` does not belong in `Program.body_instructions` because the `PRAGMA EXTERN` instructions are managed in `Program.pragma_extern_map`. Therefore, in order to properly support `Program.copy_everything_except_instructions`, we need to expose `Program.pragma_extern_map` so that the associated Pragma instructions may be copied.

On a related note, we must ensure that `Program.pragma_extern_map` is included when adding one program to another.

erichulburd:
Closes #439 . See issue description for motivation here.


quil-rs/src/instruction/extern_call.rs
```diff
@@ -295,6 +296,11 @@ impl ExternPragmaMap {
         )
     }
 
+    /// Extend the [`ExternPragmaMap`] with another.
```
antalsz, line 299:
```suggestion
    /// Extend this [`ExternPragmaMap`] with another.
```
We should also document which map wins if there's a collision of keys.

antalsz (approved):
I'm unclear why we want to expose the `pragma_extern_instructions` Python method rather than just working with the `pragma_extern_map` directly.  Other than that, LGTM.

erichulburd:
> I'm unclear why we want to expose the `pragma_extern_instructions` Python method rather than just working with the `pragma_extern_map` directly. Other than that, LGTM.

Fair enough point. I though the `ExternPragmaMap` revealed too much of what Rust is doing under the hood and from the Python client's perspective, it more straightforward to just deal with a list of instructions (e.g. pyQuil's `Program.copy_everything_except_instructions`).

That said, I'm generally more partial to clearer transparency across API boundaries, so your suggestion makes just as much sense to me. I've made that change.

antalsz (approved):

PR #442, opened 2025-03-05
fix: fix simplification so that a/(bc) = (a/b)/c

antalsz:
Now, expression simplification will try to simplify `a / (b * c)` to `(a / b) / c`.

Previously, it would try to simplify to `(a / b) * c` instead.

(Expression simplification already correctly simplifies `(x * y) / z` to `x * (y / z)`.)

### DIFF
--- a/quil-rs/src/expression/simplification/mod.rs
+++ b/quil-rs/src/expression/simplification/mod.rs
@@ -336,6 +336,12 @@ mod tests {
         "0"
     }
 
+    test_simplify! {
+        mul_inside_div_on_right,
+        "4 / (2 * x)",
+        "2 / x"
+    }
+
     // TODO doesn't fully simplify in a reasonable amount of recursion
     //     test_simplify! {
     //         the_big_one,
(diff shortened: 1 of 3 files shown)


quil-rs/src/expression/simplification/mod.rs
```diff
@@ -336,6 +336,12 @@ mod tests {
...
 
+    test_simplify! {
+        mul_inside_div_on_right,
+        "4 / (2 * x)",
+        "2 / x"
+    }
+
     // TODO doesn't fully simplify in a reasonable amount of recursion
     //     test_simplify! {
     //         the_big_one,
```
jselig-rigetti, line 347: we want the big one!!

jselig-rigetti (approved):

quil-rs/src/expression/simplification/mod.rs
antalsz: That's what I've been working on recently, actually!  Fingers crossed

PR #444, opened 2025-03-05
Expression interning

antalsz:
This PR features changes to the representation of expressions and simplification in order to:

1. Obtain structural sharing/perform common subexpression elimination: recursive expression structures are now wrapped in [`ArcIntern<Expression>`](https://docs.rs/internment/latest/internment/struct.ArcIntern.html), so that there is at most one instance of every expression; expressions with common subterms will have pointers to the same value.

2. Use this structural sharing to make simplification more efficient by caching work, so that simplification is performed at most once per expression per simplification pass.

This produces a ×10+ speedup on some translation workloads.

### DIFF
--- a/quil-rs/src/expression/mod.rs
+++ b/quil-rs/src/expression/mod.rs
@@ -32,7 +32,9 @@ use std::{
     fmt,
     hash::{Hash, Hasher},
     num::NonZeroI32,
-    ops::{Add, AddAssign, BitXor, BitXorAssign, Div, DivAssign, Mul, MulAssign, Sub, SubAssign},
+    ops::{
+        Add, AddAssign, BitXor, BitXorAssign, Div, DivAssign, Mul, MulAssign, Neg, Sub, SubAssign,
+    },
     str::FromStr,
 };
 
@@ -262,6 +264,17 @@ impl_expr_op!(Add, AddAssign, add, add_assign, Plus);
 impl_expr_op!(Sub, SubAssign, sub, sub_assign, Minus);
 impl_expr_op!(Mul, MulAssign, mul, mul_assign, Star);
 impl_expr_op!(Div, DivAssign, div, div_assign, Slash);
+
+impl Neg for Expression {
+    type Output = Self;
+
+    fn neg(self) -> Self {
+        Expression::Prefix(PrefixExpression {
+            operator: PrefixOperator::Minus,
+            expression: ArcIntern::new(self),
+        })
+    }
+}
 
 /// Compute the result of an infix expression where both operands are complex.
 #[inline]
--- a/quil-rs/src/expression/mod.rs
+++ b/quil-rs/src/expression/mod.rs
@@ -56,6 +56,23 @@ pub enum EvaluationError {
 
 /// The type of Quil expressions.
 ///
+/// Quil expressions take advantage of *structural sharing*; if a Quil expression contains the same
+/// subexpression twice, such as `x + y` in `(x + y) * (x + y)`, the two children of the `*` node
+/// will be the *same pointer*.  This is implemented through *interning*, also known as
+/// *hash-consing*; the recursive references to child nodes are done via [`ArcIntern<Expression>`]s.
+/// Creating an [`ArcIntern`] from an `Expression` will always return the same pointer for the same
+/// expression.
+///
+/// The structural sharing means that equality, cloning, and hashing on Quil expressions are all
+/// very cheap, as they do not need to be recursive: equality of [`ArcIntern`]s is a single-pointer
+/// comparison, cloning of [`ArcIntern`]s is a pointer copy and an atomic increment, and hashing of
+/// [`ArcIntern`]s hashes a single pointer.  It is also very cheap to key [`HashMap`]s by an
+/// [`ArcIntern<Expression>`], which can allow for cheap memoization of operations on `Expression`s.
+///
+/// The structural sharing also means that Quil expressions are fundamentally *immutable*; it is
+/// *impossible* to get an owned or `&mut` reference to the child `Expression`s of any `Expression`,
+/// as the use of interning means there may be multiple references to that `Expression` at any time.
+///
 /// Note that when comparing Quil expressions, any embedded NaNs are treated as *equal* to other
 /// NaNs, not unequal, in contravention of the IEEE 754 spec.
 #[derive(Clone, Debug)]
@@ -70,6 +87,9 @@ pub enum Expression {
 }
 
 /// The type of function call Quil expressions, e.g. `sin(e)`.
+///
+/// Quil expressions take advantage of *structural sharing*, which is why the `expression` here is
+/// wrapped in an [`ArcIntern`]; for more details, see the documentation for [`Expression`].
 ///
 /// Note that when comparing Quil expressions, any embedded NaNs are treated as *equal* to other
 /// NaNs, not unequal, in contravention of the IEEE 754 spec.
@@ -89,6 +109,10 @@ impl FunctionCallExpression {
 }
 
 /// The type of infix Quil expressions, e.g. `e1 + e2`.
+///
+/// Quil expressions take advantage of *structural sharing*, which is why the `left` and `right`
+/// expressions here are wrapped in [`ArcIntern`]s; for more details, see the documentation for
+/// [`Expression`].
 ///
 /// Note that when comparing Quil expressions, any embedded NaNs are treated as *equal* to other
 /// NaNs, not unequal, in contravention of the IEEE 754 spec.
@@ -114,6 +138,9 @@ impl InfixExpression {
 }
 
 /// The type of prefix Quil expressions, e.g. `-e`.
+///
+/// Quil expressions take advantage of *structural sharing*, which is why the `expression` here is
+/// wrapped in an [`ArcIntern`]; for more details, see the documentation for [`Expression`].
 ///
 /// Note that when comparing Quil expressions, any embedded NaNs are treated as *equal* to other
 /// NaNs, not unequal, in contravention of the IEEE 754 spec.
--- a/quil-rs/src/expression/mod.rs
+++ b/quil-rs/src/expression/mod.rs
@@ -264,10 +264,15 @@ impl_expr_op!(Mul, MulAssign, mul, mul_assign, Star);
 impl_expr_op!(Div, DivAssign, div, div_assign, Slash);
 
 /// Compute the result of an infix expression where both operands are complex.
-fn calculate_infix(left: &Complex64, operator: &InfixOperator, right: &Complex64) -> Complex64 {
+#[inline]
+pub(crate) fn calculate_infix(
+    left: Complex64,
+    operator: InfixOperator,
+    right: Complex64,
+) -> Complex64 {
     use InfixOperator::*;
     match operator {
-        Caret => left.powc(*right),
+        Caret => left.powc(right),
         Plus => left + right,
         Minus => left - right,
         Slash => left / right,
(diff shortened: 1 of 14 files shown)


quil-rs/src/expression/simplification/by_hand.rs
```diff
@@ -1,13 +1,20 @@
...
-    InfixOperator, PrefixExpression, PrefixOperator,
+    calculate_function, calculate_infix, interned, real, Expression, ExpressionFunction,
+    FunctionCallExpression, InfixExpression, InfixOperator, PrefixExpression, PrefixOperator,
 };
-use std::cmp::min_by_key;
 
 /// Simplify an [`Expression`].
-pub(super) fn run(expression: &Expression) -> Expression {
-    simplify(expression, LIMIT)
+pub fn run(expression: &Expression) -> Expression {
```
jselig-rigetti, line 13: What's the motivation for changing this to fully `pub`?

quil-rs/src/expression/simplification/by_hand.rs
```diff
@@ -17,74 +24,109 @@ pub(super) fn run(expression: &Expression) -> Expression {
...
+
+impl Simplifier {
+    fn new() -> Self {
+        Self {
+            memo_table: HashMap::new(),
+        }
+    }
+
+    #[allow(dead_code)]
+    fn debug_memo_table(&self, w: &mut impl std::io::Write) -> std::io::Result<()> {
```
jselig-rigetti, line 64: does/did this get used for tests?

quil-rs/src/expression/simplification/by_hand.rs
```diff
@@ -17,74 +24,109 @@ pub(super) fn run(expression: &Expression) -> Expression {
...
+        let result = match e.as_ref() {
+            // Even at a limit of 0, replace `pi` with 3.14….  This gets us the extremely useful
+            // invariant that `Expression::PiConstant` can never be returned from `simplify`.
+            Expression::PiConstant => ArcIntern::new(Expression::Number(PI)),
+
+            // We're out of gas; this is as simplified as things get
+            _ if limit == 0 => e.clone(),
+
+            // Atoms don't need to be simplified
             Expression::Address(_) | Expression::Number(_) | Expression::Variable(_) => e.clone(),
```
jselig-rigetti, line 102: if these don't need simplification can this go above the limit check?

quil-rs/src/expression/simplification/by_hand.rs
```diff
@@ -119,189 +161,125 @@ fn size(expr: &Expression) -> usize {
...
+        operator: InfixOperator,
+        right: ArcIntern<Expression>,
+        limit: Limit,
+    ) -> ArcIntern<Expression> {
+        let left = self.simplify(left, limit);
+        let right = self.simplify(right, limit);
+
         // There are … many cases here
-        match (simplify(l, limit - 1), op, simplify(r, limit - 1)) {
+        match (left.as_ref(), operator, right.as_ref()) {
```
jselig-rigetti, line 214: I wonder if it would be more efficient to have two levels of nesting, the outer one being the `operator` such that each expression doesn't have to try to match against every case.

I see there are some cases that match on multiple (or `_`) operators, so maybe that would be more trouble than it's worth

quil-rs/src/expression/simplification/by_hand.rs
```diff
@@ -311,156 +289,120 @@ fn simplify_infix(l: &Expression, op: InfixOperator, r: &Expression, limit: u64)
...
-                );
-                min_by_key(original, new, size)
-            }
+                // Recreate the original structure but having simplified the contents
+                let original_structure = interned::sub(left.clone(), right.clone());
 
-            // Multiplication with negation
+                // Factor out the negation
+                let inner_sum = self.simplify(interned::add(expression.clone(), right), limit - 1);
+                let outer_neg = self.simplify(interned::neg(inner_sum), limit - 1);
```
jselig-rigetti, line 335: why does `self.simplify` need to be called again for `outer_neg`, isn't `inner_sum` already simplified?

jselig-rigetti (commented):
`simplify_infix` is a beast of a function!

LGTM for the most part - it would be interesting to see the overhead of using `ArcIntern` plotted against the "level of repetition" in expressions, since I would expect a slight performance decrease for expressions that have no identical branches, though I am definitely not as familiar with the current performance characteristics

quil-rs/src/expression/simplification/by_hand.rs
jselig-rigetti: I guess I see that it would be fairly cheap to do, since `inner_sum` would already be cached, so you only need to do the `neg`

quil-rs/src/expression/simplification/by_hand.rs
```diff
@@ -474,395 +416,348 @@ fn simplify_infix(l: &Expression, op: InfixOperator, r: &Expression, limit: u64)
...
-                    &simplify_infix(left_b, InfixOperator::Plus, right_b, limit - 1),
-                    limit - 1,
-                )
+
+                let sum_as =
+                    self.simplify(interned::add(left_a.clone(), right_a.clone()), limit - 1);
+                let sum_bs =
+                    self.simplify(interned::add(left_b.clone(), right_b.clone()), limit - 1);
+                let mul_as_x = self.simplify(interned::mul(sum_as, x.clone()), limit - 1);
+                self.simplify(interned::add(mul_as_x, sum_bs), limit - 1)
```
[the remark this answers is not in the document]
jselig-rigetti: going to resolve this as I'm guessing `self.simplify` is cheap for cached instructions

quil-rs/src/expression/simplification/by_hand.rs
antalsz: If (!) I understood correctly, this is `pub` _relative to the containing module_, which is not `pub`, so it's just about letting us `pub(super) use` it in `simplification/mod.rs`

quil-rs/src/expression/simplification/by_hand.rs
antalsz: Yes, while I was making sure I'd written the code right – I'll add a comment to that effect

quil-rs/src/expression/simplification/by_hand.rs
antalsz: Yeah, it's six of one, I can move them

quil-rs/src/expression/simplification/by_hand.rs
antalsz: Re. changing the simplification: I did think about that, but I'm not convinced it's a good idea for the reason you say, and I don't actually know that there's an efficiency win – in all cases, the first thing you want to do is check if you have two `Expression::Number`s, so there will never be a case where you *don't* have to immediately match on both `left` and `right`.

Re. `.as_ref()` vs `owned`: I added a comment explaining this, I don't know if you caught it in your review as I might've added it after you started, but owning an expression now *does not* give you access to its children as owned, because there could be multiple references to that subexpression.  However, in the past, you had to perform deep clones in a few cases of simplification, and this removes those as a concern.  But more important is the removal of redundant simplification work that this enables, so I think it's worth it regardless.

quil-rs/src/expression/simplification/by_hand.rs
antalsz: Suppose `inner_sum` simplifies to `-x`; then you want `outer_neg` to be `x` instead of `-(-x)`.  Simplification is not preserved by constructors!  (That's I suppose a short summary of why this file is so nasty, simplification is a weird and picky property.)

quil-rs/src/expression/simplification/by_hand.rs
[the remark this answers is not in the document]
antalsz: Same answer as above; for instance, if `sum_bs = (-x + x)`, then it should simplify to `sum_bs = 0`, and so `mul_as_x` should return `0` instead of `0 * x`.

quil-rs/src/expression/simplification/by_hand.rs
```diff
@@ -474,395 +416,348 @@ fn simplify_infix(l: &Expression, op: InfixOperator, r: &Expression, limit: u64)
...
+                    left: other,
                     operator: InfixOperator::Star,
-                    right: ref same_1,
+                    right: same,
                 }),
                 InfixOperator::Slash,
-                ref same_2,
-            ) if **same_1 == *same_2 => simplify(other, limit - 1),
+                _,
+            ) if &right == same => other.clone(),
```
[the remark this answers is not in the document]
antalsz: I'm not sure if Rust allows that, but if it does I like it.  (This whole file is crying out for [`if let` guards](https://rust-lang.github.io/rfcs/2294-if-let-guard.html) to be stabilized, which would make this whole thing cleaner.)

quil-rs/src/expression/simplification/by_hand.rs
jselig-rigetti: oh, oops, I see what you mean

antalsz:
@jselig-rigetti:
> it would be interesting to see the overhead of using `ArcIntern` plotted against the "level of repetition" in expressions, since I would expect a slight performance decrease for expressions that have no identical branches

I agree, we're paying a cost here for the nasty cases; this is your classic "better big-O runtime, but what's the constant factor?".  My _intuition_ is that this breaks down as follows for simplification:
- Small expressions: Probably fine with or without this, things are fast enough that any overhead from `ArcIntern` is effectively irrelevant
- Big expressions without sharing: You're hosed either way, any overhead from `ArcIntern` is dwarfed by the size of your expression
- Big expressions with sharing: The aforementioned ×10+ speedups.

However, that's not backed up by anything, and I think you'd be right to be concerned about medium-size expressions.

`ArcIntern`s also mean expression equality, cloning, and hashing are faster everywhere, though, and that's another avenue I'd want to measure.  It seems possible, even probable, that we'd win overall even with slight hits to simplification.

quil-rs/src/expression/simplification/by_hand.rs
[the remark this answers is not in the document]
antalsz: "Why doesn't the return value also need to be passed to `self.simplify`" – the first answer is "I copied the old code", but the reason is that if I understand this file right, simplification is preserved downwards but not upwards.  What I mean by that is that _every subterm_ of a simplified term is also simplified, but it is not guaranteed that _using_ only simplified subterms to make a term makes a simplified term.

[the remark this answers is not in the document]
jselig-rigetti: >  the first answer is "I copied the old code",

Maybe I'm reading the diff wrong but the old code has `simplify(other, limit - 1)`

> simplification is preserved downwards but not upwards

Ah, I guess I missed that simplification has already occurred downwards based on this at the top of the function

```rs
        let left = self.simplify(left, limit);
        let right = self.simplify(right, limit);
```

>  but it is not guaranteed that using only simplified subterms to make a term makes a simplified term.

Ah, so it's the parent's responsibility to ensure that it's simplifying its use of this subterm, but the subterm itself cannot be further simplified. That makes sense. So the pre-change diff call to `simplify(other, limit - 1)` was superfluous since this subterm had already been simplified.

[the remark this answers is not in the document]
antalsz: No you're right, in this case I did make a code change, so it's not that I copied the old code here!  But yes, it's the fact that we start by simplifying our subterms that gets us this.

quil-rs/src/expression/mod.rs
```diff
@@ -675,6 +758,119 @@ impl fmt::Display for InfixOperator {
...
 
+/// Convenience constructors for creating [`ArcIntern<Expression>`]s for recursive use
+pub mod interned {
+    use super::*;
+
+    macro_rules! atoms {
+        ($($func:ident: $ctor:ident$(($typ:ty))?),+ $(,)?) => {
+            $(
+                #[inline(always)]
+                pub fn $func($(value: $typ)?) -> ArcIntern<Expression> {
```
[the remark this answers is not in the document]
antalsz: Done!

antalsz:
@jselig-rigetti Responded to your review and pushed a change to cache expression sizes as well

jselig-rigetti (approved):
All the changes LGTM 👍 IMO it's good to go after https://github.com/rigetti/rigetti-pyo3/pull/58 ships and this PR points to a stable release, if this gets merged with a conventional-commits `feat: ` prefix

PR #446, opened 2025-03-24
feat!: support multi-argument/extern functions in expressions

antalsz:
This branch adds support for multi-argument functions in Quil expressions, so you can write e.g.

```quil
SET-PHASE 0 "tx0" 2*some_extern_function(x, sin(y), z*z)
```

The major weakness of this PR is that I had to manually expand a call to `rigetti_pyo3::py_wrap_union_enum!` in `quil_py/src/expression.rs` and edit the result due to some type errors I couldn't fix.  This has proven very difficult to address.

One question for this MR, that can block merge or not as we decide, is what to do about type checking for extern function calls in expressions.  Typechecking can be more complex now; there's a "FUTURE WORK" comment in `quil_rs::program::type_check::should_be_real` where we simply punt on checking extern functions.  This isn't a regression, as it doesn't affect existing programs, and it's similar to how we don't check `CALL` instructions now (although for those we always succeed, and for extern functions in expressions we always fail).  Fixing this would require either (1) ignoring non-real function arguments, or (2) doing actual type-checking on those.  Both are entirely doable but maybe not necessary for v1.

### DIFF
--- a/quil-rs/src/program/type_check.rs
+++ b/quil-rs/src/program/type_check.rs
@@ -258,7 +258,9 @@ fn should_be_real(
                     expected: 1,
                 }),
             },
-            // TODO: Make this precise
+            // FUTURE WORK: We currently skip all `CALL` functions when type checking the program,
+            // and we likewise (but oppositely) assume that all extern functions are potential type
+            // errors when found in expressions.  This could be made more precise in the future.
             ExpressionFunction::Extern(_) => real_value_required(
                 instruction,
                 this_expression,


quil-rs/src/program/type_check.rs
```diff
@@ -230,9 +245,26 @@ fn should_be_real(
...
+                [argument] => should_be_real(instruction, argument, memory_regions),
+                _ => Err(TypeError::WrongNumberOfArguments {
+                    instruction: instruction.clone(),
+                    function: function.clone(),
+                    actual: arguments.len(),
+                    expected: 1,
+                }),
+            },
+            // TODO: Make this precise
+            ExpressionFunction::Extern(_) => real_value_required(
```
erichulburd, line 262: When I think of type checking extern functions, I think of a few things:

1. `Program::try_extern_signature_map_from_pragma_map` to get the `ExternSignatureMap`. This would fail on invalid signatures (which I think is appropriately fallible during type checking).
2. Use the `ExternSignatureMap` to get the function signature and do something similar to what `Call:resolve_to_signature` does. The rub here is function expressions aren't `CALL` instructions, so we probably need to rip out a bunch of code from `impl CALL` and share it with expression functions (hopefully possible without breaking changes).
3. Tie `CALL` instructions and extern functions together by implementing `impl TryFrom<Expression> for UnresolvedCallArgument` (or using `UnresolvedCallArgument` on the `ExpressionFunction::Extern` variant).

Maybe that all doesn't _need_ to happen at the point of type checking; however, those mechanics have to be setup to leverage this upstream in a non-divergent way to `CALL` instruction handling.

quil-rs/src/parser/expression.rs
```diff
@@ -316,11 +350,114 @@ mod tests {
...
+                Expression::Number(real!(3f64)).into(),
+                Expression::Number(real!(4f64)).into(),
+            ],
+        })
+    );
+
+    test!(
+        extern_call_nested,
+        parse_expression,
+        "outer(inner(x+y[1]), sin(%z), pi*(theta + 0.5))",
```
erichulburd, line 418: The [spec for CALL](https://quil-lang.github.io/#9-3-3Call-Instructions) is:

> ⟨Extern Call Instruction⟩ ::= CALL ⟨Identifier⟩(⟨Identifier⟩ | ⟨Memory Reference⟩ | ⟨Complex⟩)+

There's no further specification for call arguments used as function expressions. The builtin functions don't appear in the spec, only in the grammar IIUC, so things are open to a bit of interpretation here.

I'm inclined to go with precedent, but there are two perspectives here:

1. From the POV of builtin functions, arguments can be expressions.
  * This is more expressive (pun intended) and less restrictive (favorable in this case IMO).
  * An extern function expression is syntactically more similar to a builtin function expression than a `CALL` instruction.
3. From the POV of a `CALL` function, arguments cannot be expressions.
  * This is a simpler implementation and all we need in practice (at least at the moment as far as I know).
  * This spec is more explicit in its treatment of `CALL` than function expressions.
  * Extern function calls will ultimately have to resolve a signature just like `CALL` instructions and that resolution implementation should probably be shared.

Where does that leave us? If we're to take for granted that implementations for `CALL` instructions and extern function expressions share an implementation for argument resolution, then either:

1. We raise a syntax error when parsing expressions used as extern function arguments.
2. We parse extern function expression arguments, but raise a semantic error when we try to resolve the arguments (or when calling `UnresolvedCallArgument::try_from(Expression)`).

I don't have a strong preference for either option. If we want to generally push the spec towards expressiveness and flexibility by supporting expressions in the `CALL` instruction arguments, then (1) is obviously better. I'm not motivated to push us along that path; I leave it up to you.

erichulburd (commented):
I've a added a couple discussions to resolve before approving.

quil-rs/src/program/type_check.rs
antalsz: I agree, it's clear precisely what to do here, but it hasn't been done yet and I'm not sure if it should block this PR.  In particular – and there was supposed to be a "FUTURE WORK" comment to this effect here, not a "TODO" comment, but I forgot to commit it – `quil-rs` doesn't currently typecheck `CALL` instructions either.  So I could absolutely add type checking here but that's why I'm not sure it's necessary for v1.

quil-rs/src/parser/expression.rs
antalsz: Hmm, thanks for the point about `CALL` instructions only taking references and literals.  Per the parser, we must accept ⟨**Term**⟩ ::= ⟨**Identifier**⟩ ( ⟨**Expression List**⟩ ), so we definitely need to parse arbitrary call expressions (and you ~always want to implement parsing for that, even if you reject it later – it's much simpler).  The Quil spec says:

> Extern `CALL`s may appear in arithmetic expressions with some restrictions: the extern MUST have a declared function signature, which MUST include a return type, and which MUST NOT include any mutable parameters.

Because of those restrictions (and thank you for the foresight to include those, they seem ideal), it should be harmless to include expressions as parameters inside expressions.  Expressions are a naturally recursive structure, and I don't think we should be adding artificial restrictions here.  In particular, (1) we already forbid expressions to be used in places where we don't want the extra complexity, e.g. `MOVE` or `ADD`, and (2) our parser already accepts expressions in places where the spec requires reals/identifiers, e.g. `SET-SCALE`, and this causes no problems and is largely advantageous.  The denotation is also very clear – evaluate the expression, put it in a fresh memory location, and call the function on that.  That said, I could be convinced otherwise; the particular use case I have in mind for this would have `… func(mem_a, mem_b, 1.0) …`, which could be could be replaced with `CALL func fresh_dest mem_a mem_b 1.0; … fresh_dest …` for some fresh memory location `fresh_dest`.  It would require creating that memory location, which isn't ideal, but isn't the worst.

(I actually think we should probably allow expressions in `CALL` instructions as well, for non-`mut` arguments – it adds almost no complexity and is significantly more expressive.  But that's a separate question.)

erichulburd: One other thing I want to be explicit about here: Expression functions introduce new fidelity requirements between patch time and execution time evaluation that haven't practically existed before.

This is doesn't introduce any new or inordinate practical burden as I see it.

quil-rs/src/program/memory.rs
```diff
@@ -342,9 +342,10 @@ impl Expression {
     pub fn get_memory_references(&self) -> Vec<&MemoryReference> {
         match self {
             Expression::Address(reference) => vec![reference],
-            Expression::FunctionCall(FunctionCallExpression { expression, .. }) => {
-                expression.get_memory_references()
-            }
+            Expression::FunctionCall(FunctionCallExpression { arguments, .. }) => arguments
```
erichulburd, line 345: I think somewhere we should explicitly reference the quote you pulled from the spec:

> Extern CALLs may appear in arithmetic expressions with some restrictions: the extern MUST have a declared function signature, which MUST include a return type, and which MUST NOT include any mutable parameters.

This is why we can simply return `Vec<MemoryReference>` here (all presumed to be reads), while `Call::get_memory_accesses` returns a `MemoryAccessesResult`.

erichulburd (approved):
I encourage making the documentation comment above, but I otherwise agree to merge as-is.

I do think we should create three follow on issues (or add a comment where they might otherwise exist):

1. Implement argument resolution for `FunctionCallExpression`. Essentially, we get the `ExternSignatureMap` from `Program::try_extern_signature_map_from_pragma_map` and then do something analogous to `Call:resolve_arguments`. This may not be strictly required for downstream implementations, but it seems most appropriate to keep them in quil-rs. If this turns out to be sub-optimal for some reason or other, we can let the issue languish or close it out with an explanation.
2. Implement type checking for `CALL` and `FunctionCallExpression`. Will build on the mechanics established in (1).
3. Support expressions in CALL, MOVE, ADD, and other classical instructions. This otherwise imposes unnecessary restrictions in Quil expressiveness. We should open a PR to get this change into the spec prior to implementation.

quil-rs/src/parser/expression.rs
antalsz: I was thinking about that too, and I think the initial thing to do is reject patch time references to extern functions just as we would to writeable memory.  We can change that but that's the most circumspect option.

PR #469, opened 2025-08-27
ci: update Cargo.lock when releasing

antalsz:
This commit updates `knope.toml` so that the automatically-generated release commits contain the necessary changes to `Cargo.lock`.


jselig-rigetti:
So what does this do - a lockfile change causes a version bump, even if there are no other changes? Isn't that not recommended for libraries?

antalsz:
> > if I got this right

> Looks like I didn't :-(

Looks like I did :-)  I just needed to update the version of `knope` we were using.

jselig-rigetti (approved):

PR #477, opened 2025-09-19
fix!: memory dependencies correctly remember writes

### LINKED ISSUE #476: Memory dependency calculation forgets writes
Currently, the calculation of memory dependency constraints forgets about writes as soon as a single read depends on them:

```
MOVE x 0
MOVE y x
MOVE z y
```

will generate a dependency from `MOVE x 0` to `MOVE y x`, and nothing else.  We can see this in an existing test case; [`parametric_pulses_using_capture_results`](https://github.com/rigetti/quil-rs/blob/9db8cb4cfe75e67a4fa572e96e86ae8e59d0bdfc/quil-rs/src/program/scheduling/graphviz_dot.rs#L540-L551); the program

```
DECLARE ro BIT
DECLARE param REAL
CAPTURE 0 "ro_rx" test(a: param[0]) ro
NONBLOCKING PULSE 0 "rf" test(a: ro[0])
NONBLOCKING PULSE 1 "rf" test(a: ro[0])
CAPTURE 0 "ro_rx" test(a: param[0]) ro
NONBLOCKING PULSE 0 "rf" test(a: ro[0])
NONBLOCKING PULSE 1 "rf" test(a: ro[0])
```

produces [output](https://github.com/rigetti/quil-rs/blob/9db8cb4cfe75e67a4fa572e96e86ae8e59d0bdfc/quil-rs/src/program/scheduling/snapshots/quil_rs__program__scheduling__graphviz_dot__tests__graph__parametric_pulses_using_capture_results.snap) that lacks an edge from the first `CAPTURE 0 "ro_rx" test(a: param[0]) ro` to the second pulse, `NONBLOCKING PULSE 1 "rf" test(a: ro[0])`:

DOT output as described above

This is incorrect – it's very important that the `PULSE` wait until `ro[0]` is available!

antalsz:
This PR fixes memory dependency calculation so that it correctly remembers writes; now, *all* reads after a write depend on the write, rather than just the first one.

For instance, the test case given in #476, [`parametric_pulses_using_capture_results`](https://github.com/rigetti/quil-rs/blob/476-memory-dependencies-remember-writes/quil-rs/src/program/scheduling/graphviz_dot.rs#L540-L551), correctly gets its extra edge: the program

```
DECLARE ro BIT
DECLARE param REAL
CAPTURE 0 "ro_rx" test(a: param[0]) ro
NONBLOCKING PULSE 0 "rf" test(a: ro[0])
NONBLOCKING PULSE 1 "rf" test(a: ro[0])
CAPTURE 0 "ro_rx" test(a: param[0]) ro
NONBLOCKING PULSE 0 "rf" test(a: ro[0])
NONBLOCKING PULSE 1 "rf" test(a: ro[0])
```

now produces [scheduling output](https://github.com/rigetti/quil-rs/blob/476-memory-dependencies-remember-writes/quil-rs/src/program/scheduling/snapshots/quil_rs__program__scheduling__graphviz_dot__tests__graph__parametric_pulses_using_capture_results.snap) that correctly has an edge from the first `CAPTURE 0 "ro_rx" test(a: param[0]) ro` to the second pulse, `NONBLOCKING PULSE 1 "rf" test(a: ro[0])`.  Colorizing that edge bold green, along with the corresponding added edge in the second half of the output, results in the following, now correct, graph:

The DOT output described above

Closes #476.


asaites:
I know you got a lot of CI test failures due to the linking bug I introduced (it's now fixed, if you want to pull it in), but [the Windows ones also failed](https://github.com/rigetti/quil-rs/actions/runs/17864802001/job/50804448054?pr=477) due to one of the snapshot test cases, and that's what I also see locally -- does it need to be regenerated, or is it legit?

antalsz:
@asaites Looks like that's an actual error – I regenerated all the test cases because the stray extraneous `.snap` files were getting confusing, and I must just have missed regenerating those.  I've pushed them now.

antalsz:
Okay maybe now I actually committed all the `.snap` files for real.

mingyoungjeng:
I think it is worth mentioning that after this change, `MemoryAccessQueue` and `PreviousNodes` now behave nearly identically, where 
- `MemoryAccessQueue::reads` $\approx$ `PreviousNodes::blocking` 
    - On [read / `get_dependency_for_next_blocker`]: [Push / extend] to [reads / blocking], return [write / using].
    - (It doesn't appear the `Vec<_>` vs. `HashMap<_>` implementation is significant)
- `MemoryAccessQueue::write` $\approx$ `PreviousNodes::using` 
    - (although the former uses `MemoryAccessDependency`)
    - On [write / get_dependencies_for_next_user]: replace old [write / using], return [reads / blocking] and original [write / using]

It may be beneficial to combine these into a single struct representing the generic "multiple-reader-single-writer queue" model, push the write/capture distinction of `MemoryAccessDependency` onto the user (`ScheduledBasicBlock::build`), and optionally use newtypes if it is desirable to distinguish between classical memory accesses and frame dependencies

antalsz:
@mingyoungjeng Thanks for the catch!  I tried unifying those in a new `dependency_queue.rs` module, but I'm not 100% sure I'm happy with the trait design for the unification.  I'm also a bit worried this will have trouble extending to fine-grained memory region tracking, where `x[0] := 0` and `x[1] := 0` are allowed to be concurrent but `x[n] := 0` must count as a write to every portion of `x`.  Would love some review on this change from you or @asaites.

asaites (approved):
I left a few change requests for little documentation changes, but nothing major.

It looks like this gets the job done, so I'm not opposed to merging it as-is, though overall I feel the API is a little clunky. I made some suggestions but won't be heartbroken if you'd rather leave it be.

PR #481, opened 2025-09-24
feat: generate `.pyi` stub files' contents in a consistent order

### LINKED ISSUE #480: pyo3_stub_gen generates output in an inconsistent order
`pyo3_stub_gen` appears, through an oversight, to produce method declarations in `.pyi` files in an inconsistent order.  Thankfully the order is deterministic if the program is deterministic, but due to the use of the [`inventory`](https://docs.rs/inventory/latest/inventory/) crate, it's not deterministic if the code changes.  We should determinize the output, either in quil-rs or with an upstream PR.

antalsz:
This PR sorts the fields of `pyo3_stub_gen::StubInfo` so that, hopefully, the `.pyi` files stop changing every time the code changes.

Closes #480.


quil-rs/src/bin/stub_gen.rs
```diff
@@ -3,12 +3,316 @@
...
+            $struct:ident { $($field:ident),* $(,)? };
+        )*) => {
+            $(
+                #[automatically_derived]
+                impl ArbitraryOrd for $struct {
+                    fn cmp(&self, other: &Self) -> Ordering {
+                        // This guarantees us exhaustiveness
+                        let $struct { $($field),* } = self;
+
+                        let result = Ordering::Equal;
```
asaites, line 77: Minor suggestion: put a comment here like

```suggestion
                        // Return the first non-Equal result when comparing field pairs,
                        // or return Equal if all field pairs compare Equal.
                        let result = Ordering::Equal;
```

quil-rs/src/bin/stub_gen.rs
```diff
@@ -3,12 +3,316 @@
...
+
+        // `function` is an ordered map from function names to overload sets; since overload sets
+        // themselves can't be reordered, there's nothing else to do.
+        let _: &BTreeMap<&str, Vec<FunctionDef>> = function;
+
+        // [`VariableDef`]s are atomic; they don't themselves have internal structure that needs to
+        // be reordered.
+        let _: &BTreeMap<&str, VariableDef> = variables;
+
+        // Names (being strings) don't any adjustment.
```
asaites, line 279:
```suggestion
        // Names (being strings) don't need any adjustment.
```

asaites (approved):
I had a couple minor comments, but I think this looks good.

We might want to pull it out at some point, into its own crate, to use elsewhere. It could be worth asking the upstream maintainers to consider integrating this kind of sorting themselves, but --while this is a fair bit of code-- it's mostly boilerplate, so I don't expect maintenance to be too bad, baring major upstream changes.

Thanks again for this! It'll really cut down on noise in the diffs.

antalsz:
@asaites I agree re. upstream; I can definitely see about making a PR, but I thought this local change seemed more expedient for getting a fix to the issue fast.

PR #482, opened 2025-09-25
feat!: backport breaking changes from #479

antalsz:
This PR backports the following breaking changes from #479, as they're generally useful for #462 and we want to minimize the number of breaking change releases.  The changes are as follows

1. Deleting the unused-but-public `quil_rs::program::MemoryAccess` type.
2. Renaming the `Calibration` type to `CalibrationDefinition` for consistency, along with some updates to associated code.
3. Fixing the `MeasureCalibrationDefinition` type to correctly have a *required* qubit and an *optional* memory address, rather than the other way around.
4. Regenerating all the test files – not technically a breaking change, but very noisy.  However, it gets things into a consistent state and deletes some `.snap` files that no longer correspond to anything, so it needed to happen.


asaites (approved):
LGTM.

PR #484, opened 2025-10-12
feat!: change `InstructionHandler` to be a trait and require it everywhere in Rust

antalsz:
This PR removes the `InstructionHandler` type that is defined in terms of boxed closures, and replaces it with an `InstructionHandler` *trait*.  This has the potential to provide signficant performance improvements (and I think it's probably cleaner regardless).

This PR also removes every Rust method that could use `InstructionHandler` but doesn't, requiring a deliberate use of an instruction handler.  However, because we don't expose instruction handlers to Python, this is not true for our Python bindings.  Ideally, this should not be a change for Python users, but I would appreciate close review of this part.

### DIFF
--- a/quil-rs/src/expression/mod.rs
+++ b/quil-rs/src/expression/mod.rs
@@ -106,6 +106,21 @@ pub enum Expression {
     Variable(String),
 }
 
+#[cfg(test)]
+impl proptest::prelude::Arbitrary for Expression {
+    type Parameters = ();
+    type Strategy = proptest::prelude::BoxedStrategy<Self>;
+
+    fn arbitrary_with(args: Self::Parameters) -> Self::Strategy {
+        use self::proptest_helpers::{arb_complex64, arb_expr_custom_leaves, arb_name};
+        use proptest::prelude::*;
+
+        let () = args;
+
+        arb_expr_custom_leaves(any::<MemoryReference>, arb_name, arb_complex64).boxed()
+    }
+}
+
 /// The type of function call Quil expressions, e.g. `sin(e)`.
 ///
 /// Quil expressions take advantage of *structural sharing*, which is why the `expression` here is
@@ -1017,11 +1032,6 @@ pub mod proptest_helpers {
         })
     }
 
-    // Better behaved than the auto-derived version re: names & indices
-    pub fn arb_memory_reference() -> impl Strategy<Value = MemoryReference> {
-        (arb_name(), any::<u64>()).prop_map(|(name, index)| MemoryReference { name, index })
-    }
-
     // Better behaved than the auto-derived version via arbitrary floats
     pub fn arb_complex64() -> impl Strategy<Value = Complex64> {
         ((-TAU..TAU), (-TAU..TAU)).prop_map(|(re, im)| Complex64 { re, im })
@@ -1095,11 +1105,6 @@ pub mod proptest_helpers {
                 ]
             },
         )
-    }
-
-    /// Generate an arbitrary [`Expression`] for a property test.
-    pub fn arb_expr() -> impl Strategy<Value = Expression> {
-        arb_expr_custom_leaves(arb_memory_reference, arb_name, arb_complex64)
     }
 }
 
@@ -1239,7 +1244,7 @@ mod tests {
 
     proptest! {
         #[test]
-        fn eq(a in any::<f64>(), b in any::<f64>()) {
+        fn eq(a: f64, b: f64) {
             let first = Expression::Infix (InfixExpression {
                 left: ArcIntern::new(Expression::Number(real!(a))),
                 operator: InfixOperator::Plus,
@@ -1251,7 +1256,7 @@ mod tests {
         }
 
         #[test]
-        fn hash(a in any::<f64>(), b in any::<f64>()) {
+        fn hash(a: f64, b: f64) {
             let first = Expression::Infix (InfixExpression {
                 left: ArcIntern::new(Expression::Number(real!(a))),
                 operator: InfixOperator::Plus,
@@ -1266,17 +1271,17 @@ mod tests {
         }
 
         #[test]
-        fn eq_iff_hash_eq(x in arb_expr(), y in arb_expr()) {
+        fn eq_iff_hash_eq(x: Expression, y: Expression) {
             prop_assert_eq!(x == y, hash_to_u64(&x) == hash_to_u64(&y));
         }
 
         #[test]
-        fn reals_are_real(x in any::<f64>()) {
+        fn reals_are_real(x: f64) {
             prop_assert_eq!(Expression::Number(real!(x)).to_real(), Ok(x))
         }
 
         #[test]
-        fn some_nums_are_real(re in any::<f64>(), im in any::<f64>()) {
+        fn some_nums_are_real(re: f64, im: f64) {
             let result = Expression::Number(Complex64{re, im}).to_real();
             if is_small(im) {
                 prop_assert_eq!(result, Ok(re))
@@ -1286,7 +1291,7 @@ mod tests {
         }
 
         #[test]
-        fn no_other_exps_are_real(expr in arb_expr().prop_filter("Not numbers", |e| !matches!(e, Expression::Number(_) | Expression::PiConstant()))) {
+        fn no_other_exps_are_real(expr in any::<Expression>().prop_filter("Not numbers", |e| !matches!(e, Expression::Number(_) | Expression::PiConstant()))) {
             prop_assert_eq!(expr.to_real(), Err(EvaluationError::NotANumber))
         }
 
@@ -1299,13 +1304,13 @@ mod tests {
         }
 
         #[test]
-        fn exponentiation_works_as_expected(left in arb_expr(), right in arb_expr()) {
+        fn exponentiation_works_as_expected(left: Expression, right: Expression) {
             let expected = Expression::Infix (InfixExpression { left: ArcIntern::new(left.clone()), operator: InfixOperator::Caret, right: ArcIntern::new(right.clone()) } );
             prop_assert_eq!(left ^ right, expected);
         }
 
         #[test]
-        fn in_place_exponentiation_works_as_expected(left in arb_expr(), right in arb_expr()) {
+        fn in_place_exponentiation_works_as_expected(left: Expression, right: Expression) {
             let expected = Expression::Infix (InfixExpression { left: ArcIntern::new(left.clone()), operator: InfixOperator::Caret, right: ArcIntern::new(right.clone()) } );
             let mut x = left;
             x ^= right;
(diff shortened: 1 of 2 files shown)


quil-rs/src/program/memory.rs
```diff
@@ -45,13 +40,54 @@ pickleable_new! {
     }
 }
 
-impl Eq for MemoryRegion {}
```
asaites: You might want to add `Eq` to the `derive` list on `MemoryRegion`, since [it's `PartialEq` and all its members are `Eq`](https://rust-lang.github.io/rust-clippy/master/index.html#derive_partial_eq_without_eq). I don't see it as a major commitment, API-wise.

quil-rs/src/program/memory.rs
```diff
@@ -45,13 +40,54 @@ pickleable_new! {
...
+}
+
+impl MemoryAccesses {
+    /// Returns a new empty set of memory accesses
+    #[inline]
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// A more semantically-meaningful alias for [`Self::new`].
```
asaites: 👍

quil-rs/src/program/memory.rs
```diff
@@ -67,323 +103,150 @@ pub enum MemoryAccessType {
...
-            Expression::Variable(_) => vec![],
-        }
+    /// Return an iterator over all the memory references contained within this expression.
+    pub fn memory_references(&self) -> expression::MemoryReferences<'_> {
+        expression::MemoryReferences { stack: vec![self] }
     }
 }
 
 impl WaveformInvocation {
     /// Return, if any, the memory references contained within this WaveformInvocation.
```
asaites: While you're here anyway...

```suggestion
    /// Return, if any, the memory references contained within this `WaveformInvocation`.
```

quil-rs/src/program/mod.rs
```diff
@@ -239,27 +239,6 @@ impl Program {
...
-    ///   [`Program::extern_pragma_map`]).
-    ///
-    /// When a valid program is simplified, it remains valid.
-    ///
-    /// # Note
-    ///
-    /// If you need custom instruction handling during simplification,
-    /// use [`InstructionHandler::simplify_program`] instead.
-    #[allow(clippy::wrong_self_convention)] // It's a Breaking Change to change it now.
-    pub fn into_simplified(&self) -> Result<Self> {
```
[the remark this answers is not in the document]
asaites: And of course, the `pyo3(name = ...)` isn't strictly necessary, if you prefer setting the function name directly.

asaites (changes requested):
Mostly, this looks great. Just one change that I see: this removed `quil.program.Program.into_simplified`. I recommend adding it in `quilpy` and regenerating the stubs.

quil-rs/src/instruction/mod.rs
```diff
@@ -382,7 +361,7 @@ fn write_parameter_string(f: &mut impl std::fmt::Write, parameters: &[String]) -
 impl Quil for Instruction {
     fn write(
         &self,
-        f: &mut impl std::fmt::Write,
+        f: &mut impl fmt::Write,
```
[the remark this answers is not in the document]
antalsz: I'd rather not inflate the diff with the big change to the body of the method that this would produce, if that's okay with you – I wouldn't object to a cleanup PR that made the names consistent in either direction, though!

quil-rs/src/program/memory.rs
antalsz: I thought I had – just an oversight.  Thanks!

quil-rs/src/program/memory.rs
```diff
@@ -45,13 +40,54 @@ pickleable_new! {
...
+        Self::default()
+    }
+
+    /// A more semantically-meaningful alias for [`Self::new`].
+    #[inline]
+    pub fn none() -> Self {
+        Self::new()
+    }
+
+    pub fn union(mut self, rhs: MemoryAccesses) -> Self {
```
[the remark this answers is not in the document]
antalsz: Only kinda.  This is also the signature of binary operations in general; for instance, you have `fn Add::add(self, other: Rhs) -> Self::Output)`, which is typically just `fn Add::add(self, other: Self) -> Self` – i.e., exactly this signature.

quil-rs/src/program/mod.rs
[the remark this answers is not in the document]
antalsz: Done – I also marked it as `#[deprecated]` so that we can't accidentally use it in Rust.  Let me know what you think of this approach!

quil-rs/src/instruction/extern_call.rs
```diff
@@ -967,8 +967,10 @@ impl Call {
     }
 
     /// Return the [`MemoryAccesses`] for the [`Call`] instruction given the [`ExternSignatureMap`].
-    /// This assumes ALL parameters are read, including mutable parameters.
-    pub(crate) fn get_memory_accesses(
+    /// If the type of the `EXTERN` has a return type, then the first argument is assumed to be
+    /// written to and not read from; all other parameters are assumed to be read from, and all
+    /// other mutable parameters are assumped to be written to as well.
```
Shadow53:
```suggestion
    /// If the type of the `EXTERN` has a return type, then the first argument is assumed to be
    /// written to and not read from; all other parameters are assumed to be read from, and all
    /// other mutable parameters are assumed to be written to as well.
```

quil-rs/src/program/memory.rs
```diff
@@ -45,11 +40,54 @@ pickleable_new! {
...
+    #[inline]
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// A more semantically-meaningful alias for [`Self::new`].
+    #[inline]
+    pub fn none() -> Self {
+        Self::new()
+    }
```
Shadow53:
```suggestion
    /// A more semantically-meaningful alias for [`Self::new`].
    #[inline]
    pub fn none() -> Self {
        Self::default()
    }
```

I don't think we need both `new` and `none`. Might as well only use the semantically meaningful one.

quil-rs/src/program/memory.rs
```diff
@@ -65,323 +103,150 @@ pub enum MemoryAccessType {
...
-    CallResolution(#[from] CallResolutionError),
-}
+    /// An iterator over all the memory references contained in an expression.
+    #[derive(Clone, Debug)]
+    pub struct MemoryReferences<'a> {
+        pub(super) stack: Vec<&'a Expression>,
+    }
 
-pub type MemoryAccessesResult = Result<MemoryAccesses, MemoryAccessesError>;
+    impl<'a> Iterator for MemoryReferences<'a> {
```
Shadow53: Maybe I just haven't seen it yet, but do we have tests that this works correctly?

quil-rs/src/program/memory.rs
```diff
@@ -65,323 +103,150 @@ pub enum MemoryAccessType {
...
+    pub fn memory_references(&self) -> expression::MemoryReferences<'_> {
+        expression::MemoryReferences { stack: vec![self] }
     }
 }
 
 impl WaveformInvocation {
-    /// Return, if any, the memory references contained within this WaveformInvocation.
-    pub fn get_memory_references(&self) -> Vec<&MemoryReference> {
+    /// Return, if any, the memory references contained within this `WaveformInvocation`.
+    pub fn memory_references(&self) -> impl std::iter::FusedIterator<Item = &MemoryReference> {
```
Shadow53: Per the `FusedIterator` docs:

> Note: In general, you should not use `FusedIterator` in generic bounds if you need a fused iterator. Instead, you should just call [`Iterator::fuse()`](https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.fuse) on the iterator. If the iterator is already fused, the additional [`Fuse`](https://doc.rust-lang.org/stable/std/iter/struct.Fuse.html) wrapper will be a no-op with no performance penalty.

quil-rs/src/program/scheduling/graph.rs
```diff
@@ -243,34 +243,44 @@ impl<'a> ScheduledBasicBlock<'a> {
...
+
                     // This memory dependency now has an outgoing edge, so it is no longer a
                     // trailing classical instruction. If the memory dependency is not a classical
                     // instruction, this has no effect.
                     trailing_classical_instructions.remove(&memory_dependency.node_id);
                 }
             }
 
-            match custom_handler.role_for_instruction(instruction) {
+            let leading_instruction = leading_instruction;
```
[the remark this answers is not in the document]
antalsz: It is, yes.  The reason I went with the mutless-rebinding version rather than the group version – and I've done both in code before – is that the interior loop does more than just set up that variable, and I find it clearer to use the `let foo = { let mut foo = …; … foo };` version when defining `foo` is *all* that the nested block does.

quil-rs/src/program/memory.rs
```diff
@@ -65,323 +103,150 @@ pub enum MemoryAccessType {
...
+            // vector in the tail call case because no leaf node has more than one reference, so
+            // once we've popped a stack frame off it'll fully bottom out.
+            //
+            // Note also that in the two-subexpression case we actually swap the order from the
+            // `collect_into` example.  This is because `collect_into` reuses the same `output`, so
+            // the state is preserved and `collect_into(right, output)` appends after `left`.
+            // However, when we are emitting our results immediately, the tail call will happen
+            // *before* the delayed stack frame.
+            //
+            // And yes, this would all be simpler with `gen` blocks.
```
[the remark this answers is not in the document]
Shadow53: Informational while letting you use your judgement to determine whether to use them or not.

quil-rs/src/program/memory.rs
```diff
@@ -45,11 +40,54 @@ pickleable_new! {
...
+}
+
+impl MemoryAccesses {
+    /// Returns a new empty set of memory accesses
+    #[inline]
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// A more semantically-meaningful alias for [`Self::new`].
```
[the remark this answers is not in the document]
Shadow53: > Do we use the CMoS?

Not specifically, AFAIK. I mentioned it just because it's advice that I'd heard and internalized before

quil-rs/src/expression/mod.rs
```diff
@@ -1001,15 +1002,116 @@ pub mod interned {
...
+    pub fn arb_name() -> impl Strategy<Value = String> {
+        r"[A-Za-z_]([A-Za-z0-9_-]*[A-Za-z0-9_])?".prop_filter("Exclude reserved tokens", |t| {
+            ReservedToken::from_str(t).is_err()
+        })
+    }
+
+    // Better behaved than the auto-derived version re: names & indices
+    pub fn arb_memory_reference() -> impl Strategy<Value = MemoryReference> {
+        (arb_name(), any::<u64>()).prop_map(|(name, index)| MemoryReference { name, index })
+    }
```
Shadow53: Fun fact, you could use this as well:

```rust
#[cfg_attr(test, proptest_derive::Arbitrary)]
pub struct MemoryReference {
    // Might require `= "arb_name()"` instead
    #[proptest(strategy = arb_name)]
    pub name: String,
    pub index: u64,
}
```

PR #485, opened 2025-10-18
Waveform improvements

antalsz:
This PR improves the support for working with actual waveform data in `quil-rs`.  In particular, these means there are a few categories of changes:

1. There is now a single type for each sort of builtin waveform, e.g. `DragGaussian` for `drag_gaussian` that, thanks to the usage of generic parameters, can represent any of
    1. A concrete waveform with numeric values (`DragGaussian<Concrete>`) – the only thing we could work with previously;
    2. A parsed waveform from Quil containing expressions (`DragGaussian<Syntactic>`); or
    3. A Python-facing waveform containing arbitrary Python values (`DragGaussian<Pythonic>`).

2. The different waveform types are handled uniformly through a mix of type and macro machinery, which is exposed in a way that hides most of the complexity.  While the special type parameters are still present, the user does not need to be aware of, for instance, some of the higher-kinded type machinery that allows for easy definition of waveform sampling.

3. Waveform sampling is defined in a way that distinguishes between flat waveforms and waveforms that become an actual list of samples by using the new `IqSamples` type.  All built-in waveforms can be sampled to produce one of these results.

4. Waveform definitions and sampling are *ideally* easy enough to write that the waveform experts can dip in, write just the code they need by looking at existing patterns, and dip out again, without having to actually deal with any of the machinery that I put in place.  (This is, in fact, the reason for a lot of the machinery!)

5. There is a uniform presentation for built-in waveforms that factors out the common parameters of `duration`, `scale`, `phase`, and `detuning` (into the type `CommonBuiltinParameters`), which can be applied to any of the built-in waveforms.  This is an extension of functionality; it was not previously possible to detune flat waveforms, for instance.

6. The `stub_gen` binary has been extended to be able to edit the generated type stubs; these edits are specifically to work with generic parameters (e.g., to make classes generic).  This is admittedly kind of hacky, but is a necessary workaround to provide the right Python interface to waveforms and sampling given that py-stub-gen doesn't support generic types.

7. As an extension for very specific needs, there's also support for "partial" waveforms (e.g., `DragGaussian<Partial<Concrete>>`) that may be missing some of their parameters and which can be sampled anyway; this produces either a placeholder indicating the shape of sample that will be produced if all the parameters are filled in, or an actual sample result.


mingyoungjeng (changes requested):

quil-rs/src/quilpy/mod.rs
```diff
@@ -177,14 +177,16 @@ macro_rules! impl_to_quil {
...
+        $(
+            #[cfg_attr(feature = "stubs", gen_stub_pymethods)]
+            #[pyo3::pymethods]
+            impl $name {
+                fn __repr__(&self) -> String {
+                    format!("{self:?}")
+                }
             }
-        }
+        )*
```
asaites: I appreciate the goal here, but this is migrated to `rigetti-pyo3` now.

quil-rs/src/waveform/quilpy.rs
```diff
@@ -1,234 +1,319 @@
...
+                            .map_err(PyValueError::new_err)
+
+                    }
+                }
+            )*
+        }
+    }
+}
+
+python_waveforms! {
```
asaites: Overall, I think the macros make this easy to read, though.

asaites (commented):
I've made some comments, but think I need to spend a bit more time with the code first.

In general, my concern is that the abstraction level is grander than necessary for the goals, and that it makes the code particularly challenging to follow (and perhaps use and maintain).

What would help a lot for me for working out how this impacts Python users/what suggestions I can make for the PyO3 usage: could you update some or all of the Python tests that are currently broken?

quil-rs/src/waveform/quilpy.rs
```diff
@@ -1,234 +1,319 @@
...
+    m.add_class::<ConcreteFlat>()?;
+    m.add_class::<SyntacticGaussian>()?;
+    m.add_class::<ConcreteGaussian>()?;
+    m.add_class::<SyntacticDragGaussian>()?;
+    m.add_class::<ConcreteDragGaussian>()?;
+    m.add_class::<SyntacticErfSquare>()?;
+    m.add_class::<ConcreteErfSquare>()?;
+    m.add_class::<SyntacticHermiteGaussian>()?;
+    m.add_class::<ConcreteHermiteGaussian>()?;
     m.add_class::<BoxcarKernel>()?;
```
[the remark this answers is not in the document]
antalsz: I'm not sure what you're looking at – the only `#[pyclass…]`es I see are in `quil-rs/src/program/quilpy.rs` now.

quil-rs/src/waveform/sampling.rs
```diff
@@ -0,0 +1,78 @@
...
+    ///
+    /// This is a more optimizable special-case representation for [`IqSamples::Samples(vec![iq;
+    /// sample_count])`][Self::Samples], but is otherwise equivalent.
+    Flat { iq: Complex64, sample_count: usize },
+
+    /// A literal sequence of IQ samples.
+    Samples(Vec<Complex64>),
+}
+
+impl IqSamples {
```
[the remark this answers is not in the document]
antalsz: I couldn't figure out how to implement a Python class for a `pyo3` type (assuming you're referring to [`collections.abc.Sequence`](https://docs.python.org/3/library/collections.abc.html#collections.abc.Sequence)).  Any guidance would be appreciated!

quil-rs/src/waveform/builtin.rs
```diff
@@ -0,0 +1,1391 @@
...
+//     pub fn try_evaluate<T: WaveformData, E>(
+//         self,
+//         real: impl Fn(S::Real) -> Result<T::Real, E>,
+//         complex: impl Fn(S::Complex) -> Result<T::Complex,E>,
+//     ) -> Result<MathematicalFunction<T> ,E>  {
+//         … evaluate each field …
+//     }
+// }
+//
+// TODO Python ...
```
asaites: Is this something you will fill in before this PR merges?

quil-rs/src/waveform/builtin.rs
```diff
@@ -0,0 +1,1391 @@
...
+    iq_value * Complex64::cis(Radians::from(phase).0)
+}
+
+/// Convert polar coordinates to rectangular coordinates.
+#[inline]
+fn polar_to_rectangular(magnitude: f64, angle: Cycles<f64>) -> Complex64 {
+    Complex64::from_polar(magnitude, Radians::from(angle).0)
+}
+
+pub mod quilpy {
```
asaites: I'm not opposed necessarily to keeping this within the same file, but in most other cases, we've isolated Python-specific elements to a separate file to help keep things tidier. Is there a particular benefit here to including it here?

quil-rs/src/waveform/quilpy.rs
```diff
@@ -1,236 +1,685 @@
...
+            Waveform::Custom { .. } => None,
         }
     }
 
-    /// Get the `complex` value this `BoxcarKernel` represents.
-    #[pyo3(name = "into_iq_value")]
-    fn py_into_iq_value(&self) -> Complex64 {
-        self.into_iq_value()
+    #[gen_stub(override_return_type(
+        type_repr = "tuple[builtins.str, builtins.dict[builtins.str, builtins.complex]]",
```
asaites: It looks like this should be wrapped in `Optional` or written with `| None`.

quil-rs/src/waveform/sampling.rs
```diff
@@ -0,0 +1,92 @@
...
+}
+
+#[derive(Clone, Debug, PartialEq, serde::Serialize, serde::Deserialize, thiserror::Error)]
+pub enum SamplingError {
+    #[error(
+        "A duration of {duration} s cannot be discretized with a sample rate of {sample_rate} Hz, \
+         as the resulting number of samples ({sample_count}) \
+         is not in the representable range \u{5B}0, 2³²)."
+        // U+005B is `[` (i.e., LEFT SQUARE BRACKET), but we have to hide it with an escape so as
+        // not to confuse our homegrown PyO3 linter script (`quil-rs/scripts/lint-quil-rs.py`).
```
asaites, line 153: 😮‍💨

quil-rs/tests_py/test_waveforms.py
```diff
@@ -1,155 +1,349 @@
...
+    return in_range(-10, 10)
+
+
+@st.composite
+def common_builtin_parameters(draw: DrawFn) -> ConcreteCommonBuiltinParameters:
+    t = draw(duration())
+    s = draw(scale())
+    ϕ = draw(phase())
+    d = draw(detuning())
+    common = ConcreteCommonBuiltinParameters(duration=t, scale=s, phase=ϕ, detuning=d)
```
asaites:
```suggestion
    common = ConcreteCommonBuiltinParameters(duration=t, scale=s, phase=phi, detuning=d)
```

I appreciate that you _can_, but I'm not convinced that you _should_ 😄 .

asaites (commented):
Thanks for the hard work here; I really appreciate the documentation changes and the comprehensive the Python tests. I do have a couple of minor comments on a few things that are more style than substance, I think.

I'm going to look into what's up with CI right now, as there are a lot of failures. I think they may require some customization to the bindings' linter to handle the macros you've added.

quil-rs/src/waveform/builtin.rs
antalsz: Not particularly, I think I just wrote this module earlier.  It seemed short enough that I was happy to leave it place, but you're right that it'll be more consistent to move it out.

quil-rs/tests_py/test_waveforms.py
antalsz: Frankly I'm mostly surprised the physicists *don't* 😛  (I'll change it.)

quil-rs/src/waveform/quilpy.rs
```diff
@@ -1,236 +1,685 @@
...
+        SyntacticErfSquare,
+        ConcreteErfSquare,
+        SyntacticHermiteGaussian,
+        ConcreteHermiteGaussian,
         BoxcarKernel,
-        DragGaussian,
-        ErfSquare,
-        Gaussian,
-        HermiteGaussian
+    ],
```
asaites: `IqSamples` should actually be in the `complex_enums` section so that during initialization we'll apply the `fix_complex_enums` macro to it.

```suggestion
        SyntacticWaveform,
        ConcreteWaveform,
        SyntacticBuiltinWaveform,
        ConcreteBuiltinWaveform,
        SyntacticCommonBuiltinParameters,
        ConcreteCommonBuiltinParameters,
        ExplicitCommonBuiltinParameters,
        SyntacticFlat,
        ConcreteFlat,
        SyntacticGaussian,
        ConcreteGaussian,
        SyntacticDragGaussian,
        ConcreteDragGaussian,
        SyntacticErfSquare,
        ConcreteErfSquare,
        SyntacticHermiteGaussian,
        ConcreteHermiteGaussian,
        BoxcarKernel,
    ],
    complex_enums: [ IqSamples ],
```

asaites:
I pushed a couple commits to this branch to address the CI failures related to linting issues.

Mostly, they were false positives: the new waveform macros generating classes, so the linter needed to be taught how to interpret the macro invocation to connect it to the exported classes.

There was one legitimate issue: `IqSamples` is a "complex enum" class, so we want to apply a fix to it to ensure it can be pickled. That's as easy as moving it to the `complex_enums` section of `create_init_submodule`.

It looks like there's a new `cargo-deny` issue, though 🙄 .

antalsz:
The `cargo-deny` is trivially fixable by updating `rand` (either to a semver-compatible release or to a future version of the crate, so it's easy) – I'll do that when I apply the other changes.  Also, @mingyoungjeng has suggested some API improvements to this code that I want to apply before we merge this.

antalsz (commented):
Here are some comments I meant to submit a while ago but they got trapped in "pending"

quil-rs/src/waveform/sampling.rs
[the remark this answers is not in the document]
antalsz: I figured this out!

quil-rs/scripts/lint-quil-rs.py
```diff
@@ -53,6 +64,45 @@ def main():
...
+
+    lines = join_lines(iter_delim(ctx.lines, "{}"))
+    text = lines.text.removeprefix("python_waveforms! {").removesuffix("}")
+    parts = re.finditer(r"(?P<fields>\{.+?\})|(?P<name>[^\s {]+)", text)
+    m = next(parts, None)
+
+    while m is not None:
+        if m.lastgroup == "fields":
+            m = next(parts, None)
+            continue
```
asaites: My comments here might be moot based on what I'll say later about updates to `pyo3_stub_gen` (and our linter), but I reckon I'll mention that style-wise, I'd usually embrace the walrus:

```suggestion
    while m := next(parts, None):
        if m.lastgroup == "fields":
            continue
```

quil-rs/src/expression/mod.rs
```diff
@@ -55,11 +55,11 @@ mod simplification;
...
 #[derive(Clone, Debug, PartialEq, Eq, thiserror::Error)]
 pub enum EvaluationError {
-    #[error("There wasn't enough information to completely evaluate the expression.")]
+    #[error("there wasn't enough information to completely evaluate the expression.")]
     Incomplete,
-    #[error("The operation expected a real number but received a complex one.")]
+    #[error("the operation expected a real number but received a complex one.")]
     NumberNotReal,
-    #[error("The operation expected a number but received a different type of expression.")]
+    #[error("the operation expected a number but received a different type of expression.")]
```
asaites: If you're dropping the capital letter, do you want to drop the period, too?

quil-rs/src/bin/stub_gen.rs
```diff
@@ -4,15 +4,745 @@
...
+///    
+///    ```text
+///    def evaluate[OtherReal, OtherComplex = OtherReal](
+///        self,
+///        real: collections.abc.Callable[[Real], OtherReal],
+///        complex: collections.abc.Callable[[Complex], OtherComplex]
+///    ) -> Waveform[OtherReal, OtherComplex]:
+///    ```
+///
+/// 3. Methods may have a different type annotation placed on `self`.  For instance, this can
```
asaites, line 55: Since I'm pretty sure that can only be useful on an `@overload`, I suspect this case would be better covered by using `pyo3-stub-gen`'s features for specifying those, but doing so will require stepping up to the latest `rigetti-pyo3` version, and overall the change is going to take a little more time to get all the pieces together (particularly around the linter and CI). I don't see any reason for this to block this PR, but it _might_ mean all this work here would be better suited to just writing inline using existing `pyo3-stub-gen` features.

quil-rs/src/quilpy/errors.rs
```diff
@@ -113,6 +113,31 @@ exception!(
     "Errors that may occur when validating a Quil identifier."
 );
 
+// waveform errors
+exception!(
+    crate::waveform::WaveformParameterError,
+    quil.waveform,
+    WaveformParameterError,
+    QuilError,
```
asaites: It's probably a good idea to have a common error type in their hierarchy, as we do with other module error types (it's particularly useful for scoping tests that check for expected exceptions, if you ask me):

```python
create_exception!(
    quil.instructions,
    WaveformError,
    QuilError,
    "Base error type for errors related to waveform processing."
);
```

Then this line is:


```suggestion
    WaveformError,
```

quil-rs/src/waveform/sampling/quilpy.rs
```diff
@@ -0,0 +1,421 @@
...
+    },
+    PyStubType, TypeInfo,
+};
+
+use crate::waveform::sampling::*;
+
+// The `init_submodule` occurs directly in `waveform/quilpy.rs`, as otherwise our linter is unhappy.
+
+/// Register the relevant classes as subclasses of `collections.abc.*`.  Must be called exactly one;
+/// ideally would be called from `init_submodule`, but we can't hook into that, so we expose this as
```
asaites, line 23: Note there's https://github.com/rigetti/rigetti-pyo3/issues/63. I have an (unpublished) branch for it, as I make use of it in my changes to `quil-rs`. There's no need to block this PR for it, but it would address this specific usecase.

quil-rs/src/waveform/sampling/quilpy.rs
```diff
@@ -0,0 +1,421 @@
...
+    Ok(())
+}
+
+// A duplication of [`IqSamples<Complex64>`], but nongeneric so it can be exposed to Python.  It
+// also uses a named argument for the `Samples` constructor because that produces a nicer Python
+// interface and avoids conflicting `__getitem__` definitions.
+/// The result of sampling a waveform, representing a sequence of IQ value samples.
+#[derive(Clone, PartialEq, Debug)]
+#[cfg_attr(feature = "stubs", gen_stub_pyclass_complex_enum)]
+#[pyclass(module = "quil.waveform.sampling", name = "IqSamples", eq, frozen)]
```
asaites, line 40: We might should mention that this isn't marked `sequence` because CPython would then automatically add `__len__` to negative indexes, which may result in overflow errors in some scenarios we otherwise support.

quil-rs/src/waveform/sampling/quilpy.rs
```diff
@@ -0,0 +1,421 @@
...
+        // If there's arithmetic overflow, returns an error.
+        let fix_negative_index = |i: isize| -> PyResult<Option<usize>> {
+            if i < 0 {
+                let i = i.checked_add_unsigned(sample_count).ok_or_else(error)?;
+                Ok(usize::try_from(i).ok())
+            } else {
+                // Positive `isize` fits in `usize`
+                Ok(Some(i as usize))
+            }
+        };
```
asaites:
```suggestion
        let fix_negative_index = |i: isize| -> usize {
            if i >= 0 {
                i as usize
            } else if i.unsigned_abs() >= sample_count {
                0
            } else {
                // start < 0 && |start| < sample_count <= usize::MAX
                // 0 < sample_count - |start| = sample_count + start < sample_count <= usize::MAX
                sample_count - i.unsigned_abs()
            }

        let start = fix_negative_index(start);
        let stop = stop.map(fix_negative_index).unwrap_or(sample_count).min(sample_count);
```

quil-rs/src/waveform/sampling/quilpy.rs
```diff
@@ -0,0 +1,421 @@
...
+
+        let sample_count = self.sample_count();
+
+        // If this is a positive index or a legal negative index, returns `Ok(Some(_))` of the
+        // corresponding positive index.  If this is an illegal negative index, returns `Ok(None)`.
+        // If there's arithmetic overflow, returns an error.
+        let fix_negative_index = |i: isize| -> PyResult<Option<usize>> {
+            if i < 0 {
+                let i = i.checked_add_unsigned(sample_count).ok_or_else(error)?;
```
asaites: I don't think it makes sense to use that error here. In particular, if `sample_count >= isize::MAX`, this would be triggered incorrectly.

quil-rs/src/waveform/sampling/quilpy.rs
```diff
@@ -0,0 +1,421 @@
...
+        IqSamples::from(self.clone())
+            .into_iq_values()
+            .into_pyarray(py)
+    }
+}
+
+// Pymethods that need manual stub generation for overloads (handled below).  Our linter doesn't
+// know about `#[gen_stub(skip)]` and gets concerned when we have a `pymethods` block without any
+// stub generation, so we have to move this to a different impl block and then lie to the linter
+// about generating stubs.
```
asaites: I'm happy to report that the updated version of `pyo3-stub-gen` (used by `rigetti-pyo3`) now supports generating stubs for overloaded methods, and can do so from Python syntax. In addition, `optipy` knows about `#[gen_stub(...)]` attributes and will strip them, so if that was a limitation before, it isn't any longer.

Add to that, `pyo3-stub-gen` now generates the `__all__` module attribute, so we can rely on `stubgen` to tell us about mismatches between the stubs and actual extension module contents (modulo the values assigned to simple enums, but those are reasonable to ignore).

quil-rs/src/waveform/sampling/quilpy.rs
```diff
@@ -0,0 +1,421 @@
...
+        match index {
+            IqSamplesIndex::Int(index) => Ok(get_one(self, index).map(IqSamplesIndexed::One)),
+
+            IqSamplesIndex::Slice(indices) => {
+                let PySliceIndices {
+                    start,
+                    stop,
+                    step,
+                    slicelength,
+                } = indices.indices(self.sample_count() as isize)?;
```
asaites: Can't `self.sample_count() >= isize::MAX`?

quil-rs/src/waveform/sampling/quilpy.rs
```diff
@@ -0,0 +1,421 @@
...
+
+                let sliced_samples = match self {
+                    Self::Flat {
+                        iq,
+                        sample_count: _,
+                    } => vec![*iq; slicelength],
+
+                    Self::Samples { samples } => {
+                        if step > 0 {
+                            let start = start as usize; // Guaranteed by documentation
```
asaites: Can you state which documentation/provide a link?

quil-rs/python/quil/waveform/__init__.pyi
```diff
@@ -0,0 +1,285 @@
...
+    def custom(name: builtins.str, parameters: builtins.dict[builtins.str, _Complex]) -> Waveform: ...
+    @staticmethod
+    def drag_gaussian(*, duration: builtins.float, scale: typing.Optional[_Real] = None, phase: typing.Optional[_Real] = None, detuning: typing.Optional[_Real] = None, fwhm: _Real, t0: _Real, anh: _Real, alpha: _Real) -> Waveform[_Real, _Complex]: ...
+    @staticmethod
+    def erf_square(*, duration: builtins.float, scale: typing.Optional[_Real] = None, phase: typing.Optional[_Real] = None, detuning: typing.Optional[_Real] = None, risetime: _Real, pad_left: builtins.float, pad_right: builtins.float) -> Waveform[_Real, _Complex]: ...
+    def evaluate(self, real: collections.abc.Callable[[_Real], _OtherReal], complex: collections.abc.Callable[[_Complex], _OtherComplex]) -> Waveform[_OtherReal, _OtherComplex]: ...
+    @staticmethod
+    def flat(*, duration: builtins.float, scale: typing.Optional[_Real] = None, phase: typing.Optional[_Real] = None, detuning: typing.Optional[_Real] = None, iq: _Complex) -> Waveform[_Real, _Complex]: ...
+    @staticmethod
+    def from_quil(invocation: WaveformInvocation) -> Waveform[quil.expression.Expression]: ...
```
asaites: This line changed for me when I ran `cargo make test-all` locally on this branch.

asaites (commented):
I'm still working through some of this, but overall things are looking pretty good. I have a few general comments at the moment and figured I'd submit them now, but I'll submit more soon.

quil-rs/src/waveform/builtin/macros.rs
```diff
@@ -0,0 +1,981 @@
...
+                   )?
+                   .into())
+            }
+
+            pub(crate) fn __eq__<'py>(
+                &self,
+                #[gen_stub(override_type(type_repr = "builtins.object", imports = ("builtins")))]
+                other: Bound<'py, PyAny>,
+            ) -> bool {
+                if let Ok(Self) = other.extract() {
```
asaites: It'd be more performant to use


```suggestion
                if other.is_instance_of::<Self() {
```

When `extract` fails, it requires generation of a `PyErr`, which is costly. This usually comes up with `cast` chains (so, when you'd actually want to use the result, not just type check it), so [there's a little section in of the PyO3 guide that calls out the performance hit](https://pyo3.rs/v0.29.0/performance.html#extract-versus-cast). Of course, I'm assuming there's nothing special here about the `FromPyObject` implementation that makes `extract` the right choice, but if there is, it'd likely be best pulled into a separate method returning a non-`PyErr` error so that this method could call it.

Also note there's [`is_exact_instance_of`](https://docs.rs/pyo3/0.29.0/pyo3/types/trait.PyAnyMethods.html#tymethod.is_exact_instance). These both [delegate to trait methods](https://docs.rs/pyo3/0.29.0/src/pyo3/types/any.rs.html#1382).

quil-rs/src/waveform/builtin/macros.rs
```diff
@@ -0,0 +1,981 @@
...
+
+            $(python_get_set!($name, $field, $ty);)*
+
+            impl super::$name<Reference<'_, Pythonic>> {
+                pub(crate) fn py_eq<'py>(
+                    &self,
+                    py: Python<'py>,
+                    other: Bound<'py, PyAny>,
+                ) -> PyResult<bool> {
+                    if let Ok($name(other)) = other.extract() {
```
asaites: The extract vs cast performance aspect I mentioned in another comment would apply here.

quil-rs/src/waveform/builtin/macros.rs
```diff
@@ -0,0 +1,981 @@
...
+                    py: Python<'py>,
+                    duration: f64,
+                    #[gen_stub(override_type(
+                        type_repr = "typing.Optional[_Real]",
+                        imports = ("typing"))
+                    )]
+                    #[gen_stub(override_type(
+                        type_repr = "typing.Optional[_Real]",
+                        imports = ("typing"))
+                    )]
```
asaites: It looks like these were duplicated? I'm surprised it doesn't result in an error.

quil-rs/src/waveform/builtin/quilpy.rs
```diff
@@ -0,0 +1,514 @@
...
+            Flat[_Real, _Complex] | \
+            Gaussian[_Real, _Complex] | \
+            DragGaussian[_Real, _Complex] | \
+            ErfSquare[_Real, _Complex] | \
+            HermiteGaussian[_Real, _Complex] | \
+            BoxcarKernel\
+        ",))]
+        waveform: Bound<'py, PyAny>,
+    ) -> PyResult<Self> {
+        if let Ok(PyFlat(flat)) = waveform.extract() {
```
asaites: This chain should certainly use `cast` instead.

quil-rs/src/waveform/builtin.rs
```diff
@@ -0,0 +1,1469 @@
...
+/// Encapsulates the common pattern for generating a sequence of samples for a waveform:
+///
+/// 1. Generate information from some generally-used parameters.
+/// 2. Use that information to generate the samples.
+/// 3. Rescale those samples.
+/// 4. Apply the phase adjustment and detuning to those samples.
+///
+/// These samples are generated from a sequence of time steps ranging evenly over the half-open
+/// interval [0,1).
+fn build_samples_and_adjust_for_common_parameters<I: IntoIterator<Item = Complex64>>(
```
asaites, line 1091: These function names are getting a bit unwieldy 😓

quil-rs/Cargo.toml
```diff
@@ -3,7 +3,7 @@ name = "quil-rs"
 description = "Rust tooling for Quil (Quantum Instruction Language)"
 version = "0.36.0"
 edition = "2021"
-rust-version = "1.83"
+rust-version = "1.93"
```
asaites, line 6: Apparently versions 1.87-1.96 [have a miscompilation bug](https://github.com/rust-lang/rust/issues/159035) (in LLVM).

quil-rs/Cargo.toml
```diff
@@ -33,14 +33,17 @@ thiserror = "1.0.56"
 indexmap = "2.10.0"
 statrs = "0.16.0"
 paste = "1.0.15"
+numpy = { version = "0.27.1", optional = true }
```
asaites: Shouldn't this stay below line 40?

quil-rs/src/waveform/quilpy.rs
```diff
@@ -1,236 +1,482 @@
...
+        ],
+        errors: [
+            errors::SamplingError,
+        ],
+    }
+}
 
-use super::templates::*;
-use crate::quilpy::errors::ValueError;
+macro_rules! just_for_the_linter {
```
asaites: You could probably just put them in the entry script itself, sort of like [how the SDK removes some types](https://github.com/rigetti/qcs-sdk-rust/blob/main/crates/lib/scripts/lint-bindings.py#L34) that would otherwise report an issue.

asaites (commented):
Thank you for the excellent additional documentation, and the hard work you've put in to making the Python interface (and stubs!) fit together better. Overall, I think this looks pretty good, but I am a little skeptical of some of the added complexity around bits of the Python pieces. I don't see anything there that I think blocks the PR, though. We might need some cleanup for it when we upgrade to the newer versions of the stub generator and PyO3.

I've made a few comments on some aspects that I think should improve for the sake of "that's a known performance hit, and the more performant version is just as clean". Other than that, I saw a couple minor issues that should be addressed before pulling it in, but I don't think it needs to languish much longer.

quil-rs/src/waveform/sampling/quilpy.rs
antalsz: Oh, thank you for the reminder, I forgot to mention – I actually *did* try marking it `sequence`, but then I couldn't reconcile the behavior I was seeing (no added `__len__`) with the documented behavior (`__len__`).  This was one of the last things I was doing so I just left it at the time – the class works fine either way.  But yes, I'll add this in comments.

quil-rs/src/waveform/sampling/quilpy.rs
antalsz: Ah, no it can't – Rust (and I believe also Python) guarantees that any single allocated object has size at most `isize::MAX` (see [here](https://doc.rust-lang.org/stable/reference/types/numeric.html#r-type.numeric.int.size.isize)).  This comes from LLVM and I think is a pretty standard limitation.  I'll add a comment.

quil-rs/src/waveform/sampling/quilpy.rs
antalsz: I was about to say "as above, `sample_count` must be `< isize::MAX`, but this isn't quite true in the `Flat` case!  Thankfully, this is gone now :-)

quil-rs/src/waveform/sampling/quilpy.rs
antalsz: Thank you for the suggestion!  The simpler behavior you had here allowed me to figure out how to express this much more directly as clamping to `[0, sample_count]`.

quil-rs/src/waveform/sampling/quilpy.rs
antalsz: Hm, this *doesn't* work right in the `Flat` case, though.  I don't think it would be unreasonable to not support `Flat` waveforms of that size, but I'll see if I can simplify this.

antalsz: PyO3 doesn't seem to offer enough support for this (https://pyo3.rs/main/doc/pyo3/types/struct.pyslice), so I simply added an informative error

quil-rs/src/waveform/sampling/quilpy.rs
antalsz: Done, thanks.

quil-rs/src/waveform/builtin/macros.rs
antalsz: Ooh, this is nice, thank you!  It looks like `cast` requires working with native types (since you get a `Bound` back) so is the thinking here to use `is_instance_of` + `extract` to avoid the allocation hit?  (And of course, here in particular you don't need `extract` at all, but in general.)

quil-rs/src/waveform/builtin/macros.rs
antalsz: Huh, same.

quil-rs/Cargo.toml
antalsz: You're right, it should.  Not sure why I moved it.

quil-rs/Cargo.toml
antalsz: Oof, good to know.  It sounds like it's not a serious enough miscompilation bug to avoid setting our MSRV for, though, per [this comment](https://github.com/rust-lang/rust/issues/159035#issuecomment-4933829910):

- It's only on x86-64 (we should let e.g. macOS users build this library on any version)
- It can only cause harm by unexpectedly crashing (there's no security or misbehavior risk)
- It's hard to observe on Rust versions <1.97, and 1.97 got a patch release to fix this.

quil-rs/src/waveform/quilpy.rs
antalsz: The clearly better choice was just to update your `_python_waveforms` function in that file to understand the new macro.  I didn't realize that feature was available!

quil-rs/python/quil/waveform/__init__.pyi
antalsz: Same, I think I just missed committing it after some updates.

quil-rs/src/waveform/sampling/quilpy.rs
antalsz: The `gen_stub` limitation is in the linter, not `optipy` – `#[gen_stub(skip)]` suppresses generating a stub, but the linter is only aware of `gen_stub` containing "override" or "builder_field_attr" (`elif "gen_stub" in line and not ("override_" in line or "builder_field_attr" in line):`).  However, having reread the docs since writing this, I think I can just solve this by moving the overload submission before the methods instead of after them.

antalsz:
Commit 8c03dab is not a response to review – when my internet died, I needed to build the docs locally, and this revealed a number of warnings from `cargo doc` that I fixed in that commit.

quil-rs/src/waveform/builtin.rs
antalsz: I expect a physicist could name them better, but I'm stuck with my lack of understanding and clunky names 🙃

quil-rs/src/bin/stub_gen.rs
antalsz: The use case for explicit `self` is not just for overloads – I use it here to generate, for instance,

```
def iq_values_at_sample_rate(
    self: BuiltinWaveform[builtins.float, builtins.complex],
    common: CommonBuiltinParameters[builtins.float, __T],
    sample_rate: builtins.float,
) -> IqSamples:
    ...
```

That method isn't meaningful to call unless the `BuiltinWaveform` is fully concrete like that.  I use a lot of `pyo3-stub-gen`'s overload features, but this is orthogonal.

antalsz (commented):
Thanks so much for all the feedback, @asaites!  I think I've responded to everything.

I know the `cargo deny` job is failing, but that's fixed in !504 and I was hoping to just land that one first, since it's smaller.

quil-rs/src/waveform/sampling/quilpy.rs
asaites: Blah, you're right. I added an [issue](https://github.com/rigetti/rigetti-pyo3/issues/90) for it in the `rigetti-pyo3` repo, as it should get fixed up along with a few other related problems that upgrading `pyo3_stub_gen` lead to.

quil-rs/src/waveform/sampling/quilpy.rs
asaites: I knew pieces of this at one time, but apparently had forgotten. Thanks for the double-check.

quil-rs/Cargo.toml
asaites: Sounds good. Should we step to 1.97 then instead?

quil-rs/src/bin/stub_gen.rs
asaites: Honestly, I didn't think `mypy` would be happy with that, but it doesn't seem to complain, so neither shall I!

asaites (approved):

quil-rs/Cargo.toml
asaites: Nevermind, I understand your point: users sensitive to it can upgrade their toolchain as appropriate. We're probably fine to leave this at 1.93.

PR #486, opened 2025-10-20
447 defgate as sequence bak

erichulburd:
Superseded by https://github.com/rigetti/quil-rs/pull/460, but here for convenience of reviewing the diff.


asaites:
Closing, now that #460 is merged.

PR #497, opened 2026-02-25
feat: normalize behavior and usage of `Instruction::get_qubits`

### LINKED ISSUE #498: The qubits "used" by instructions and programs aren't calculated correctly
1. `Instruction::get_qubits` and `Instruction::get_qubits_mut` miss the qubits contained in certain instructions.
2. `Program::add_instruction` should also only track the qubits used by body instructions. 
    - For example, it is possible to extract qubits used in `CalibrationDefinition`s, but the qubits in a definition should not be considered "used" by the `Program`.

mingyoungjeng:
- `get_qubits` and `get_qubits_mut` now properly returns the qubits used for previously-unaccounted-for Instructions
- `Program::add_instruction` only caches _body_ instructions in `Program::used_qubits`.
- Generalizes `Instruction::get_qubits` and `Instruction::get_qubits_mut` functionality into the `InstructionHandler` trait

Closes #498


quil-rs/src/instruction/circuit.rs
```diff
@@ -1,8 +1,11 @@
 #[cfg(feature = "stubs")]
 use pyo3_stub_gen::derive::gen_stub_pyclass;
```
asaites: You've got a compilation error because this needs to be:

```suggestion
use pyo3_stub_gen::derive::{gen_stub_pyclass, gen_stub_pymethods};
```

You can test these ahead of time by running 

```sh
cargo make test-all
```

Along the same lines, you need to regenerate the type stubs and check them, since you've made changes to the Python API. You can do that with:

```sh
cargo make generate-stubs
```

Since that requires the `stubs` feature, in this instance, it would have caught the issue.

quil-rs/src/instruction/circuit.rs
```diff
@@ -12,24 +15,42 @@ use super::Instruction;
...
-            parameters: Vec<String>,
-            qubit_variables: Vec<String>,
-            instructions: Vec<Instruction>,
-        );
+#[cfg_attr(feature = "stubs", gen_stub_pymethods)]
+#[cfg_attr(feature = "python", pyo3::pymethods)]
+#[cfg_attr(not(feature = "python"), strip_pyo3)]
+impl CircuitDefinition {
+    #[new]
+    pub fn new(
```
asaites, line 41: Since you're not able to use `pickleable_new` here[^1], you should implement `__getnewargs__` manually so that `CircuitDefinition`s (and its subclasses) remain pickleable. 

Since doing so is purely for Python users, our convention is to put it in a `quilpy` module. In this case, the appropriate place is `src/instructions/quilpy.rs`, likely around line 460. 

Here's one possible implementation (the important thing is that its arguments are compatible with the method marked `#[new]`):

```rust
#[cfg_attr(feature = "stubs", gen_stub_pymethods)]
#[pymethods]
impl CircuitDefinition {
    fn __getnewargs__<'py>(&self, py: pyo3::Python<'py>) -> PyResult<Bound<'py, PyTuple>> {
        (
            self.name.clone(),
            self.parameters.clone(),
            self.qubits.clone(),
            self.instructions.clone(),
        ).into_pyobject(py)
    }
}
```

[^1]: I'd still like to look into improving this macro to make it useful across a wider range of cases, and likely move it to `rigetti-pyo3`.

quil-rs/src/instruction/circuit.rs
```diff
@@ -12,24 +15,42 @@ use super::Instruction;
...
     pub name: String,
+
+    #[pyo3(get, set)]
     pub parameters: Vec<String>,
-    // These cannot be fixed qubits and thus are not typed as `Qubit`
-    pub qubit_variables: Vec<String>,
+
+    // These cannot be fixed qubits and thus cannot be accessed directly
+    // outside of this crate
+    #[pyo3(get, set)]
```
asaites: You probably don't want `set`, if these must be `Qubit::Variable`, as otherwise Python users can set the list to other kinds of `Qubit`s.

To that end, I think you should at least document that this must be `Qubit::Variable` explicitly. You could enforce that with a newtype that wraps `String` and serves as the type of both `Qubit::Variable` and the type of this `Vec`, but its not clear to me that's any better than just using `String` itself, which you're trying to change. That leads me to the obvious question: 

Why should this type be `Vec<Qubit>` if it really is restricted to only the `Variable` variant?

quil-rs/src/instruction/gate.rs
```diff
@@ -1056,8 +1056,8 @@ impl<'a, T: AsRef<str>> GateSignature<'a, T> {
     }
 }
 
-#[cfg(feature = "python")]
 impl<'a, T: AsRef<str>> GateSignature<'a, T> {
```
asaites, line 1075: Nit: I think you should just pull the `try_new` method into a separate `impl` block instead of repeating the `#[cfg(feature = "python")]`s.

quil-rs/src/instruction/mod.rs
```diff
@@ -685,81 +685,220 @@ impl Instruction {
     }
 
     /// Return immutable references to the [`Qubit`]s contained within an instruction
-    #[allow(dead_code)]
-    pub fn get_qubits(&self) -> Vec<&Qubit> {
+    pub(crate) fn default_qubits<'a, T>(&'a self) -> T
```
asaites, line 688: As I look at this implementation, I am starting to think we should have an actual `QubitIterator` that can encapsulate this logic without all the `collect`s. I'd like to see some performance details on this, as perhaps it wouldn't matter anyway, but at the very least, the implementation gets pretty messy here. I could imagine it being more of a performance/usability improvement for the Python API, but again, we'd need some profiling data to really say.

quil-rs/src/program/mod.rs
```diff
@@ -790,7 +765,7 @@ impl Program {
         self.used_qubits = self
             .to_instructions()
             .iter()
-            .flat_map(|instruction| instruction.get_qubits().into_iter().cloned())
+            .flat_map(|instruction| DefaultHandler.get_qubits(instruction).into_iter().cloned())
```
asaites, line 780: This can result in the same `used_qubits` as the old `add_instruction` implementation: if you call `resolve_placeholders_with_custom_resolvers()` (which calls this method) on a `Program` with `DEFCAL` instructions, the `Variable`s end up in the cache. That was _also_ true before this PR, but it represents an inconsistency between the way `add_instruction` and this method work within this PR.

I'll also note that `DEFCIRCUIT` qubits will also make their way into the cache after calling this method, but that _wasn't_ true before the PR.

asaites (commented):
Overall I think this is on the right track, but needs a little adjustment. I think there's an inconsistency in the qubit caching when adding an instruction vs rebuilding the cache directly.

quil-rs/src/instruction/circuit.rs
mingyoungjeng: The main reason I had to make it `Vec<Qubit>` instead of the prior `Vec<String>` is because I needed to extract `&Qubit`s from the circuit definition.

I could try to make something like this work?
```rust
enum QubitRef<'a> {
    Fixed(u64),
    Placeholder(QubitPlaceholder),
    Variable(&str),
}
```

quil-rs/src/instruction/mod.rs
mingyoungjeng: Alternatively, I could just force it to be `HashSet` or `Vec`. It was `Vec` previously, but it makes more sense (imo) to be `HashSet` -- all the consumers either didn't care or would have preferred `HashSet`.

quil-rs/src/program/mod.rs
asaites: @mingyoungjeng Are you planning to address this specific concern, or is there a reason it doesn't apply the way I think it does? This PR has been open a while now; is there more support you need from me for now?

mingyoungjeng: Apologies, I addressed this a while back, see https://github.com/rigetti/quil-rs/blob/e180e09bd38b3dc8b3dc9401a83f1f937571a03e/quil-rs/src/program/mod.rs#L764, where the following lines were added.
```rust
// This ensures parity with the behavior of `Self::add_instruction`
.filter(|instruction| match instruction {
    Instruction::CalibrationDefinition(_)
    | Instruction::CircuitDefinition(_)
    | Instruction::FrameDefinition(_)
    | Instruction::Declaration(_)
    | Instruction::GateDefinition(_)
    | Instruction::MeasureCalibrationDefinition(_)
    | Instruction::WaveformDefinition(_) => false,
    Instruction::Pragma(pragma) if pragma.name == RESERVED_PRAGMA_EXTERN => false,
    _ => true,
})
```
However, I forgot to update this thread. To the best of my knowledge, things should be good to go, perhaps except form linting failures from new clippy.

PR #503, opened 2026-05-28
feat!: support bitshifts prior to pyo3 upgrade

### LINKED ISSUE #502: Release bitshifts on earlier version of pyo3
We have a motivated use case for bitshifts that we would like to make use of before migrating to 0.36.0. This will require porting https://github.com/rigetti/quil-rs/pull/493 to v0.35.0 and releasing a patch.

erichulburd:
This essentially cherry picks much of https://github.com/rigetti/quil-rs/pull/493 onto 9db8cb4cfe75e67a4fa572e96e86ae8e59d0bdfc so that we can release it as 0.18.0 in order to get bitshift support before migrating to the most recent breaking version of quil.

Closes #502 .


quil-py/uv.lock
asaites, line 1: Since this project is currently using `poetry`, I don't think it makes sense to include this for now. We have talked about (and still have interest in) moving to `uv`, but it's not something we're doing right away, and not something I think makes sense to do as part of this release.

quil-py/quil/expression/__init__.pyi
asaites, line 1: I don't have any issue with these updates to the stubs, but for what it's worth, the newest `quil-rs` releases generate stubs automatically, so these aren't necessarily going to be consistent with that.

asaites (commented):
From what I can see, I think a release like this should be fine, though I think we may want to just keep it as its own branch/rename it something like `v0.18`, rather than actually merging this into the `v0.17` branch (or we could create a `v0.18` branch from the same point and then merge this into that). Mostly that's just for the benefit of historical reference, as I _don't_ recommend rebasing all of `main` with this in its history.

In any case, I'm not 100% sure the automation is set up for publishing a new version in this manner, so we might need to do so manually. We can at least give it a go first and see if there's a gap.

PR #509, opened 2026-07-24
fix!: update to PyO3 0.29

BatmanAoD:


quil-rs/src/program/quilpy.rs
```diff
@@ -418,7 +418,9 @@ impl Program {
     }
 
     pub fn __setstate__(&mut self, state: &Bound<'_, PyBytes>) -> PyResult<()> {
-        *self = Self::from_str(std::str::from_utf8(state.as_bytes())?)?;
+        let quil = std::str::from_utf8(state.as_bytes())
+            .map_err(|e| pyo3::exceptions::PyValueError::new_err(e.to_string()))?;
+        *self = Self::from_str(quil)?;
```
BatmanAoD, line 423: Claude says:

>a Utf8Error no longer auto-converts to PyErr in quilpy.rs:420 (now explicit map_err)

quil-rs/src/waveform/quilpy.rs
```diff
@@ -220,7 +220,9 @@ pub(super) mod explicit_stubs {
                 ],
                 r#return: || TypeInfo {
                     name: format!("{}[_OtherReal, _OtherComplex]", T::NAME),
+                    source_module: None,
                     import: Default::default(),
+                    type_refs: Default::default(),
```
Shadow53, line 225: I'm unsure at a glance if `source_module: None` is correct, but I'm not sure where to look for docs either. I'm fine with this if you are confident, but confirmation from @asaites would be nice.

Shadow53 (approved):

BatmanAoD:
womp womp, fixing the CI failures is nontrivial:

> 1. Clippy failure (Run Checks) — mechanical, low-risk: ~99 #[pyclass(...)] sites trigger a pyo3 0.29 deprecation warning (auto-FromPyObject-via-Clone becoming opt-in) promoted to an error by -D warnings, plus one PyTypeInfo::NAME deprecation at waveform/quilpy.rs:222. Fixable, but it's a real per-type choice: #[pyclass(from_py_object)] (keep today's behavior) vs #[pyclass(skip_from_py_object)] (adopt pyo3's new default, which quietly removes the ability to .extract() these types out of a Python object built via their Clone impl). rigetti-pyo3 chose skip_from_py_object for its own internal example type during its 0.28 bump — but quil-rs's ~99 types are public API, so blanket-applying one choice could silently change behavior for downstream Python users.
> 
> 2. Stub-gen crash — this is the one blocking everything else (Run Tests on all 12 OS/Python combos, Check quil API): pyo3-stub-gen 0.23.0 added a hard validation that any #[pyclass(module = "...")] value must live under the crate's configured module-name (quil._quil, from pyproject.toml:26, mixed maturin layout). But quil-rs's classes declare friendly public paths — module = "quil.program", "quil.waveform", etc. — not "quil._quil.program". That mismatch was silently tolerated by pyo3-stub-gen 0.17.2 (no such check existed) but is now a hard error: Module 'quil' has PyO3 items ... but is not under the PyO3 module path 'quil._quil'.
> 
> That second one isn't a mechanical fix — it's a real packaging/API-layout decision (e.g. rename every module = "quil.xxx" to "quil._quil.xxx", which changes the runtime __module__/repr of every public class, vs. some other reconciliation between the friendly quil.* import surface and the compiled quil._quil extension). I don't want to guess at that. How do you want to handle it — and do you have a preference on the from_py_object/skip_from_py_object question above?

I'll wait for @asaites to weigh in, because I have no context on why we're doing this `._quil` thing.

BatmanAoD:
superseded by https://github.com/rigetti/quil-rs/pull/513

PR #513, opened 2026-07-28
feat!: upgrade pyo3 and pyo3_stub_gen

### LINKED ISSUE #512: Update to PyO3 v0.29 and others
`rigetti-pyo3` is now on PyO3 v0.29 with additional updates for related dependencies. This repo should update as well.

Likely before it can, [this `rigetti-pyo3` issue](https://github.com/rigetti/rigetti-pyo3/issues/90) should be resolved.

Note: [this PR](https://github.com/rigetti/quil-rs/pull/509) already addresses the bulk of the work within `quil-rs`. Mostly, the remaining work relates to how the updated `pyo3_stub_gen` generates the package layout and

asaites:
Changes needed for upgraded `pyo3` and `pyo3_stub_gen`:

- Added `from_py_object` to all Cloneable `#[pyclass]`es.
  - This change isn't strictly necessary, but without it, we'll get a ton a deprecation warnings.
  - This was PyO3's default behavior before this change,
    but we'll eventually want to apply `skip_from_py_object` to some of our classes
    and implement custom version of `FromPyObject` with more care.
- Fixed some generated type names that break `mypy` due to its mishandling of `@property` names
  that overlap module names.
  - For example, `quil._quil.Program` has an `instructions` property,
    so [`mypy` will choke](https://github.com/python/mypy/issues/4146)
    if we try to use `instructions.Instruction` within the same class.
  - As a work-around, we can force `pyo3_stub_gen` to use e.g. `_quil.instructions` instead of `instructions`
    whenever referencing that module within the `Program` class.
    I've opened [an issue](https://github.com/Jij-Inc/pyo3-stub-gen/issues/493) there
    so hopefully in the future we won't have to resort to this work-around.
- Updated the `module = quil.<submod>` to `module = quil._quil.<submod>`
  - Technically, this was always the correct module path,
    but now that `pyo3_stub_gen` handles exporting the private module contents
    to the public module namespace, we need to use the private module path name.
- Added `reexport_module_members!` declarations to the `quilpy` module to export the public `quil` API.
- Fixed the linter script to work with the updated module names.

General improvements to the `quil` type stubs:

- Improved consistency around usage of `builtins` and `typing` modules in stubs.
- Moved `quilpy`-related functionality out of the `program` module and into its `quilpy` module.
- Simplified type stub generation for `Waveform`s in macro usage.
- Moved `quil`'s module docstring to the `quilpy` module, since the entire `__init__.pyi` is now generated.

Other updates needed to pass our CI checks:

- Updated `itertools` to `0.14.0` to resolve a `cargo-deny` issue.
- Updated `deny.toml` to remove outdated items and add exception for `rustpython-parser`.

Closes: #512


quil-rs/src/program/quilpy.rs
```diff
@@ -417,14 +536,119 @@ impl Program {
...
+        from quil._quil import instructions
+
+        class Program:
+            @overload
+            def wrap_in_loop(
+                self,
+                loop_count_reference: _quil.instructions.MemoryReference,
+                start_target: _quil.instructions.Target,
+                iterations: builtins.int,
+            ):
```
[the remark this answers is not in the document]
asaites: Fixed.

quil-rs/src/waveform/quilpy.rs
```diff
@@ -142,136 +143,39 @@ macro_rules! define_py_evaluate {
...
-                pyo3::inventory::submit! {
-                    explicit_stubs::py_evaluate_method_info::<[<Py$name>]>()
-                }
             }
         )*
     };
 }
 
-#[cfg(feature = "stubs")]
-pub(super) mod explicit_stubs {
```
antalsz, line 169: I take it this module getting removed means that `pyo3-stub-gen` is now capable of doing these things by itself?

antalsz (approved):
I just reviewed the waveform stuff and things look fine there 👍

quil-rs/src/instruction/quilpy.rs
```diff
@@ -745,11 +760,39 @@ impl PauliGate {
     }
 }
 
+#[cfg_attr(not(feature = "stubs"), optipy::strip_pyo3(only_stubs))]
+#[cfg_attr(feature = "stubs", gen_stub_pymethods)]
+#[pymethods]
+impl PauliTerm {
```
BatmanAoD, line 766: Is this related to the PyO3 upgrade somehow, or separate?

quil-rs/src/program/mod.rs
```diff
@@ -350,7 +360,15 @@ impl Program {
     )]
     pub fn wrap_in_loop(
         &self,
+        #[gen_stub(override_type(
+            type_repr = "_quil.instructions.MemoryReference",
+            imports = ("quil._quil"),
+        ))]
```
BatmanAoD, line 366: Jeeze, this is noisy...

quil-rs/src/program/quilpy.rs
```diff
@@ -401,6 +461,65 @@ impl Program {
         Ok(self.to_unitary(n_qubits)?.to_pyarray(py))
     }
 
+    #[gen_stub(skip)]
+    #[pyo3(
+        name = "wrap_in_loop",
+        signature = (loop_count_reference, start_target, end_target = None, iterations = None),
+    )]
+    fn backwards_compatible_py_wrap_in_loop(
```
BatmanAoD, line 469: Why are there multiple `impl Program`s, and why move this method from one to the other?

quil-rs/src/program/quilpy.rs
```diff
@@ -930,38 +1173,7 @@ struct PyTargetResolver(Py<PyFunction>);
 mod stubs {
     use pyo3_stub_gen::{PyStubType, TypeInfo};
 
-    #[allow(clippy::wildcard_imports)]
-    use super::*;
-
-    /// Create a `Callable[[A], R]` stub.
-    fn callable_of<A: PyStubType, R: PyStubType>() -> TypeInfo {
```
BatmanAoD, line 937: No longer needed?

quil-rs/src/quilpy/mod.rs
```diff
@@ -84,3 +80,33 @@ pub(crate) fn py_cast_and_clone<'a, 'py, T: PyClass + FromPyObject<'a, 'py> + Cl
...
+    reexport_module_members!("quil.waveform.sampling" from "quil._quil.waveform.sampling");
+
+    module_doc!(
+        "quil._quil",
+        r#"
+        The `quil` package provides tools for constructing, manipulating,
+        parsing, and printing [Quil](https://github.com/quil-lang/quil) programs.
+
+        ⚠️ This package is still in early development
+        and breaking changes should be expected between minor versions.
```
BatmanAoD, line 106: Is this notice still true?

deny.toml
```diff
@@ -91,6 +87,8 @@ unknown-registry = "deny"
 # Lint level for what to happen when a crate from a git repository that is not
 # in the allow list is encountered
 unknown-git = "deny"
+# Don't warn if we're not (currently) using a crate from one of the allowed sources.
+unused-allowed-source = "allow"
```
BatmanAoD, line 91: Ah, I should add that to the base config

BatmanAoD (approved):

quil-rs/src/waveform/quilpy.rs
asaites: Sort of: the `evaluate` stubs are handled by the `gen_stub` annotations within the macro, but for the moment, I've updated the `stubtest-allowlist` to ignore the `__class_getitem__` stubs.

For the former, my motivation was two-fold:
1. PyO3 is deprecating the `T::NAME` value you were using, and the direct replacement would require runtime access to a `Python` token to access the Python type name. I modified it to use the `$name` passed to the macro, though technically neither of these are certain to be the correct type name[^1]. 
2. The attribute syntax (using their derive macros) seems much more stable than trying to write the full submission. I also think it's more readable/maintainable.

To those points, after giving this more thought, maybe a better solution is to have `evaluate` return a simple wrapper around `Self`[^2]:

```rust
/// A simple wrapper used to provide the correct type annotation for the result of `evaluate`.
#[derive(pyo3::IntoPyObject, pyo3::IntoPyObjectRef)]
struct PyEvaluatedWaveform<T>(T);

#[cfg(feature = "stubs")]
impl<T: pyo3_stub_gen::PyStubType> pyo3_stub_gen::PyStubType for PyEvaluatedWaveform<T> {
    fn type_output() -> pyo3_stub_gen::TypeInfo {
        let mut info = T::type_output();
        info.name = format!("{}[_OtherReal, _OtherComplex]", info.name);
        info
    }
}

// In the macro, the return type then becomes
// PyResult<PyEvaluatedWaveform<Self>>
```

The  `gen_stub(override_type(...))` attributes added to the `real` and `complex` parameters could also be replaced with wrapper types, but the wrapper would either need to be around `Bound<'py, PyAny>` (which has additional overhead to increment the `refcount`s of the inputs) or would need to be around `Borrowed<'a, 'py, PyAny>` with a manual implementation of `Deref<Target = Bound<'py, PyAny>>` that forwards to the existing one.

----

For the latter change, this sent me down a ridiculous rabbit hole of trying to determine what value a custom implementation could possibly have. In the end, my conclusion is that [since `mypy` doesn't care about the actual  `__class_getitem__` definition](https://github.com/python/mypy/issues/11501), it makes more sense to just tell `stubtest` to stop complaining rather than maintain extra code just to satisfy it.

If we really want to annotate it in the stubs, [we could write an actual implementation](https://pyo3.rs/v0.29.1/python-typing-hints.html#advanced-users) and add the annotations using attributes. One advantage to that is we wouldn't need to explicitly mark those classes as `#[pyclass(generic)]`. Another is we _could_ force users to use exactly two type parameters:

```rust
#[pyo3::pymethods]
impl PyWaveform {
    #[classmethod]
    #[pyo3(signature = (key, /))]
    pub fn __class_getitem__<'py>(cls: &Bound<'py, PyType>, key: &Bound<'py, PyAny>) -> PyResult<Bound<'py, PyGenericAlias>> {
        let (real, complex) = key.extract::<(Bound<'_, PyType>, Bound<'_, PyType>)>()?;
        PyGenericAlias::new(cls.py(), cls.as_any(), &(real, complex).into_bound_py_any(cls.py())?)
    }
}
```

That means if you write `Waveform[float, complex]` it'll give back an appropriate `types.GenericAlias`, but if you write `Waveform[int]` or `Waveform[int, int, int]`, you'll get an error with `expected tuple of length 2`. Using the default `generic` implementation that PyO3 provides allows any number of values to parameterize it. Note that it would only be caught at runtime. It might be possible to take advantage of the returned `cls` to return a _different_ class depending on the `key`, but [the docs say](https://docs.python.org/3/reference/datamodel.html#object.__class_getitem__) don't do that:

> Using __class_getitem__() on any class for purposes other than type hinting is discouraged.

[^1]: That's a bit of a bummer, but it's currently the case that the names match, and if it changes in the future, it'll very likely be caught by `stubtest`.
[^2]: This would also reduces the use of `paste!` to just the `impl [<Py$name>]` line, and if we just pass the actual type names directly to the macro, we could choose to eliminate its use of `paste!` entirely.

quil-rs/src/instruction/quilpy.rs
asaites: It's to resolve [this issue](https://github.com/Jij-Inc/pyo3-stub-gen/issues/493) that appeared after upgrading `pyo3_stub_gen` -- `PauliGate` has an `expression` field, so within its body, we have to force references to the module `quil._quil.expression` to be fully qualified. Unfortunately, that's very verbose.

quil-rs/src/program/mod.rs
asaites: 💯 I completely agree.

quil-rs/src/program/quilpy.rs
asaites: There are actually three! Their distinction is whether a method is Rust-only, Python-only, or shared:

- The Rust-only methods land in the ordinary `src/program/mod.rs` `impl Program`.
- The shared methods are also in `src/program/mod.rs`, but in an `impl Program` block that PyO3 processes using the `#[pymethods]` macro.
- The Python-only methods are in this file, `src/program/quilpy.rs`. Splitting Python-specific stuff into separate modules came about during the Great Crate Merge; a reviewer requested it to help with code organization. It also reduces some of the `#[cfg_attr(...)]` noise, since the whole `quilpy` module is selectively included/excluded.

Anyway, I moved these items from `src/program/mod.rs` to `src/program/quilpy.rs` to maintain that separation. I hadn't noticed at the time of the original PR that added these items that they were Python-only. This move could be split into another PR, but since I had to modify parts of the code anyway (to fix some elements of the generated stub annotations), it seemed reasonable to move it. I'd be fine splitting it out, if you'd prefer.

quil-rs/src/quilpy/mod.rs
asaites: I'd say so, though arguably being `v0.x.y` says the same thing.

quil-rs/src/program/quilpy.rs
asaites: Correct. The implementation here would need to be updated to work with the updated `pyo3_stub_gen`, but in the end I _had_ to annotate the only uses of this method to work around the module name clash. If I reintroduce it, I'll likely recommend putting it in `rigetti-pyo3` or (better yet) trying to upstream it to `pyo3_stub_gen` itself (they have similar constructors for `dict_of`, `set_of`, and the like).

quil-rs/src/program/quilpy.rs
BatmanAoD: No, keeping that three-part organization makes sense now that you've explained it.

PR #517, opened 2026-08-05
feat: add `RaisedCosine` builtin waveform

### LINKED ISSUE #516: Add a waveform template for RaisedCosineEnvelope

mingyoungjeng:
Closes #516


quil-rs/src/waveform/snapshots/quil_rs__waveform__builtin__tests__RaisedCosine__rolloff_0_0_pad_left_0_0_pad_right_0_0___CommonBuiltinParameters__duration_0_0001_scale_Some(1_0)_phase_Some(Cycles(0_0))_detuning_Some(0_0)___plot.snap
jselig-rigetti, line 1: Should `RaisedCosine` always start and end on 0 and have a value at some point which matches the `scale`?

quil-rs/src/waveform/snapshots/quil_rs__waveform__builtin__tests__RaisedCosine__rolloff_4e-5_pad_left_0_0_pad_right_0_0___CommonBuiltinParameters__duration_0_0001_scale_Some(1_0)_phase_Some(Cycles(0_0))_detuning_Some(0_0)___plot.snap
jselig-rigetti, line 1: Is this waveform expected? Shouldn't this be symmetrical?

quil-rs/src/waveform/snapshots/quil_rs__waveform__builtin__tests__RaisedCosine__rolloff_0_0_pad_left_0_0_pad_right_0_0___CommonBuiltinParameters__duration_0_0001_scale_Some(1_0)_phase_Some(Cycles(0_0))_detuning_Some(0_0)___plot.snap
mingyoungjeng: I am actually not sure what the behavior at $t=0$ should be. The behavior here matches the behavior of the reference, but I don't 100% agree that it is correct. Especially since the discrete samples are in the half-open interval $[0, duration)$, meaning we don't get symmetry when $\texttt{rolloff} \notin \{ 0, 1 \}$.

quil-rs/src/waveform/snapshots/quil_rs__waveform__builtin__tests__RaisedCosine__rolloff_4e-5_pad_left_0_0_pad_right_0_0___CommonBuiltinParameters__duration_0_0001_scale_Some(1_0)_phase_Some(Cycles(0_0))_detuning_Some(0_0)___plot.snap
mingyoungjeng: I realized I picked a bad example for `rolloff` here. Updated test case with `rolloff=0.5`, which is a much better showcase.

quil-rs/src/waveform/builtin.rs
```diff
@@ -912,6 +938,107 @@ impl<T: WaveformData> HermiteGaussian<T> {
...
+            } else {
+                IqSamplesFor::Partial(
+                    is_partial,
+                    IqSamples::Samples(vec![
+                        ();
+                        left_padding_samples
+                            + sample_count
+                            + right_padding_samples
+                    ]),
+                )
```
jselig-rigetti, line 979: I don't understand what "partial" values are for, but what does `vec![(); total]` do differently that `all_zero`?

quil-rs/src/waveform/builtin.rs
```diff
@@ -912,6 +938,107 @@ impl<T: WaveformData> HermiteGaussian<T> {
...
+                let square_pulse = real!(1.0);
+                let cosine_pulse = |el: f64, rolloff_factor: f64| {
+                    real!(
+                        0.5 * (1.0
+                            + f64::cos(PI + 2.0 * PI * el / (rolloff_factor * active_duration)))
+                    )
+                };
+
+                // TEMP: we do not use fwhm or sigma.
+                let fwhm = 0.0;
```
jselig-rigetti: Why `TEMP`, getting removed or just a note?

quil-rs/src/waveform/builtin.rs
```diff
@@ -912,6 +938,107 @@ impl<T: WaveformData> HermiteGaussian<T> {
...
+    ) -> Result<IqSamplesFor<T>, SamplingError>
+    where
+        Self: ConcretizableFromTo<T, RaisedCosine<Concrete>>,
+    {
+        let scale_is_zero = common
+            .scale
+            .is_some_and(|scale| T::eval_real(scale) == Ok(0.0));
+
+        let left_padding_samples = (self.pad_left * sample_rate).ceil() as usize;
+        let right_padding_samples = (self.pad_right * sample_rate).ceil() as usize;
```
jselig-rigetti, line 954: What are the units of sample rate - is this `Hz`? What makes it need to be `f64` and not like uint, the quil spec?

jselig-rigetti (approved):
In general looks good, mostly just questions 👍

quil-rs/src/waveform/builtin.rs
mingyoungjeng: Yep, just removed it! Just wanted to make sure the waveform implementation itself was sound first.

quil-rs/src/waveform/builtin.rs
mingyoungjeng: "Partial" refers to incomplete evaluation of the waveform, so analogous to the difference between null and zero.

quil-rs/src/waveform/builtin.rs
mingyoungjeng: It is $Hz$. 

If the spec says it should be a `uint` value, it probably should be.  However, that would be breaking, so probably make that a separate issue.

antalsz: @jselig-rigetti `SAMPLE-RATE` is part of `DEFFRAME`, and so not addressed by the Quil spec.  The choice of `f64` has been present in quil-rs for a while

PR #520, opened 2026-08-11
feat!: promote padding to a common parameter for builtin waveforms

mingyoungjeng:


quil-rs/src/waveform/builtin.rs
mingyoungjeng, line 1546: These tests rely on the debug representation of the waveform structs; however, snapshot testing is not working for some tests because the filename is too long.

PR #522, opened 2026-08-14
`py.typed` file and pyproject authors fix

BatmanAoD:
* adds `py.typed` file so that downstream projects don't get an `[import-untyped]` error
* fixes the `authors` list so that it mentions the QCS team specifically and uses our email address (consistent with other projects)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -36,9 +36,9 @@ jobs:
         git config --global user.email "${{ github.triggering_actor}}@users.noreply.github.com"
 
     - name: Install Knope
-      uses: knope-dev/action@v2.1.0
+      uses: knope-dev/action@v2.1.2
       with:
-        version: 0.21.3 # Test before updating, breaking changes likely: https://github.com/knope-dev/action#install-latest-version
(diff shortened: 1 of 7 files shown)


.github/workflows/release.yml
```diff
@@ -36,9 +36,9 @@ jobs:
         git config --global user.email "${{ github.triggering_actor}}@users.noreply.github.com"
 
     - name: Install Knope
-      uses: knope-dev/action@v2.1.0
+      uses: knope-dev/action@v2.1.2
       with:
-        version: 0.21.3 # Test before updating, breaking changes likely: https://github.com/knope-dev/action#install-latest-version
+        version: 0.23.0 # Test before updating, breaking changes likely: https://github.com/knope-dev/action#install-latest-version
```
BatmanAoD, line 41: permits publishing prereleases from branches

quil-rs/python/quil/py.typed
BatmanAoD, line 1: This is the only change that affects downstream packages/users

Shadow53 (approved):

PR #524, opened 2026-08-20
fix: remove 'paste' from dependency tree

### LINKED ISSUE #506: Migrate from `paste` to `pastey`
`paste` is unmaintained

BatmanAoD:
Resolves #506, and fully removes `paste` by updating `statrs`. Looking at the [changelog](https://github.com/statrs-dev/statrs/blob/main/CHANGELOG.md), the breaking changes mentioned look to me like they should be build errors rather than runtime errors, so the upgrade should be safe.


quil-rs/src/parser/error/error.rs
```diff
@@ -18,7 +18,6 @@ use nom::error::{Error as NomError, ParseError};
 use std::convert::Infallible;
 use std::fmt;
 use std::sync::Arc;
-use thiserror::__private::AsDynError;
```
BatmanAoD, line 21: This `__private` is no longer present (or maybe just no longer public) in `thiserror` v2. So we need to do the coercion ourselves.

Cargo.toml
```diff
@@ -5,6 +5,42 @@ resolver = "2"
 
 [workspace.dependencies]
 clap = { version = "4.5.4", features = ["derive"] }
+anyhow = "1.0.103"
+approx = "0.5.1"
```
BatmanAoD, line 9: Bringing these dependencies to the workspace level was done automatically, with `cargo autoinherit`.

Cargo.toml
```diff
@@ -5,6 +5,42 @@ resolver = "2"
...
+optipy = "0.1.0"
+pastey = "0.2"
+petgraph = "0.6.4"
+pretty_assertions = "1.4.0"
+proptest = "1.0.0"
+proptest-derive = "0.5.1"
+pyo3 = "0.29"
+pyo3-build-config = "0.29"
+pyo3-stub-gen = "0.23.0"
+quil-rs = { version = "0.37.1", path = "quil-rs" }
```
asaites: I think the `knope.toml` file needs to be updated with this change. It currently has:

```toml
[packages.quil-rs]
versioned_files = [
    "quil-rs/Cargo.toml",
    "quil-rs/pyproject.toml",
    "Cargo.lock",
    # Update the quil-rs version in quil-cli
    { path = "quil-cli/Cargo.toml", dependency = "quil-rs" },
]
```

I think the last bit should be:

```toml
    # Update the quil-rs version in the workspace
    { path = "Cargo.toml", dependency = "quil-rs" },
```

asaites (changes requested):
Mostly this looks fine, but the `knope.toml` needs updating with the workspace dependency change.

Cargo.toml
BatmanAoD: Yep, seems right; thanks!

asaites (approved):

